PR #4, opened 2018-09-18
Add support for Bean Validation 2.0

gscheibel:


src/main/kotlin/com/expedia/graphql/KotlinDataFetcher.kt
```diff
@@ -39,11 +47,27 @@ class KotlinDataFetcher(
...
+                    parseArgument(environment, name, klazz, annotations)
                 }
             }.toTypedArray())
         } else {
             null
         }
+
+    }
+
+    private fun parseArgument(environment: DataFetchingEnvironment, name: String, klazz: Class<*>, annotations: List<Annotation>): Any? {
```
dkuc84: this will attempt to parse annotations for each query/field invocation which might be an overkill

src/main/kotlin/com/expedia/graphql/schema/SchemaGenerator.kt
```diff
@@ -45,6 +47,7 @@ internal class SchemaGenerator(
     private val propertyFilters: List<PropertyFilter> = listOf(isPublic, isNotGraphQLIgnored)
     private val functionFilters: List<FunctionFilter> = listOf(isPublic, isNotGraphQLIgnored, isNotBlackListed)
     private val additionTypes = mutableSetOf<GraphQLType>()
+    private val validator = Validation.buildDefaultValidatorFactory().validator
```
dkuc84: this means we cannot register custom validators

dkuc84:
This seems like a pretty heavy process. While the validation will be run on the server it does not tell anything about the required validation to the clients besides the documentation. Maybe this should be exposed as directives instead?

PR #9, opened 2018-09-24
Add example app

dkuc84:
* sample example app that exposes queries/mutations used in the README
* fixes selection of data fetcher for classes implementing interfaces
* add support for directives


dkuc84:
I could move the example app under test directory and use it for integration tests but that would also mean those tests would rely on graphql-servlet and spring boot to run. As it is right now example app is a sub project that within the repo but it also means that i cannot infer version from the graphql-kotlin pom.xml.

PR #11, opened 2018-09-27
Add support to unwrap CompletableFuture and override with config

brennantaylor:


src/main/kotlin/com/expedia/graphql/schema/monads.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+
+data class MonadType(val klass: KClass<*>, val indexOfInnerType: Int)
+
+internal fun KType.unwrapMonad(config: SchemaGeneratorConfig): KType {
+    for (monadType in config.monadTypes) {
+        if (monadType.klass == this.classifier) {
+            return this.arguments[monadType.indexOfInnerType].type!!
```
smyrick: I'm confused why we need the index. Why can't we just find the argument matching the type. 

Also we should avoid using `!!` or atleast comment why we need to use it

brennantaylor: If we fail I can fallback to returning the original type.

Also imagine some Monadic type where the result type is not the first index.

`MyMonad<T, U>` where `U` is the result of the computation.

`graphql-java-tools` supports similar config for handling `genericWrappers` https://github.com/graphql-java/graphql-java-tools#schema-parser-options

smyrick: Maybe we should just accept a functional interface on how to resolve the monad. Let me type it out

smyrick:
```
typealias MonadResolver = (KType) -> KType

data class MonadType(val klass: KClass<*>, val resolver: MonadResolver)

val completableMonad: MonadType = MonadType(CompletableFuture::class) { type: KType ->
    type.arguments[0].type ?: type
}

data class SchemaGeneratorConfig(
        val supportedPackages: String,
        val topLevelQueryName: String = "TopLevelQuery",
        val topLevelMutationName: String = "TopLevelMutation",
        val hooks: SchemaGeneratorHooks = NoopSchemaGeneratorHooks(),
        val monadTypes: Set<MonadType> = listOf(completableMonad).toSet()
)

fun KType.unwrapMonad(config: SchemaGeneratorConfig): KType {
    for (monadType in config.monadTypes) {
        if (monadType.klass == this.classifier) {
            return monadType.resolver(this)
        }
    }
    return this
}
```

brennantaylor: Seems fair.

PR #12, opened 2018-09-27
fix: hide all internal classes

smyrick:
Only requires a patch update.

Mark all internal classes as `internal` and move any internal classes to a more appropriate package


src/main/kotlin/com/expedia/graphql/schema/KotlinDataFetcher.kt
```diff
@@ -20,7 +20,7 @@ data class Parameter(val klazz: Class<*>, val annotations: List<Annotation>)
  * @param fn The Kotlin function being invoked
  * @param args The GraphQL arguments passed to the data fetcher
  */
-class KotlinDataFetcher(
+internal class KotlinDataFetcher(
```
dkuc84: i don't think we want to make this internal -> e.g. in instrumentation you can do a simple check `if you are using this data fetcher then it means you are instrumenting function call`

smyrick: But we don't have anyway for users to pass in or define a `KotlinDataFetcher` through the `toSchema` call right now. Isn't this just being used internally?

PR #14, opened 2018-09-27
fix: resolve some detekt issues

smyrick:
Fix Detekt issues and reduce the number of overrides/supressions we have


src/main/kotlin/com/expedia/graphql/schema/hooks/SchemaGeneratorHooks.kt
```diff
@@ -8,6 +8,9 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.KProperty
 import kotlin.reflect.KType
 
+/**
+ * Collection of all the hooks when generating a schema.
```
brennantaylor: Maybe define what the hooks are? Something about lifecycle events while the schema is building to allow the user to customize the schema.

src/main/kotlin/com/expedia/graphql/schema/exceptions/NestingNonNullTypeException.kt
```diff
@@ -3,5 +3,8 @@ package com.expedia.graphql.schema.exceptions
 import graphql.schema.GraphQLType
 import kotlin.reflect.KType
 
+/**
+ * Throws on wrapping a non-null graphql type twice.
```
brennantaylor: Maybe `wrapping -> nesting`?

src/main/kotlin/com/expedia/graphql/schema/hooks/SchemaGeneratorHooks.kt
smyrick: Updated.

PR #17, opened 2018-10-01
Enable union support

dkuc84:


src/main/kotlin/com/expedia/graphql/schema/SchemaGenerator.kt
```diff
@@ -270,6 +271,12 @@ internal class SchemaGenerator(
         return builder.build()
     }
 
+    private fun abstractType(kClass: KClass<*>): GraphQLType = if (kClass.declaredMemberFunctions.isEmpty() && kClass.declaredMemberProperties.isEmpty()) {
```
gscheibel: I think it's easier to read is the type detection logic stays in the `getGraphQLType` method.

src/main/kotlin/com/expedia/graphql/schema/SchemaGenerator.kt
```diff
@@ -297,4 +304,22 @@ internal class SchemaGenerator(
...
+
+        builder.name(kClass.simpleName)
+        builder.description(kClass.graphQLDescription())
+
+        builder.typeResolver { env: TypeResolutionEnvironment -> env.schema.getObjectType(env.getObject<Any>().javaClass.simpleName) }
+
+        val reflections = Reflections(config.supportedPackages)
+        val implementations = reflections.getSubTypesOf(Class.forName(kClass.javaObjectType.name))
+        implementations.forEach {
+            val possibleType = objectType(it.kotlin, null)
```
gscheibel: interfaces should be filtered out otherwise you'll end up in a bad state when a type implements another interface

dkuc84:
Closing it in favor of another PR

PR #19, opened 2018-10-01
Add support for graphql union types

gscheibel:


example/src/main/kotlin/com.expedia.graphql.sample/query/UnionQuery.kt
```diff
@@ -0,0 +1,17 @@
+package com.expedia.graphql.sample.query
+
+import org.springframework.stereotype.Component
+
+@Component
+class QueryWithUnion: Query {
```
dkuc84: do we want to move this to `PolymorphicQuery` instead?

example/src/main/kotlin/com.expedia.graphql.sample/query/UnionQuery.kt
```diff
@@ -0,0 +1,17 @@
...
+    fun whichHand(whichHand: String): AUnionType = when(whichHand) {
+        "right" -> RightHand(12)
+        else -> LeftHand("hello world")
+    }
+}
+
+interface AUnionType
```
dkuc84: by convention I put all model classes into `model` package, lets keep it consistent

src/main/kotlin/com/expedia/graphql/schema/SchemaGenerator.kt
```diff
@@ -41,6 +42,7 @@ internal class SchemaGenerator(
     private val typesCache: MutableMap<String, KGraphQLType> = mutableMapOf()
     private val additionTypes = mutableSetOf<GraphQLType>()
     private val directives = mutableSetOf<GraphQLDirective>()
+    private val reflections = Reflections(config.supportedPackages)
```
dkuc84: i think you want to use it for interfaces as well

src/main/kotlin/com/expedia/graphql/schema/SchemaGenerator.kt
```diff
@@ -291,12 +299,28 @@ internal class SchemaGenerator(
...
+        val builder = GraphQLUnionType.newUnionType()
+
+        builder.name(kClass.simpleName)
+        builder.description(kClass.graphQLDescription())
+        builder.typeResolver { env: TypeResolutionEnvironment -> env.schema.getObjectType(env.getObject<Any>().javaClass.simpleName) }
+
+        val implementations = reflections.getSubTypesOf(Class.forName(kClass.javaObjectType.name))
+        implementations
+            .forEach {
+                builder.possibleType(objectType(it.kotlin) as GraphQLObjectType)
```
dkuc84: dont you have to add it to the `additionTypes`?

gscheibel: you don't have to as `objectType` adds the type to schema directly

example/src/main/kotlin/com.expedia.graphql.sample/query/UnionQuery.kt
gscheibel: 👍

example/src/main/kotlin/com.expedia.graphql.sample/model/UnionModel.kt
```diff
@@ -0,0 +1,7 @@
+package com.expedia.graphql.sample.model
+
+interface AUnionType
```
dkuc84: do we want to rename it to something like `BodyPart` (`AUnionType` is pretty weird)?

PR #38, opened 2018-10-07
update example app pom with spring-boot starter parent

tapaderster:


example/pom.xml
```diff
@@ -2,6 +2,12 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-parent</artifactId>
+        <version>2.0.1.RELEASE</version>
```
dkuc84: can you update it to `2.0.5.RELEASE` (to be consistent)

tapaderster: @dkuc84 updated

PR #45, opened 2018-10-17
Adding optional datafetcher predicate to run test on function parameters

gscheibel:


src/main/kotlin/com/expedia/graphql/schema/hooks/DataFetcherExecutionPredicate.kt
```diff
@@ -0,0 +1,18 @@
+package com.expedia.graphql.schema.hooks
+
+interface DataFetcherExecutionPredicate {
+
+    /**
+     * The test to perform on the argument.
+     *
+     * With this test, runtime evaluation can be performed such as validation
+     */
+    fun <T> test(value: T): Boolean
```
smyrick: Should we follow the `will` `did` pattern instead here and have two fields on the hooks?

src/main/kotlin/com/expedia/graphql/schema/generator/SchemaGenerator.kt
```diff
@@ -140,7 +140,7 @@ internal class SchemaGenerator(
         }
 
         if (!abstract) {
-            val dataFetcher: DataFetcher<*> = KotlinDataFetcher(target, fn, args)
+            val dataFetcher: DataFetcher<*> = KotlinDataFetcher(target, fn, args, config.hooks.dataFetcherExecutionPredicate)
```
dkuc84: i dont think this makes much sense -> you are setting the same validator for ALL queries/functions which will execute against ALL parameters. 

Given current signature of the `DataFetcherExecutionPredicate` it just executes `test(value: T)` without any context on what it should be checking, e.g. lets say I add a validation to check if numbers are positive -> you would not only have to make the validator "smart" to skip all non number fields but it would also apply that check against ALL number fields (or maybe just ints or longs). 

IMHO in order for this to be useful, users should be able to define those validators per query/function (or even better per input param). Otherwise I am unsure how much value those global validators would give you.

gscheibel: The mechanic is generic but then in terms of implementation and usages it will be done at the parameter / type level

dkuc84: What I meant is that in the current implementation it takes just a value to be validated. There is no notion of which query/function nor which parameter is being evaluated. It only checks if value is correct which implies global validation logic.

After thinking about it I am unsure if adding extra validation hooks (whether through config, directives or some annotations) will be simpler than just implementing validation logic to the target method/functions.

gscheibel: A possible usage is actually via annotation using bean validation.
Where the argument passed to the data fetcher are annotated.

For example:

```
class ValidatedQuery : Query {
    fun typeWithValidation(arg: NeedValidation): Int = arg.greaterThan2

    fun argumentWithValidation(@GraphQLValidation arg: TypeWithPattern): String = arg.lowerCaseOnly
}

@GraphQLValidation
data class NeedValidation(
    @field:Min(2)
    val greaterThan2: Int
)


data class TypeWithPattern(
    @field:Pattern(regexp = "[a-z]*")
    val lowerCaseOnly: String
)
```

Note that in this example `@GraphQLValidation` is actually a directive marking the that the element must be validated.
The advantage of using a directive here is it can be monitored via instrumentation.

The predicate logic must be totally from the usage and so using Bean Validation or a more manual approach is up to the consumer of the library if they chose to trigger the predicate.

dkuc84: in the current logic you are still just passing raw value to be evaluated by the validator so it doesn't know anything about any annotations at all

gscheibel:
The annotation logic is independent from the generator lib, I'm going to
push an implementation in the example app using the annotation and the
validator provided by spring

src/main/kotlin/com/expedia/graphql/schema/hooks/DataFetcherExecutionPredicate.kt
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Perform the predicate logic by evaluating the argument and its value
+     * Then depending on the result either returning the value itself to continue the datafetcher invocation
+     * or call break the data fetching execution.
+     *
+     * @param parameter the function argument reference
+     * @param argumentName the name of the argument as declared in the query / kotlin function
+     * @param value the value passed in the query by the user
+     */
+    fun execute(parameter: Parameter, argumentName: String, value: Any): Any {
```
dkuc84: this is still too limited -> I think in order for this to be useful you would also need to pass the info about the target class that defines the function as well as the target function. 

argumentName is pretty useless maybe except for the logging purposes

src/main/kotlin/com/expedia/graphql/schema/generator/SchemaGenerator.kt
dkuc84: current logic is still too limited -> validation doesnt know anything about the target function (and I'd assume different functions might require different validations)

gscheibel:
The Parameter object contains the targeted kclass and the annotations of
the argument

example/src/main/kotlin/com.expedia.graphql.sample/extension/CustomSchemaGeneratorHooks.kt
```diff
@@ -22,6 +25,7 @@ class CustomSchemaGeneratorHooks: NoopSchemaGeneratorHooks() {
         else -> null
     }
 
+    override val dataFetcherExecutionPredicate: DataFetcherExecutionPredicate? = DataFetcherExecutionValidator(validator)
```
dkuc84: i think you can drop the nullability `?` from the declaration

example/src/main/kotlin/com.expedia.graphql.sample/validation/DataFetcherExecutionValidator.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.expedia.graphql.sample.exceptions.ConstraintError
+import com.expedia.graphql.sample.exceptions.ValidationException
+import com.expedia.graphql.schema.Parameter
+import com.expedia.graphql.schema.hooks.DataFetcherExecutionPredicate
+import graphql.schema.DataFetchingEnvironment
+import javax.validation.ConstraintViolation
+import javax.validation.Valid
+import javax.validation.Validator
+
+class DataFetcherExecutionValidator(private val validator: Validator?) : DataFetcherExecutionPredicate() {
```
dkuc84: since you define the class to be initialized with validator -> why not make it non nullable?

src/main/kotlin/com/expedia/graphql/schema/hooks/DataFetcherExecutionPredicate.kt
```diff
@@ -0,0 +1,63 @@
...
+     * or call break the data fetching execution.
+     *
+     * @param parameter the function argument reference
+     * @param argumentName the name of the argument as declared in the query / kotlin function
+     * @param value the value passed in the query by the user
+     */
+    fun <T> execute(value: T, parameter: Parameter, argumentName: String, environment: DataFetchingEnvironment): T {
+        val evaluationResult = evaluate(value, parameter, argumentName, environment)
+
+        return when {
```
dkuc84: if there are just two cases maybe we should make it `if` instead?

src/main/kotlin/com/expedia/graphql/schema/hooks/DataFetcherExecutionPredicate.kt
```diff
@@ -0,0 +1,63 @@
...
+            test(evaluationResult) -> value
+            else -> onFailure(evaluationResult, parameter, argumentName, environment)
+        }
+    }
+
+    /**
+     * Evaluate if the given data fetcher parameter.
```
dkuc84: looks like sentence is not finished

src/main/kotlin/com/expedia/graphql/schema/hooks/SchemaGeneratorHooks.kt
```diff
@@ -64,4 +64,10 @@ interface SchemaGeneratorHooks {
      * Called after converting the function to a field definition but before adding to the schema to allow customization
      */
     fun didGenerateMutationType(function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition
+
+    /**
+     * Execute a execute on each function parameters after their deserialization
```
dkuc84: think this needs to be rephrased

example/src/main/kotlin/com.expedia.graphql.sample/extension/CustomSchemaGeneratorHooks.kt
```diff
@@ -1,18 +1,21 @@
...
 import java.util.UUID
+import javax.validation.Validator
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
 
 /**
  * Schema generator hook that adds additional scalar types.
  */
-class CustomSchemaGeneratorHooks: NoopSchemaGeneratorHooks() {
+class CustomSchemaGeneratorHooks(validator: Validator? = null) : NoopSchemaGeneratorHooks() {
```
dkuc84: you can drop the `?` from the variable

src/main/kotlin/com/expedia/graphql/schema/hooks/DataFetcherExecutionPredicate.kt
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Perform runtime evaluations of each parameter passed to any KotlinDataFetcher.
+ *
+ * The DataFetcherExecutionPredicate is declared globally for all the datafetchers instances and all the parameters.
+ * However a more precise logic (at the field level) is possible depending on the implement of `evaluate` and `test`
+ *
+ * The predicate logic is split into two parts (evaluate and test) so the result of the evaluation like a list of errors
+ * can be passed to the onFailure method and added to an exception
+ *
+ * Because the DataFetcherExecutionPredicate is global, it's not possible to have methods where the type is inferred.
```
dkuc84: i think this is incorrect? we are passing the `DataFetchingEnvironment` now

src/main/kotlin/com/expedia/graphql/schema/generator/SchemaGenerator.kt
dkuc84: for the sake of completeness -> due to other changes in the PR above comments were already addressed

PR #47, opened 2018-10-18
refactor: start moving type mappers to separate files

smyrick:
Start breaking up the massive class, SchemaGenerator, and move bits into their own functions in files with unit tests


src/main/kotlin/com/expedia/graphql/schema/generator/types/enumType.kt
```diff
@@ -0,0 +1,16 @@
+package com.expedia.graphql.schema.generator.types
+
+import graphql.schema.GraphQLEnumType
+import kotlin.reflect.KClass
+
+internal fun enumType(kClass: KClass<*>): GraphQLEnumType {
+    val enumKClass = @Suppress("UNCHECKED_CAST") (kClass as KClass<Enum<*>>)
```
dkuc84: should we do a safe cast `as?` instead? e.g. if you pass a kClass that is not an enum we could fail gracefully

smyrick: Will do. I will add a unit test as well

smyrick: I have moved the cast to a location where we have already checked the type

Issue #51, opened 2018-10-26
Support for GraphQLID


Togrias:
Hi,

Of the "built-in" scalars, GraphQLID is missing. Is it possible to return an output type of GraphQLID?

brennantaylor:
Hi there! Thanks for opening an issue. It does seem like we missed this and would like to support it in the lib. We're going to think about how to represent the object in kotlin a bit so this may take some time.

In the meantime PRs are welcome and you could add support for UUID (I think we have an example)of orust use a string in the meantime to unblock yourself.

smyrick:
In the meantime you can also use custom resolvers to resolve whatever Kotlin object you use back into the `graphql.GraphQLID`

smyrick:
@ExpediaDotCom/graphql-kotlin I think the best approach here is to just map all `java.util.UUID` to `graphql.GraphQLID`. `graphql-java` supports it [here in their mapping](https://github.com/graphql-java/graphql-java/blob/25378bd600cd1cd5af6c067be3a2a1bf140b7649/src/main/java/graphql/Scalars.java#L266) along with `String`, `Integer`, and `Long` but I think those types could get confusing. Clients can still provide custom overrides if they want to change this behaviour.

smyrick:
Fixed in https://github.com/ExpediaDotCom/graphql-kotlin/pull/54

PR #54, opened 2018-10-30
Add support for GraphQL Scalar ID

gscheibel:
This PR addresses the issue #51


wsulemanexpedia:
Is it not reasonable to assume that a UUID type object is an ID without having to annotate it?

PR #55, opened 2018-10-30
GraphQL ID example

gscheibel:


example/src/main/kotlin/com.expedia.graphql.sample/query/ScalarQuery.kt
```diff
@@ -0,0 +1,31 @@
...
+class ScalarQuery: Query {
+
+    @GraphQLDescription("generates random UUID")
+    fun generateRandomUUID() = UUID.randomUUID()
+
+    fun findPersonById(id: Int) = Person(id, "Nelson")
+}
+
+@Component
+class ScalarMutation : Mutation {
```
dkuc84: maybe we should move it to mutation package?

PR #67, opened 2018-11-14
Correct syntax for Gradle dependency in README.md

dennisfen:


brennantaylor:
Updated the wiki. Thanks!

Issue #60, opened 2018-11-12
How to wire up GraphQLDirective


d4rken:
I'm trying to use `GraphQLDirective` to do authentication,[ but I don't understand how I wire them up](https://github.com/ExpediaDotCom/graphql-kotlin/wiki/Schema-Directives).

According to the [GraphQL-Java docs](https://www.graphql-java.com/documentation/v11/sdl-directives/) I need the add the `Directive` to a `RuntimeWiring` instance and include that instance in the generated `GraphQLSchema`.

How can I do that using `graphql-kotlin`?

I'm trying to use them for authentication, e.g.:

* https://www.prisma.io/blog/graphql-directive-permissions-authorization-made-easy-54c076b5368e/
* https://hackernoon.com/easy-peasy-graphql-authentication-and-authorization-using-schema-directives-f28f1845da20

d4rken:
I think my idea doesn't work because we can only use `GraphQLObjectType.field.transform {  }` in `fun didGenerateGraphQLType(type: KType, generatedType: GraphQLType)` which doesn't alter the existing object instance, but we need to modify/wrap the existing object in `SchemaGenerator`...

There is `fun willGenerateGraphQLType(type: KType): GraphQLType?` but this doesn't yet have the directive information we need...

So i guess we need to introduce something like  `SchemaGeneratorHooks.onModifyGeneratedGraphQLType(...) : GraphQLType`

Issue #53, opened 2018-10-29
Support for default schema values


smyrick:
The GraphQL spec allows for arguments to have default values:
* https://graphql.org/learn/schema/#arguments

Kotlin allows specifying default values:
* https://kotlinlang.org/docs/reference/functions.html#default-arguments

This information is not available on reflection though due to the way Kotlin compiles to the JVM:
* https://discuss.kotlinlang.org/t/kotlin-reflection-and-default-values/2254
* https://discuss.kotlinlang.org/t/retrieve-default-parameter-value-via-reflection/7314

We could solve this issue by adding a new annotation `@GraphQLDefaultValue`. This would require developers to duplicate some information but I think it would be worth it until we have a better way of extracting the info

```kotlin
fun getWidget(id: Int, @GraphQLDefaultValue(WidgetType.UNICORN) type: WidgetType = WidgetType.UNICORN) {
    return database.getWidgetByIdAndType(id, type)
}
```

amandaducrou:
I like the annotation solution. This way we can add a check to the schema validation that all arguments with default values have the annotation.

smyrick:
A small issue: the values of annotations can only be a select list of values, basically primitives: https://kotlinlang.org/docs/reference/annotations.html#constructors

-----

I thought this might work:

```
annotation class GraphQLDefaultValue(val kClass: KClass<*>, val valueAsString: String)
```

Then we could use it like `kClass.parse(valueAsString)` but there is not a generic method to parse a value from a string as far as I can see. We could add a custom extension function but I am not sure if this would work for all use cases.

-----

Another approach is to have an annotation for each value `@GraphQLDefaultString` `@GraphQLDefaultInt`. But that means having to use and check lots of different values.

----

Any other ideas?

smyrick:
https://github.com/ExpediaDotCom/graphql-kotlin/pull/56

d4rken:
Here is an example on how default values can be achieved via custom directives, see:

https://github.com/ExpediaDotCom/graphql-kotlin/blob/861443970ae9859710967ae65c257e23bee47595/example/src/main/kotlin/com.expedia.graphql.sample/query/CustomDirectiveQuery.kt#L12

https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/example/src/main/kotlin/com.expedia.graphql.sample/directives/StringEvalDirectiveWiring.kt

amandaducrou:
[...]

PR #69, opened 2018-11-15
Additional SchemaGeneratorHook that allows modifying GraphQLTypes.

d4rken:
This enables rewiring based on directives (see #60) by offering general purpose hooks that allows us to rewire/modify generated graphqltypes.

The base example for directive based authentication, would work with this:
https://www.graphql-java.com/documentation/v11/sdl-directives/

With this hook available I can then use a helper class such as this can be used to wire up the directives:
https://gist.github.com/d4rken/7cb60a6a7b315d25766ff15884f9b99b

The new hook has a placeholder `WiringContext` which would need to get some functionality to allow for feature parity with graphql-java's [`SchemaDirectiveEnvironment`](https://github.com/graphql-java/graphql-java/blob/5eb2ab743bf7afc05e2243d3dc0bcf22a0384366/src/main/java/graphql/schema/idl/SchemaDirectiveWiringEnvironment.java)

I was a bit out of my depth there though and couldn't find a clean solution.

Theoretically we could just keep it empty until someone needs the functionality then provide it. As it's passed via "context" object this wouldn't lead to breaking changes. `WiringContext` is created inside the `SchemaGenerator` where it would have access to the necessary information it's just unclear in what form to provide it.

I think it comes down to how close we want to mirror graphql-java and it's `SchemaDirectiveWiringEnvironment`...

### DIFF
--- a/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
+++ b/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
@@ -11,7 +11,6 @@ import graphql.schema.GraphQLArgument
 import graphql.schema.GraphQLDirective
 import graphql.schema.GraphQLInputType
 import kotlin.reflect.KAnnotatedElement
-import kotlin.reflect.KClass
 import kotlin.reflect.full.findAnnotation
 import com.expedia.graphql.annotations.GraphQLDirective as DirectiveAnnotation
 
@@ -64,11 +63,10 @@ internal fun KAnnotatedElement.isGraphQLIgnored() = this.findAnnotation<GraphQLI
 internal fun KAnnotatedElement.isGraphQLID() = this.findAnnotation<GraphQLID>() != null
 
 private fun Annotation.getDirectiveInfo(): DirectiveInfo? {
-    val directiveAnnotation = this.annotationClass.annotations.find { it is DirectiveAnnotation } as? DirectiveAnnotation
-    return when {
-        directiveAnnotation != null -> DirectiveInfo(this.annotationClass.simpleName ?: "", directiveAnnotation)
-        else -> null
-    }
+    return this.annotationClass.annotations
+        .filterIsInstance(DirectiveAnnotation::class.java)
+        .map { DirectiveInfo(this, it) }
+        .firstOrNull()
 }
 
 internal fun KAnnotatedElement.directives(hooks: SchemaGeneratorHooks) =
@@ -79,21 +77,24 @@ internal fun KAnnotatedElement.directives(hooks: SchemaGeneratorHooks) =
 
 @Throws(CouldNotGetNameOfAnnotationException::class)
 private fun DirectiveInfo.getGraphQLDirective(hooks: SchemaGeneratorHooks): GraphQLDirective {
-    val kClass: KClass<out DirectiveAnnotation> = this.annotation.annotationClass
-    val builder = GraphQLDirective.newDirective()
-    val name: String = this.effectiveName ?: throw CouldNotGetNameOfAnnotationException(kClass)
+    val directiveClass = this.directive.annotationClass
+    val name: String = this.effectiveName ?: throw CouldNotGetNameOfAnnotationException(directiveClass)
 
     @Suppress("Detekt.SpreadOperator")
+    val builder = GraphQLDirective.newDirective()
+        .name(name.normalizeDirectiveName())
+        .validLocations(*this.directiveAnnotation.locations)
+        .description(this.directiveAnnotation.description)
 
-    builder.name(name.normalizeDirectiveName())
-        .validLocations(*this.annotation.locations)
-        .description(this.annotation.description)
+    directiveClass.getValidProperties(hooks).forEach { property ->
+        val propertyName = property.name
+        val value = property.call(this.directive)
 
-    kClass.getValidFunctions(hooks).forEach { kFunction ->
-        val propertyName = kFunction.name
-        val value = kFunction.call(kClass)
         @Suppress("Detekt.UnsafeCast")
-        val type = defaultGraphQLScalars(kFunction.returnType) as GraphQLInputType
+        var type: GraphQLInputType? = defaultGraphQLScalars(property.returnType) as? GraphQLInputType
+        if (type == null) {
+            type = hooks.willGenerateGraphQLType(property.returnType) as? GraphQLInputType
+        }
         val argument = GraphQLArgument.newArgument()
             .name(propertyName)
             .value(value)
@@ -107,10 +108,10 @@ private fun DirectiveInfo.getGraphQLDirective(hooks: SchemaGeneratorHooks): Grap
 
 private fun String.normalizeDirectiveName() = CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_CAMEL, this)
 
-private data class DirectiveInfo(private val name: String, val annotation: DirectiveAnnotation) {
+private data class DirectiveInfo(val directive: Annotation, val directiveAnnotation: DirectiveAnnotation) {
     val effectiveName: String? = when {
-        annotation.name.isNotEmpty() -> annotation.name
-        name.isNotEmpty() -> name
+        directiveAnnotation.name.isNotEmpty() -> directiveAnnotation.name
+        directive.annotationClass.simpleName.isNullOrEmpty().not() -> directive.annotationClass.simpleName
         else -> null
     }
 }
--- a/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
+++ b/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
@@ -93,19 +93,15 @@ private fun DirectiveInfo.getGraphQLDirective(hooks: SchemaGeneratorHooks): Grap
         .validLocations(*this.directiveAnnotation.locations)
         .description(this.directiveAnnotation.description)
 
-    directiveClass.getValidProperties(hooks).forEach { property ->
-        val propertyName = property.name
-        val value = property.call(this.directive)
+    directiveClass.getValidProperties(hooks).forEach { prop ->
+        val propertyName = prop.name
+        val value = prop.call(this.directive)
 
-        @Suppress("Detekt.UnsafeCast")
-        var type: GraphQLInputType? = defaultGraphQLScalars(property.returnType) as? GraphQLInputType
-        if (type == null) {
-            type = hooks.willGenerateGraphQLType(property.returnType) as? GraphQLInputType
-        }
+        val type = defaultGraphQLScalars(prop.returnType) ?: hooks.willGenerateGraphQLType(prop.returnType)
         val argument = GraphQLArgument.newArgument()
             .name(propertyName)
             .value(value)
-            .type(type)
+            .type(type as? GraphQLInputType)
             .build()
         builder.argument(argument)
     }
--- a/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
+++ b/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
@@ -11,6 +11,7 @@ import graphql.schema.GraphQLArgument
 import graphql.schema.GraphQLDirective
 import graphql.schema.GraphQLInputType
 import kotlin.reflect.KAnnotatedElement
+import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 import com.expedia.graphql.annotations.GraphQLDirective as DirectiveAnnotation
 
@@ -70,6 +71,12 @@ private fun Annotation.getDirectiveInfo(): DirectiveInfo? {
 }
 
 internal fun KAnnotatedElement.directives(hooks: SchemaGeneratorHooks) =
+    this.annotations.asSequence()
+        .mapNotNull { it.getDirectiveInfo() }
+        .map { it.getGraphQLDirective(hooks) }
+        .toList()
+
+internal fun KParameter.directives(hooks: SchemaGeneratorHooks) =
     this.annotations.asSequence()
         .mapNotNull { it.getDirectiveInfo() }
         .map { it.getGraphQLDirective(hooks) }
--- a/src/main/kotlin/com/expedia/graphql/schema/generator/SchemaGenerator.kt
+++ b/src/main/kotlin/com/expedia/graphql/schema/generator/SchemaGenerator.kt
@@ -56,7 +56,6 @@ internal class SchemaGenerator(
 
     private val state = SchemaGeneratorState(config.supportedPackages)
     private val subTypeMapper = SubTypeMapper(config.supportedPackages)
-    private val wiringContext = WiringContext()
 
     internal fun generate(): GraphQLSchema {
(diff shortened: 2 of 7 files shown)


src/main/kotlin/com/expedia/graphql/schema/generator/SchemaGenerator.kt
```diff
@@ -163,11 +164,18 @@ internal class SchemaGenerator(
                 .type(propertyType)
                 .deprecate(prop.getDeprecationReason())
 
-        return if (config.dataFetcherFactory != null && prop.isLateinit) {
+        prop.directives().forEach {
```
d4rken: Was this intentional that directives were not added to properties?

dkuc84: our use case was just for functions. i believe spec allows it to be on arbitrary field so we should update it

d4rken:
Some attempt at providing the node tree and registry, but I don't think it's clean.

```Kotlin
class WiringContext internal constructor(val typeCache: TypesCache) {
    private val schemaParser = SchemaParser()
    private val schemaPrinter = SchemaPrinter()

    internal lateinit var builder: GraphQLSchema.Builder

    /**
     * @return a shared map that can be used to hold context information during schema generation
     */
    val buildContext: MutableMap<String, Any> = mutableMapOf()

    fun getTypeDefinitionRegistry(): TypeDefinitionRegistry? {
        return try {
            schemaParser.parse(schemaPrinter.print(builder.build()))
        } catch (e: AssertException) {
            return null
        }
    }

    fun getNodeParentTree(): NodeParentTree<NamedNode<*>>? {
        val list = typeCache.getCacheStack()
                .map { it.graphQLType }
                .map { if (it is GraphQLNonNull) it.wrappedType else it }
                .mapNotNull {
                    return@mapNotNull when (it) {
                        is GraphQLObjectType -> it.definition
                        is GraphQLFieldDefinition -> it.definition
                        is GraphQLInterfaceType -> it.definition
                        is GraphQLUnionType -> it.definition
                        is GraphQLScalarType -> it.definition
                        is GraphQLEnumType -> it.definition
                        is GraphQLEnumValueDefinition -> EnumValueDefinition(it.name)
                        is GraphQLArgument -> it.definition
                        is GraphQLInputObjectType -> it.definition
                        is GraphQLInputObjectField -> it.definition
                        else -> null
                    }
                }
                .filter { NamedNode::class.isInstance(it) }
                .map { NamedNode::class.cast(it) }
                .toList()

        return if (list.isEmpty()) null else NodeParentTree(ArrayDeque<NamedNode<*>>(list))
    }
}
```

src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -69,32 +69,38 @@ internal fun KAnnotatedElement.isGraphQLID() = this.findAnnotation<GraphQLID>()
...
     this.annotations.asSequence()
-        .mapNotNull { it.getDirectiveInfo() }
-        .map { it.getGraphQLDirective() }
+        .mapNotNull { it.getGraphQLDirective(config) }
         .toList()
 
 @Throws(CouldNotGetNameOfAnnotationException::class)
-private fun DirectiveAnnotation.getGraphQLDirective(): GraphQLDirective {
-    val kClass: KClass<out DirectiveAnnotation> = this.annotationClass
+private fun Annotation.getGraphQLDirective(config: SchemaGeneratorConfig): GraphQLDirective? {
```
d4rken: Instead of building directives based just on `GraphQLDirectives` (aka `DirectiveAnnotation` here) we also include the base annotation, e.g.:

```kotlin
@GraphQLDirective(description = "A permission the user needs to hold to use this") // <---- getDirectiveInfo()
annotation class HasPermission(val permission: Permission) // <---- `this`
```

src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -69,32 +69,38 @@ internal fun KAnnotatedElement.isGraphQLID() = this.findAnnotation<GraphQLID>()
...
     }
 
     @Suppress("Detekt.SpreadOperator")
     builder.name(normalizeDirectiveName(name))
-        .validLocations(*this.locations)
-        .description(this.description)
+        .validLocations(*directiveInfo.locations)
+        .description(directiveInfo.description)
+
+    kClass.getValidProperties().forEach { property ->
```
d4rken: Not sure why this was previously `getValidFunctions`?
AFAIK annotations can have no functions, but they can have parameters such as the above `permission`.

src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -69,32 +69,38 @@ internal fun KAnnotatedElement.isGraphQLID() = this.findAnnotation<GraphQLID>()
...
+        val value = property.call(this)
 
-    kClass.getValidFunctions().forEach { kFunction ->
-        val propertyName = kFunction.name
-        val value = kFunction.call(kClass)
         @Suppress("Detekt.UnsafeCast")
-        val type = defaultGraphQLScalars(kFunction.returnType) as GraphQLInputType
+        var type: GraphQLInputType? = defaultGraphQLScalars(property.returnType) as? GraphQLInputType
+        if (type == null) {
+            type = config.hooks.willGenerateGraphQLType(property.returnType)as? GraphQLInputType
```
d4rken: By proving the schemaconfig we can allow the user to use custom types in their custom directives via `Coercing` (e.g. `Permission` enums).

src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -69,32 +69,38 @@ internal fun KAnnotatedElement.isGraphQLID() = this.findAnnotation<GraphQLID>()
 internal fun Annotation.getDirectiveInfo(): DirectiveAnnotation? =
     this.annotationClass.annotations.find { it is DirectiveAnnotation } as? DirectiveAnnotation
 
-internal fun KAnnotatedElement.directives() =
+internal fun KAnnotatedElement.directives(config: SchemaGeneratorConfig) =
     this.annotations.asSequence()
-        .mapNotNull { it.getDirectiveInfo() }
-        .map { it.getGraphQLDirective() }
+        .mapNotNull { it.getGraphQLDirective(config) }
```
d4rken: Instead of building directives based just on `GraphQLDirectives` (aka `DirectiveAnnotation` here) we also include the base annotation, e.g.:

```kotlin
@GraphQLDirective(description = "A permission the user needs to hold to use this") // <---- getDirectiveInfo()
annotation class HasPermission(val permission: Permission) // <---- `this`
```

smyrick:
@d4rken We merging this PR https://github.com/ExpediaDotCom/graphql-kotlin/pull/68 which moves the default implementation to the interface. After the merge conflicts have been resolved with that PR we can help you out

dkuc84:
After looking into this I believe we'll need https://github.com/graphql-java/graphql-java/issues/1228 (or potentially https://github.com/graphql-java/graphql-java/issues/1076) to be resolved before we can add better directive support in graphql-kotlin.

Potential implementations:
*  [1228]( https://github.com/graphql-java/graphql-java/issues/1228) - since we are building schema programmatically we'll need to add custom data fetcher that would potentially wrap the execution of the original datafetcher while we are processing the given directive info. I'm thinking that those custom data fetchers would have to be explicitly provided by the users to the schema configuration.
* [1076](https://github.com/graphql-java/graphql-java/issues/1076) -  currently RuntimeWiring can only be applied when using `graphql-java` `SchemaGenerator` class to build schema. If we could specify runtime to be applied directly against the schema after its already build we could easily register custom `SchemaDirectiveWiring` implementations.

d4rken:
> After looking into this I believe we'll need graphql-java/graphql-java#1228 (or potentially graphql-java/graphql-java#1076) to be resolved before we can add better directive support in graphql-kotlin.

I'm not sure what you mean by that.

This PR isn't about perfection but about basic support. Currently it's not possible to setup custom logic via directives at all, e.g. this example is not possible:
https://www.graphql-java.com/documentation/v11/sdl-directives/

with my proposed changes it becomes possible and basically reaches feature parity with graphql-java.

> 1228 - since we are building schema programmatically we'll need to add custom data fetcher that would potentially wrap the execution of the original datafetcher while we are processing the given directive info. I'm thinking that those custom data fetchers would have to be explicitly provided by the users to the schema configuration.

That's what the above example does and what I'm trying to accomplish.
The proposed changes are the minimally required changes to make it work. I can now write code that does what you said and it looks like this:

https://gist.github.com/d4rken/7cb60a6a7b315d25766ff15884f9b99b

To be able to wrap the original datafetchers during schema generation we need the additional hook (and correctly parse the directive arguments in `annotationExtensions.kt`).

While some of this code _could_ be made part of this library, I don't think it is necessary and I'd rather keep `graphql-kotlin` lightweight. We could copy these snippets as example into the wiki though.

src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -79,21 +77,24 @@ internal fun KAnnotatedElement.directives(hooks: SchemaGeneratorHooks) =
...
+    directiveClass.getValidProperties(hooks).forEach { property ->
+        val propertyName = property.name
+        val value = property.call(this.directive)
 
-    kClass.getValidFunctions(hooks).forEach { kFunction ->
-        val propertyName = kFunction.name
-        val value = kFunction.call(kClass)
         @Suppress("Detekt.UnsafeCast")
-        val type = defaultGraphQLScalars(kFunction.returnType) as GraphQLInputType
+        var type: GraphQLInputType? = defaultGraphQLScalars(property.returnType) as? GraphQLInputType
```
dkuc84: can you rewrite this as (and remove the suppression)
```
val type = defaultGraphQLScalars(prop.returnType) ?: hooks.willGenerateGraphQLType(prop.returnType)
val argument = GraphQLArgument.newArgument()
    .name(propertyName)
    .value(value)
    .type(type as? GraphQLInputType)
    .build()
```

src/main/kotlin/com/expedia/graphql/schema/generator/WiringContext.kt
```diff
@@ -0,0 +1,9 @@
+package com.expedia.graphql.schema.generator
+
+/**
+ * Schema generation related information
+ */
+@Suppress("Detekt.EmptyClassBlock")
+class WiringContext internal constructor() {
```
dkuc84: lets drop this for now as it is not needed

d4rken: The idea was that we can improve upon this later without a breaking API change.

Any extra information we'd pass to the developer in a later library update would require either changing the method or adding a new one.

Due to knowing graph-java's implementation and assuming they made smart decisions, we know it might be useful to provide more data for wiring purposes ([see here](https://github.com/ExpediaDotCom/graphql-kotlin/issues/60#issuecomment-438514404)).

The current use-case doesn't require it, so I'm okay with removing this if you want me to, I just thought that we'd want to avoid breaking API changes / cluttering the public interface.

Specifically the "NodeParentTree" might be interesting for #53, which I think can be implemented via custom directives.

What do you think?

d4rken:
> Thanks for providing the example use case snippet. Can you also include DirectiveWiringHelper in this PR? I think it's functionality is pretty generic and we would end up reimplementing this.

Sure, in which package should I put it?

src/main/kotlin/com/expedia/graphql/schema/generator/WiringContext.kt
d4rken: I've got two examples now that work without further information, so YAGNI?

A breaking change might be okay anyways seeing as this library is a `0.0.X` and a breaking change will make developers aware of new possibilities ;).

So I concur and will remove it.

dkuc84:
lets put `DirectiveWiringHelper` under `com.expedia.graphql.schema.generator.directive`

src/main/kotlin/com/expedia/graphql/schema/generator/WiringContext.kt
d4rken: Can't actually find any examples using the extra information graphql-java provides, but maybe Google/Github search just sucks...

d4rken:
One more fix,  #53 is now possible too, (see 4067944d61f43a02e36c6c0d5e2ae343cfbcd5c5).

You can now make a custom directive and wire it up to check whether an argument is null and if it is replace it with the directives argument.

 E.g.

`DefaultValueDirective(defaultString = "strawberry")`

```kotlin
class DefaultValueDirectiveWiring @Inject constructor() : DirectiveWiring {
    private val dirName = getDirectiveName(DefaultValueDirective::class)

    override fun isResponsible(environment: SchemaDirectiveWiringEnvironment<*>): Boolean {
        (environment.element as? GraphQLFieldDefinition)?.arguments?.forEach { arg ->
            if (arg.getDirective(dirName) != null) {
                return true
            }
        }
        return false
    }

    override fun onField(wiringEnv: SchemaDirectiveWiringEnvironment<GraphQLFieldDefinition>): GraphQLFieldDefinition {
        val field = wiringEnv.element
        val originalDataFetcher: DataFetcher<Any> = field.dataFetcher

        val defaultValueFetcher = DataFetcher<Any> { dataEnv ->
            val newArguments = HashMap(dataEnv.arguments)
            wiringEnv.element.arguments.forEach { arg ->
                if ((dataEnv.getArgument(arg.name) as String?).isNullOrEmpty()) {
                    newArguments[arg.name] = arg.getDirective(dirName).getArgument(DefaultValueDirective::defaultString.name).value as String
                }
            }
            val newEnv = DataFetchingEnvironmentBuilder.newDataFetchingEnvironment(dataEnv)
                .arguments(newArguments)
                .build()
            originalDataFetcher.get(newEnv)
        }
        return field.transform { it.dataFetcher(defaultValueFetcher) }
    }
}
```

src/main/kotlin/com/expedia/graphql/schema/generator/WiringContext.kt
dkuc84: Since we currently don't have a use case for use of those fields its hard to say whether we will need it or not. Let's drop it for now.

d4rken: 80a629def32a9c130a5fc1b1778788217e471775

d4rken:
> lets put `DirectiveWiringHelper` under `com.expedia.graphql.schema.generator.directive`

7d8b986b29cd2c7fa1cf7f317d38ee8f5e155647

On a side note, klint is a bit overzealous...

src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
d4rken: Oh that's much nicer :) :+1: 

ba1c4f1080082643d5c179d0d21ef3c4a1356eb6

d4rken:
Is there interest in some demo/examples on custom directives with wiring for the wiki?

PR #70, opened 2018-11-16
Reconfigure surefire plugin

gscheibel:
to be sure that all the unit tests are executed properly

### DIFF
--- a/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
+++ b/src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
@@ -15,7 +15,7 @@ import kotlin.reflect.full.findAnnotation
 import com.expedia.graphql.annotations.GraphQLDirective as DirectiveAnnotation
 
 internal fun KAnnotatedElement.graphQLDescription(): String? {
-    val directiveNames = listOfDirectives().map { normalizeDirectiveName(it) }
+    val directiveNames = listOfDirectives().map { it.normalizeDirectiveName() }
 
     val description = this.findAnnotation<GraphQLDescription>()?.value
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -66,8 +61,13 @@ internal fun KAnnotatedElement.isGraphQLIgnored() = this.findAnnotation<GraphQLI
 
 internal fun KAnnotatedElement.isGraphQLID() = this.findAnnotation<GraphQLID>() != null
 
-internal fun Annotation.getDirectiveInfo(): DirectiveAnnotation? =
-    this.annotationClass.annotations.find { it is DirectiveAnnotation } as? DirectiveAnnotation
+private fun Annotation.getDirectiveInfo(): DirectiveInfo? {
+    val directiveAnnotation = this.annotationClass.annotations.find { it is DirectiveAnnotation } as? DirectiveAnnotation
```
d4rken: What's the benefit of these changes (this and the following lines)? Why are they problematic for the surefire plugin?

dkuc84: indeed this is unrelated to surefire plugin configuration. It is addressing another bug

d4rken: What was the bug? I changed the same code in #69 and need to resolve the merge conflict.

gscheibel: The problem here was when we have to set the name of the directive there are 2 options:
- from the `name` property of the `GraphQLDirective` annotation
- the name of the custom annotation (which is annotated with `@GraphQLDirective` 

However as we extracted the `@GraphQLDirective` type we lost access to the name of the custom directive class name.
So I extracted it into a new class and centralized the logic in the `effectiveName` method as it's shared with the description logic

Issue #78, opened 2018-11-28
Improving SchemaGenerator


d4rken:
I'm working on a PR that refactors `SchemaGenerator`. It's currently a bit unwieldy.

* Move responsibilities for different GraphQLTypes into their own classes
* Write extra test cases for each class (e.g. cover things like #76 )

PR #80, opened 2018-11-28
Refactored SchemaGenerator into multiple modules with individual tests.

d4rken:
See #78 

This is a work in progress.


d4rken:
Done. Looking for feedback.

This should be merged after #77 (merge conflicts).

d4rken:
This contains no breaking API change, it's purely internal.

smyrick:
Awesome! This is exactly what I was working towards earlier with some of the clean ups but didn't have time to get to it. I'm glad you saw the same patterns.  👍 

You have a few conflicts now but this looks good to me otherwise

d4rken:
Nice, just some finishing another PR+release, then I'll fix the conflicts here.

I was a bit unsure where to place the `objectFromReflection`  stuff whether `TypeBuilder` or `SchemaGenerator` would be best, I think the current solution is okay.

I also tried a bit more abstraction by removing the coupling between SchemaGenerator and TypeBuilder implementations via interfaces, but it ended up just being bloat and no advantage to testing, which works fine like it is now. `mockk` is great :+1:

d4rken:
lgtm :smile: 

For good measure I've compiled it against the example app too, works.

PR #82, opened 2018-11-29
refactor: move and delete duplicate code

smyrick:
Internal changes only

* Add some unit tests for `annotationExtenstions`
* Updated Detekt rules and finally delete the baseline file
* Removed `KParameter.directives()` since it is duplicate of `KAnnotatedElement.directives()`
* Remove default values for `getValid*()` functions as we don't use them in any source code


src/main/kotlin/com/expedia/graphql/schema/extensions/annotationExtensions.kt
```diff
@@ -23,73 +15,18 @@ internal fun KAnnotatedElement.getDeprecationReason(): String? = this.findAnnota
...
-        .firstOrNull()
-}
-
-internal fun KAnnotatedElement.directives(generator: SchemaGenerator) =
-    this.annotations.asSequence()
-        .mapNotNull { it.getDirectiveInfo() }
-        .map { it.getGraphQLDirective(generator) }
-        .toList()
-
-internal fun KParameter.directives(generator: SchemaGenerator) =
```
smyrick: When moving, I deleted the `KParameter.directives()` since `KParameter` extends from `KAnnotatedElement` so it when can just use that extension function

src/main/kotlin/com/expedia/graphql/schema/extensions/directiveExtensions.kt
```diff
@@ -0,0 +1,60 @@
...
+
+import com.expedia.graphql.annotations.GraphQLDirective
+import com.expedia.graphql.schema.exceptions.CouldNotGetNameOfAnnotationException
+import com.expedia.graphql.schema.generator.SchemaGenerator
+import com.google.common.base.CaseFormat
+import graphql.schema.GraphQLArgument
+import graphql.schema.GraphQLInputType
+import kotlin.reflect.KAnnotatedElement
+
+internal fun KAnnotatedElement.directives(generator: SchemaGenerator) =
```
smyrick: We only need to expose a single function to get directives on any element that has an annotation

src/main/kotlin/com/expedia/graphql/schema/extensions/kClassExtensions.kt
```diff
@@ -2,23 +2,20 @@ package com.expedia.graphql.schema.extensions
...
 import kotlin.reflect.full.declaredMemberProperties
 
-private val noopHooks = NoopSchemaGeneratorHooks()
-
-internal fun KClass<*>.getValidProperties(hooks: SchemaGeneratorHooks = noopHooks) = this.declaredMemberProperties
+internal fun KClass<*>.getValidProperties(hooks: SchemaGeneratorHooks) = this.declaredMemberProperties
     .filter { hooks.isValidProperty(it) }
     .filter { prop -> propertyFilters.all { it.invoke(prop) } }
 
-internal fun KClass<*>.getValidFunctions(hooks: SchemaGeneratorHooks = noopHooks) = this.declaredMemberFunctions
```
smyrick: We are not using the default value anywhere in source code

src/main/kotlin/com/expedia/graphql/schema/generator/directive/DirectiveWiringHelper.kt
```diff
@@ -25,6 +25,9 @@ import graphql.schema.idl.WiringFactory
  */
 class DirectiveWiringHelper(private val wiringFactory: WiringFactory, private val manualWiring: Map<String, SchemaDirectiveWiring> = mutableMapOf()) {
 
+    /**
+     * Wire up the directive based on the GraphQL type
+     */
```
smyrick: Linting warning about undocumented public functions

src/main/kotlin/com/expedia/graphql/schema/generator/TypeBuilder.kt
```diff
@@ -43,8 +43,9 @@ internal open class TypeBuilder constructor(val generator: SchemaGenerator) {
     private fun getGraphQLType(kClass: KClass<*>, inputType: Boolean, type: KType): GraphQLType = when {
         kClass.isSubclassOf(Enum::class) -> @Suppress("UNCHECKED_CAST") (generator.enumType(kClass as KClass<Enum<*>>))
         kClass.isSubclassOf(List::class) || kClass.java.isArray -> generator.listType(type, inputType)
-        kClass.canBeGraphQLUnion() -> generator.unionType(kClass)
-        kClass.canBeGraphQLInterface() -> generator.interfaceType(kClass)
-        else -> if (inputType) generator.inputObjectType(kClass) else generator.objectType(kClass)
+        kClass.isGraphQLUnion() -> generator.unionType(kClass)
```
d4rken: :+1:

src/main/kotlin/com/expedia/graphql/schema/generator/directive/DirectiveWiringHelper.kt
d4rken: I think that lint warning should be disabled.

src/main/kotlin/com/expedia/graphql/schema/hooks/SchemaGeneratorHooks.kt
```diff
@@ -14,6 +14,7 @@ import kotlin.reflect.KType
  * Hooks are lifecycle events that are called and triggered while the schema is building
  * that allow users to customize the schema.
  */
+@Suppress("Detekt.ComplexInterface")
```
d4rken: I think `ComplexInterface` and `TooManyFunctions` should also be disabled, the issues are not something you'd need a linter for and are often very much on purpose. Detekt and klint have such aggressive settings that I think it adds `@Supress` noise to the code.

src/test/kotlin/com/expedia/graphql/schema/extensions/AnnotationExtensionsTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import kotlin.test.assertTrue
+
+internal class AnnotationExtensionsTest {
+
+    @GraphQLDescription("class description")
+    @Deprecated("class deprecated")
+    @GraphQLIgnore
+    private data class WithAnnotations(
+        @Deprecated("property deprecated")
+        @property:GraphQLDescription("property description")
```
d4rken: Why is the `property` prefix used here? I think it's works without it?

smyrick: It does not work on data class properties. See the notes here: https://github.com/ExpediaDotCom/graphql-kotlin/wiki/@GraphQLDescription

src/main/kotlin/com/expedia/graphql/schema/hooks/SchemaGeneratorHooks.kt
smyrick: Yea, I agree. Since I deleted the detekt file here anyway I will add those rules in this PR

src/main/kotlin/com/expedia/graphql/schema/generator/directive/DirectiveWiringHelper.kt
smyrick: I think we should keep this one. As library creators we should document all the points that our users can interact with.

It is only a warning right now and does not fail.

smyrick:
@d4rken I thought the `is*` format was more clear about the boolean value but after looking at more it also seems to imply it is of type `GraphQLInterfaceType` when it is not, we are just checking whether we can map it to this type or now

PR #96, opened 2018-12-13
Force manual wirings to take precedence over wiring factory

dkuc84:


d4rken:
lgtm, maybe a test for the precedence?

PR #97, opened 2018-12-13
feat: remove requirement for @property: prefix

smyrick:
The issue was only with class properties. We couldn't set the annotation target because we also need the annotation to be valid on function arguments so classes still default the annotation to the constructor function arguments.

We need to check 4 possible annotation on a property:
* `@Deprecated`
* `@GraphQLDescription`
* `@GraphQLID`
* `@GraphQLIgnore`

The way I get around this is I first check if the property has the annotation, which is what we do right now so `@property:` prefixed annotations are handled first. If I can't find anything there I look at the parent class constructor, check it's arguments for a matching type (see  `KClass<*>.findConstructorParamter`), and return any annotations there, or default to null/false

A good example of the changes is in `KPropertyExtensionsKtTest`

-----------------------
Here is the changes I made to the example app code

screen shot 2018-12-13 at 11 36 27 am

And here is the docs in GraphiQL

screen shot 2018-12-13 at 11 36 04 am

### DIFF
--- a/example/src/main/kotlin/com.expedia.graphql.sample/model/Widget.kt
+++ b/example/src/main/kotlin/com.expedia.graphql.sample/model/Widget.kt
@@ -5,12 +5,12 @@ import com.expedia.graphql.annotations.GraphQLIgnore
 
 @GraphQLDescription("A useful widget")
 data class Widget(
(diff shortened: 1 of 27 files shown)


smyrick:
Ready for merge

PR #101, opened 2018-12-15
feat: move and label all exceptions

smyrick:


d4rken:
I like the improved annotations, but I'm not a fan of moving all exceptions into the same package.

i.e. I prefer "Package by feature"  over "Package by layer", if it's a specific schema generator exception it should be within the schema generator package.

smyrick:
@d4rken For developing an application I agree with you. But this library only has one feature exposed to users at the `com.expedia.graphql` package. So I thought it would make it easier. If we exposed all the exceptions at `com.expedia.graphql.exceptions` so that users don't have to worry about our internal folder/package structure

PR #108, opened 2018-12-20
fix: make deepName public

brennantaylor:
It's a very handy utility method


smyrick:
I just noticed that we use it in the SDK as well when I was updating.

1. You don't need to specify `public`

2. I want to flatten the package structure a bit and have more of the public things at the level of `com.expedia.graphql` and then everything internal to the generator at `com.expedia.graphql.generator` (basically dropping `.schema.`). Are you opposed to moving this to a new package at `com.expedia.graphql.extensions` as the first step?

PR #112, opened 2018-12-21
refactor: move top level packages

smyrick:
Now the top level packages are all the public parts of the library and everything under `/generator` is internal

Here is all the public methods and classes in Dokka (notice `com.expedia.graphql.generator` is not listed)
screen shot 2018-12-20 at 17 34 46 pm


smyrick:
I have more changes on top of this which simplify name exceptions and add more unit tests

PR #116, opened 2018-12-27
feat: allow directive with Enums

smyrick:
Kotlin annotations can have `KClass`, Enums, and arrays as a property type https://kotlinlang.org/docs/reference/annotations.html

However we can not convert KClass to a graphql type. Right now the generator actually breaks on a directive with an enum argument so this fixes that and excludes `KClass` as a valid type


src/test/kotlin/com/expedia/graphql/generator/types/DirectiveTypeBuilderTest.kt
```diff
@@ -0,0 +1,94 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+internal class DirectiveTypeBuilderTest {
+
+    @GraphQLDirective
+    internal annotation class SimpleDirective
+
+    @GraphQLDirective
+    internal annotation class DiretiveWithString(val string: String)
```
brennantaylor: typo

PR #120, opened 2018-12-28
refactor: simplify data fetcher

smyrick:
* Removes the `Parameter` class in favor of handling the `KParameter` which has all the same info.
* Since we don't need to construct `Parameter` objects, the `FunctionTypeBuilder` also drops complexity
* Add unit tests

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,2 +1,0 @@
-ignore:
-  - "suppressedExtenstions.kt"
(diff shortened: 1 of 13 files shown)


smyrick:
Let's merge this one first to mark the release of `0.0.29` https://github.com/ExpediaDotCom/graphql-kotlin/pull/119

smyrick:
Ready for merge

PR #133, opened 2019-01-04
test: add more hook tests

smyrick:
Add unit test coverage for `SchemaGeneratorHooks`


smyrick:
I cannot cover the on failure branch because that is technically unreachable code since we expect an exception to be thrown.

I think a better predicate is this https://github.com/ExpediaDotCom/graphql-kotlin/issues/134

src/test/kotlin/com/expedia/graphql/hooks/DataFetcherExecutionPredicateTest.kt
```diff
@@ -0,0 +1,51 @@
...
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.spyk
+import io.mockk.verify
+import org.junit.jupiter.api.Test
+import kotlin.reflect.KParameter
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+internal class DataFetcherExecutionPredicateTest {
```
dkuc84: depending on https://github.com/ExpediaDotCom/graphql-kotlin/issues/134 this may need to be completely rewritten

smyrick: @dkuc84 Actually these tests could just be deleted since we don't need to test an interface and we have implementation tests here already: https://github.com/ExpediaDotCom/graphql-kotlin/blob/ce018a9876c2de503b3847043992d71c70bf19cd/src/test/kotlin/com/expedia/graphql/KotlinDataFetcherTest.kt#L54

smyrick: The other hook tests would still be valid though

smyrick: You can see the changes required to fix #134 here: https://github.com/smyrick/graphql-kotlin/compare/hooks-test...smyrick:datafetcher-predicate?expand=1

dkuc84: lets drop this test then for now, we can reintroduce it later if needed

smyrick: Done

PR #137, opened 2019-01-04
Check if query and mutation classes are public

dkuc84:
Currently we are filtering non public properties and functions and don't expose them in the schema. We should have a consistent behavior for handling classes as well. With this change schema generator requires all query and mutation objects to have PUBLIC visibility modifier. Additionally, schema generator also won't include any non-public interfaces in the resulting schema.


dkuc84:
resolves https://github.com/ExpediaDotCom/graphql-kotlin/issues/135

PR #138, opened 2019-01-04
refactor: simplify predicate

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/134

Remove multiple abstract methods in favor of a single interface method


smyrick:
@brennantaylor We haven't released that way before but we can. I am using the labels now to easily see what release notes I need to write

https://github.com/ExpediaDotCom/graphql-kotlin/releases/tag/0.0.31

But we can merge still and next release can be 0.1.0

PR #142, opened 2019-01-08
build: travis commits the snapshot updates

smyrick:
With this change, on every tagged build, Travis will push a new branch back to GitHub that updates the pom in the library and the example to the snapshot of `tagVersion` + 0.0.1

We need to add the token as a secret key here: https://travis-ci.org/ExpediaDotCom/graphql-kotlin/settings which requires repo admin access

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .idea
 .DS_Store
 target
+pom.xml.versionsBackup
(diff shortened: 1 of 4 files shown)


smyrick:
I have updated the travis setting with the hook. We can now test this out

PR #148, opened 2019-01-14
Enable coroutine support in default KotlinDataFetcher

dkuc84:


dkuc84:
@smyrick if we convert the coroutine result to reactor mono/rxjava single it also means we would return different object than method signature. I think this could have some weird side effects

smyrick:
@dkuc84 Hmm, true. Then how will we support Reactor? Do we unwrap the type in the data fetcher?

dkuc84:
looks like `CompletableFuture` can be unwrapped correctly automatically

Issue #150, opened 2019-01-22
Optional results should return null values


d4rken:
I have a query where the result has optional values

e.g.
```kotlin
    data class SomeData(
        val requiredResult: String,
        val optionalResult: String? = null
    )
```

```grapqhl
type {
    optionalResult: String
    requiredResult: String!
}
```

Querying this gives me

```json
{
  "data": {
    "someQuery": {
      "requiredResult": "somedata"
    }
  },
  "errors": [],
  "dataPresent": true
}
```

Clientside, using apollo-client, I didn't get my data. I stumpled upon the same issue here:

https://github.com/apollographql/apollo-client/issues/4267

which suggests that the graphql spec forbids just omitting values and instead we should return:

```json
    "someQuery": {
      "requiredResult": "somedata",
      "optionalResult": null,
    }
```

d4rken:
Nvm this works correctly in graphql-kotlin.

d4rken:
The issue is my json serializer removing the null value.

PR #151, opened 2019-01-22
An integration test for optional results.

d4rken:
Showed me that my error wasn't with this lib.
See #150

We have lot's of modular tests already, a few more integration tests can't hurt :smile: .

### DIFF
--- a/src/test/kotlin/com/expedia/graphql/integration/OptionalResultsTest.kt
+++ b/src/test/kotlin/com/expedia/graphql/integration/OptionalResultsTest.kt
@@ -1,4 +1,4 @@
-package com.expedia
+package com.expedia.graphql.integration
 
 import com.expedia.graphql.TopLevelObject
 import com.expedia.graphql.testSchemaConfig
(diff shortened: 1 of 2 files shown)


smyrick:
@d4rken More tests are great, thank you! We have a pattern right now for the test package `com/expedia/graphql/integration/`. Can you more the file there and name it `OptionalResultsTest.kt`

d4rken:
Done. I missed that there was an existing structure for this :sweat_smile:

src/test/kotlin/com/expedia/graphql/integration/OptionalResultsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.expedia.graphql.TopLevelObject
+import com.expedia.graphql.testSchemaConfig
+import com.expedia.graphql.toSchema
+import graphql.GraphQL
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+@Suppress("Detekt.UnusedPrivateMember",
```
dkuc84: i think some of those suppressions don't apply?

smyrick:
Not sure why it is reporting a drop. LGTM

PR #152, opened 2019-01-23
 Enable support for specifying custom DataFetchers

dkuc84:


src/main/kotlin/com/expedia/graphql/execution/KotlinDataFetcherFactoryProvider.kt
```diff
@@ -0,0 +1,43 @@
...
+                        objectMapper = defaultObjectMapper,
+                        executionPredicate = hooks.dataFetcherExecutionPredicate))
+
+    /**
+     * Retrieve instance of [DataFetcherFactory] that will be used to resolve target property.
+     *
+     * @param kClazz parent class that contains this property
+     * @param kProperty Kotlin property that should be resolved
+     */
+    open fun propertyDataFetcherFactory(kClazz: KClass<*>, kProperty: KProperty<*>): DataFetcherFactory<Any> =
```
smyrick: nit but we can have the variable be called `kClass: KClass<*>`

PR #164, opened 2019-02-15
Update library versions

smyrick:
Also add some very small changes that I have in my subscriptions branch that have no affect on behavior

### DIFF
--- a/detekt.yml
+++ b/detekt.yml
@@ -1,6 +1,3 @@
-# turn on ALL checks
-failFast: true
-
 build:
   weights:
     comments: 0   # missing comments should not fail the build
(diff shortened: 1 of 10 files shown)


detekt.yml
```diff
@@ -1,6 +1,3 @@
-# turn on ALL checks
-failFast: true
```
dkuc84: did detekt change the default? we had to explicitly set it to true in the past

smyrick: Yes, it was giving me and warning on builds saying to use the CLI flag instead

dkuc84: https://github.com/arturbosch/detekt/pull/1479

PR #166, opened 2019-02-15
fix: move the default handling of CompletableFuture

smyrick:
This will have no impact on existing clients. I just moved the handling of `CompletableFuture`s into the code so that it will always happen. Right now, clients need to remember to still handle `CompletableFuture`s if they override the hook.

We now resolve the type from the hook first, which defaults to the same type. Then we check after if it is on of the default supported wrapped types, which right now only includes `CompletableFuture`, but will extend to `Publisher` when we support subscriptions

### DIFF
--- a/src/main/kotlin/com/expedia/graphql/generator/types/FunctionTypeBuilder.kt
+++ b/src/main/kotlin/com/expedia/graphql/generator/types/FunctionTypeBuilder.kt
@@ -6,14 +6,17 @@ import com.expedia.graphql.generator.TypeBuilder
 import com.expedia.graphql.generator.extensions.getDeprecationReason
 import com.expedia.graphql.generator.extensions.getGraphQLDescription
 import com.expedia.graphql.generator.extensions.getName
+import com.expedia.graphql.generator.extensions.getTypeOfFirstArgument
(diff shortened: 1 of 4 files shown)


smyrick:
This is a patch change since all existing client code will still work and we are not adding any new features.

src/main/kotlin/com/expedia/graphql/generator/types/FunctionTypeBuilder.kt
```diff
@@ -43,12 +46,20 @@ internal class FunctionTypeBuilder(generator: SchemaGenerator) : TypeBuilder(gen
...
+        builder.type(graphQLTypeOf(returnType) as GraphQLOutputType)
         val graphQLType = builder.build()
-        return config.hooks.onRewireGraphQLType(monadType, graphQLType) as GraphQLFieldDefinition
+
+        return config.hooks.onRewireGraphQLType(returnType, graphQLType) as GraphQLFieldDefinition
     }
 
+    private fun getWrappedReturnType(returnType: KType): KType =
+        when (returnType.classifier) {
+            CompletableFuture::class -> returnType.getTypeOfFirstArgument()
```
smyrick: Checking for `Publisher` will go here

PR #167, opened 2019-02-15
feat: standardize casting of GraphQLType

smyrick:
Standardize the casting so we throw exceptions and have better coverage and logging.

Minor change due to the new exception that was added.


src/main/kotlin/com/expedia/graphql/exceptions/CouldNotCastGraphQLType.kt
```diff
@@ -0,0 +1,8 @@
+package com.expedia.graphql.exceptions
+
+import graphql.schema.GraphQLType
+
+/**
+ * Thrown when the casting a GraphQLType to some parent type is invalid
+ */
+class CouldNotCastGraphQLType(type: GraphQLType) : GraphQLKotlinException("Could not cast GraphQLType $type")
```
dkuc84: can we capture info on what we tried to cast it to? think it would provide more meaningful information

smyrick: 👍

PR #168, opened 2019-02-18
Add JDK11 build step

gscheibel:


smyrick:
We are having build issues with Dokka on JDK 11. It looks like it is unsupported right now https://github.com/Kotlin/dokka/issues/294

PR #171, opened 2019-02-20
Add example of subquery

gscheibel:


d4rken:
What do you think about including the schema that is generated by the examples within the file as comment?

smyrick:
@gscheibel Made a good comment that the example app is meant to be run so the resulting schema can easily be looked at in graphiQL, which is a little easier too.

I am going to merge. If we want to add more docs on what code will produce a schema we can do that in the wiki. I wouldn't expect anyone to look through the code for these comments.

PR #173, opened 2019-02-21
feat: pass DataFetcherEnvironment to arguments

smyrick:
Clients can now specify `DataFetchingEnvironment` as one of their arguments. It will not be generated as part of the schema but will be passed down on execution and gives access to all the context of the query, parent fields and their arguments, and anything else that is on the `graphql.schema.DataFetchingEnvironment`

This is the example query I have added
```kotlin
@Component
class EnvironmentQuery : Query {
    fun nestedEnvironment(value: Int) = NestedNode(value)
}

data class NestedNode(
    val value: Int,
    val parentValue: Int? = null
) {
    fun nested(environment: DataFetchingEnvironment, value: Int): NestedNode {
        val parentValue: Int? = if (environment.executionStepInfo.hasParent()) environment.executionStepInfo.parent.getArgument("value") else null

        return NestedNode(value, parentValue)
    }
```

screen shot 2019-02-21 at 12 12 11 pm


example/src/main/kotlin/com/expedia/graphql/sample/query/EnvironmentQuery.kt
```diff
@@ -0,0 +1,20 @@
...
+class EnvironmentQuery : Query {
+    fun nestedEnvironment(value: Int) = NestedNode(value)
+}
+
+data class NestedNode(
+    val value: Int,
+    val parentValue: Int? = null
+) {
+    fun nested(environment: DataFetchingEnvironment, value: Int): NestedNode {
+        val parentValue: Int? = if (environment.executionStepInfo.hasParent()) environment.executionStepInfo.parent.getArgument("value") else null
```
brennantaylor: multi line please

src/main/kotlin/com/expedia/graphql/generator/extensions/kParameterExtensions.kt
```diff
@@ -2,13 +2,17 @@ package com.expedia.graphql.generator.extensions
...
+import graphql.schema.DataFetchingEnvironment
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
+import kotlin.reflect.full.isSuperclassOf
 
 internal fun KParameter.isInterface() = this.type.getKClass().isInterface()
 
 internal fun KParameter.isGraphQLContext() = this.findAnnotation<GraphQLContext>() != null
 
+internal fun KParameter.isDataFetchingEnvironment() = this.type.getKClass().isSuperclassOf(DataFetchingEnvironment::class)
```
brennantaylor: Is this the correct direction? Shouldn't the parameter be the same or a subclass of `DataFetchingEnvironment`?

DFE is just an interface so we won't allow people to specify subclasses?

DFE also doesn't inherit anything so this check is basically is `this.type == DFE`

smyrick:
@brennantaylor Comments addressed

PR #190, opened 2019-03-08
Update example app to use Spring Webflux and GraphQL Playground

dkuc84:
* example app is no longer using any Servlet based API (i.e. `graphql-java-servlet` and `graphiql-spring-boot-starter`)
* example app is now using Spring Webflux and Netty for non-blocking server
* introduce GraphQL Playground (https://github.com/prisma/graphql-playground) as a replacement for GraphiQL


example/src/main/kotlin/com/expedia/graphql/sample/query/RecursiveQuery.kt
```diff
@@ -1,22 +1,22 @@
...
-
-    init {
-        root.children = listOf(nodeA, nodeB)
-        nodeB.children = listOf(nodeC)
-    }
-
-    @GraphQLDescription("Returns the root of a node graph")
-    fun nodeGraph(): Node = root
-}
+//package com.expedia.graphql.sample.query
```
smyrick: Let's uncomment the test. This will be fixed here: https://github.com/ExpediaDotCom/graphql-kotlin/pull/192

smyrick: Let's uncomment the test. This will be fixed here: https://github.com/ExpediaDotCom/graphql-kotlin/pull/192

PR #205, opened 2019-04-05
fix: allow for abstract properties and functions

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/204

### DIFF
--- a/src/test/kotlin/com/expedia/graphql/generator/extensions/KClassExtensionsTest.kt
+++ b/src/test/kotlin/com/expedia/graphql/generator/extensions/KClassExtensionsTest.kt
@@ -127,7 +127,7 @@ open class KClassExtensionsTest {
     }
 
     @Test
-    fun `test getting functions from absctract classes`() {
+    fun `test getting functions from abstract classes`() {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/com/expedia/graphql/generator/extensions/KClassExtensionsTest.kt
```diff
@@ -102,6 +126,12 @@ open class KClassExtensionsTest {
         assertEquals(listOf("publicFunction"), properties.map { it.name })
     }
 
+    @Test
+    fun `test getting functions from absctract classes`() {
```
tapaderster: typo
```suggestion
    fun `test getting functions from abstract classes`() {
```

smyrick: Nice, the suggestion tool is helpful

tapaderster: ohh you can directly commit from it. even better

PR #209, opened 2019-04-08
build: migrate to JDK 11

smyrick:
JDK 11 is the new long term support version


smyrick:
@dkuc84 Yea, you are right it is still failing. We will need to keep most of this for now.

PR #211, opened 2019-04-09
Support custom names via @GraphQLName

sgeb:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/210

I went ahead and packaged my local changes into a PR, even before hearing back whether it's desirable in upstream. It might be useful to somebody else, regardless of whether it gets merged. Thanks.

### DIFF
--- a/src/main/kotlin/com/expedia/graphql/annotations/GraphQLName.kt
+++ b/src/main/kotlin/com/expedia/graphql/annotations/GraphQLName.kt
@@ -1,0 +1,7 @@
+package com.expedia.graphql.annotations
+
+/**
+ * Set the GraphQL name to be picked up by the schema generator.
+ */
+@Target(AnnotationTarget.CLASS)
+annotation class GraphQLName(val value: String)
--- a/src/main/kotlin/com/expedia/graphql/generator/extensions/annotationExtensions.kt
+++ b/src/main/kotlin/com/expedia/graphql/generator/extensions/annotationExtensions.kt
@@ -3,10 +3,13 @@ package com.expedia.graphql.generator.extensions
 import com.expedia.graphql.annotations.GraphQLDescription
 import com.expedia.graphql.annotations.GraphQLID
 import com.expedia.graphql.annotations.GraphQLIgnore
+import com.expedia.graphql.annotations.GraphQLName
 import kotlin.reflect.KAnnotatedElement
 import kotlin.reflect.full.findAnnotation
 
 internal fun KAnnotatedElement.getGraphQLDescription(): String? = this.findAnnotation<GraphQLDescription>()?.value
+
+internal fun KAnnotatedElement.getGraphQLName(): String? = this.findAnnotation<GraphQLName>()?.value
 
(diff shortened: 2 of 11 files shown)


src/main/kotlin/com/expedia/graphql/generator/types/EnumBuilder.kt
```diff
@@ -13,7 +14,7 @@ internal class EnumBuilder(generator: SchemaGenerator) : TypeBuilder(generator)
     internal fun enumType(kClass: KClass<out Enum<*>>): GraphQLEnumType {
         val enumBuilder = GraphQLEnumType.newEnum()
 
-        enumBuilder.name(kClass.getSimpleName())
+        enumBuilder.name(kClass.getGraphQLName() ?: kClass.getSimpleName())
```
smyrick: Can we move the logic behind `getSimpleName` so that we only have to update one location?

smyrick:
@sgeb This is another good feature. We can add this in. I would hope that developers don't use it too much so that the code and schema line up but I can see some instances where you want to rename and existing class

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/210

src/main/kotlin/com/expedia/graphql/generator/types/EnumBuilder.kt
sgeb: Yes, you're right, that'd be better. As a pleasant side-effect, it would solve an issue I found earlier today related to building types already under construction: see `kClass.getSimpleName()` in https://github.com/ExpediaDotCom/graphql-kotlin/blob/ceb6b9a162f6956bf5a1c05e1e4e8d00129c9ec3/src/main/kotlin/com/expedia/graphql/generator/state/TypesCache.kt#L77

What's customary in this project: add a new commit or fixup with push-force to the PR?

smyrick: Either is fine, Github will squash all the commits anyway when we merge to master

sgeb:
I'm seeking opinions on the following: when a class is used as an input, should the name provided in`@GraphQLName`
1. override the complete type name, or
2. override the "base" type name and always have `Input` appended?

Currently my change implements 1. but I'm questioning whether that's actually a good choice. Opinions welcome, thanks.

smyrick:
@sgeb 

One thing we have to keep in mind is that a class can still be used as both input and output.

ie
```kotlin
@GraphQLName("CustomName")
data class NumberRange(val start: Int, val end: Int)
```

When I use in the following what would you expect the schema to be?

```kotlin
class Query {
  fun addOneToEnd(inputRange: NumberRange) = inputRange.copy { end = it.end + 1 } 
}
```

This would be an invalid schema since the input type would clash
```graphql
type Query {
   addOneToEnd(inputRange: CustomName!): CustomName!
}

input CustomName {
  start: Float!
  end: Float!
}

type CustomName {
  start: Float!
  end: Float
}
```

sgeb:
That was the background of my question. With the current implementation it's not possible to use the same class as input and as returned type. I'll change to option 2, in line with the default implementation without `@GraphQLName`.

PR #215, opened 2019-05-06
feat: subscriptions

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/155

Example using a Websocket test client

Screen Shot 2019-05-05 at 8 33 26 PM


example/src/main/kotlin/com/expedia/graphql/sample/GraphQLRequest.kt
```diff
@@ -1,7 +1,18 @@
...
+import com.expedia.graphql.sample.context.MyGraphQLContext
+import graphql.ExecutionInput
+
 data class GraphQLRequest(
     val query: String,
     val operationName: String? = null,
     val variables: Map<String, Any>? = null
 )
+
+fun GraphQLRequest.getExecutionInput(graphQLContext: MyGraphQLContext? = null): ExecutionInput =
```
dkuc84: since you are converting it to execution input maybe rename it to `to` instead of `get`?

smyrick: 👍 I had it as an external method and then moved it. I will rename

PR #225, opened 2019-05-20
FunctionDataFetch throws `InvocationException.cause` property when available

rharriso:
This allows thrown `GraphQLException` instances to be handled by
`graph-java` as it would if thrown directly by a `DataFetcher` lambda.

---

I'm working on a prototype for an API gateway, a snippet of my login resolve is as follows

```kotlin
        val body = /* Usernam and password json object*/
        client.post(/* port */, /* domain */, /* path */)
            .putHeader("Accept", "application/json")
            .putHeader("Content-Type", "application/json")
            .sendJsonObject(body, {ar -> tokenFuture.complete(ar); })

        val response = tokenFuture.get()
        val result = response.result()

        if (result.statusCode() != 200) {
            throw GraphQLException("Username and password combination failed.")
        }
```

When Implementing directly with Graphql-Java's `DataFetchers` the error messages would appear as expected.

```json
{
  "data": null,
  "errors": [
    {
      "message": "Exception while fetching data (/login) : Username and password combination failed.",
      "locations": [
        {
          "line": 6,
          "column": 3
        }
      ],
      "path": [
        "login"
      ]
    }
  ]
}
```

However, when Implemented using this library  it would get null for the exception details.

```json
{
  "data": null,
  "errors": [
    {
      "message": "Exception while fetching data (/login) : null",
      "locations": [
        {
          "line": 6,
          "column": 3
        }
      ],
      "path": [
        "login"
      ]
    }
  ]
}
```

---

Update:

I've added tests, and remove changes to the async branch. The message is already included in the `ExecutionException` that is thrown by that branch. And I didn't want to add another layer of synchronization just to clean up the error messages.

### DIFF
--- a/src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
+++ b/src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
@@ -37,7 +37,7 @@ internal class FunctionDataFetcherTest {
 
         fun throwException() { throw GraphQLException("Test Exception") }
 
-        suspend fun suspendThrow(): CompletableFuture<*> = coroutineScope {
+        suspend fun suspendThrow(): String = coroutineScope {
             delay(10)
             throw GraphQLException("Suspended Exception")
         }
--- a/src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
+++ b/src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
@@ -13,7 +13,9 @@ import java.util.concurrent.CompletableFuture
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 import kotlin.test.assertNull
+import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
+import kotlin.test.assertFalse
 
 internal class FunctionDataFetcherTest {
 
--- a/src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
+++ b/src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
@@ -35,7 +35,7 @@ internal class FunctionDataFetcherTest {
             string
         }
 
-        fun throwException () { throw GraphQLException("Test Exception") }
+        fun throwException() { throw GraphQLException("Test Exception") }
 
         suspend fun suspendThrow(): CompletableFuture<*> = coroutineScope {
             delay(10)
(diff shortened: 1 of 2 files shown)


smyrick:
This seems like a good add to me.

I added this in the example to debug and you are correct. It does show up as null. I think we can forward these exceptions.

```
fun throwsIfTrue(throwException: Boolean): String {
        if (throwException) {
            throw MyException("Passed in a true value")
        }

        return "Hello!"
    }
```


image

rharriso:
@smyrick I've added tests and updated the description.

rharriso:
@smyrick Are there any other changes you'd like to see?

src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
```diff
@@ -31,6 +34,13 @@ internal class FunctionDataFetcherTest {
             delay(10)
             string
         }
+
+        fun throwException() { throw GraphQLException("Test Exception") }
+
+        suspend fun suspendThrow(): CompletableFuture<*> = coroutineScope {
```
dkuc84: when using coroutines the whole point would be to avoid `CompletableFuture`, lets just change it to `String`

src/test/kotlin/com/expedia/graphql/execution/FunctionDataFetcherTest.kt
```diff
@@ -126,4 +136,34 @@ internal class FunctionDataFetcherTest {
...
+    }
+
+    @Test
+    fun `suspendThrow throws exception when resolved`() {
+        val dataFetcher = FunctionDataFetcher(target = MyClass(), fn = MyClass::suspendThrow)
+        val mockEnvironmet: DataFetchingEnvironment = mockk()
+
+        try {
+            val result = dataFetcher.get(mockEnvironmet)
+            assertTrue(result is CompletableFuture<*>)
```
dkuc84: while test still works it only works so because it throws exception in line above, can we remove line 160 and 161 as they are leftovers from `CompletableFuture` return type?

rharriso: It doesn't look like it. When debugging I see the type of `result` as `CompletableFuture` still. Without checking the type, the compiler won't let me call `get`. Removing 161 causes the assertion on line 162 to fail.

Is there a more idiomatic way this should be done?

dkuc84: Ah indeed! We are wrapping coroutine results in completable future so we can return it async (so your return type in this case is `CompletableFuture<String>`). Totally forgot about it.

PR #232, opened 2019-06-06
feat: new hook for modifying object types

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/231

Adds a new hook that can be used to modify all the GraphQLTypes that get generated from reflection. This allows for runtime documentation or any other changes that may want to be made to types after they are fully generated


rharriso:
Is there a use case for this besides adding markdown documentation? If not, should it be part of the public API?

PR #242, opened 2019-06-21
upgrade to graphql-java 13

dkuc84:
Upgrade to use `graphql-java` version 13.

Besides updating to use new `CodeRegistry` for defining all data fetchers and type resolvers this PR completely rewrites how we are handling directives.

Directive related changes
* fix: directives are now applied on all the GraphQL types
* fix: default directive annotation locations are now just SDL locations
* fix: before applying the directives we now validate whether directive is applicable to the target location
* fix: require all the defined directives to have a corresponding wiring

Resolves: https://github.com/ExpediaDotCom/graphql-kotlin/issues/216


src/test/kotlin/com/expedia/graphql/KotlinSchemaDirectiveWiringHelperTest.kt
```diff
@@ -0,0 +1,246 @@
+// package com.expedia.graphql
```
rharriso: Is this WIP?

example/src/main/kotlin/com/expedia/graphql/sample/directives/StringEvalSchemaDirectiveWiring.kt
```diff
@@ -34,17 +28,19 @@ class StringEvalDirectiveWiring : DirectiveWiring {
                     newArguments[graphQLArgumentType.name] = graphQLArgumentType.defaultValue
                 }
             }
-            val newEnv = DataFetchingEnvironmentBuilder.newDataFetchingEnvironment(dataEnv)
+            // NOTE: this relies on internal graphql-java API and may break in the future
```
rharriso: Are there any portions of the library code that also rely in `graphql-java` internals? That, in part, is what broke compatibility with v12

src/test/kotlin/com/expedia/graphql/KotlinSchemaDirectiveWiringHelperTest.kt
smyrick: Yes, this PR is in progress. Any feedback about the architecture changes is appreciated. We could have used the draft pr feature but this is fine

smyrick:
Deleting `DirectiveWiringHelper` is a breaking change. This is a major version change. We may be able to bundle this as the `1.0.0` release, we can do a `1.0.0-RC1` if we want to test the changes too

example/src/main/kotlin/com/expedia/graphql/sample/directives/StringEvalSchemaDirectiveWiring.kt
dkuc84: `graphql-kotlin` builds on top of `graphql-java` so this comment doesn't really make much sense and I'll remove it.

dkuc84:
@smyrick while this is fairly large change (especially with regards to how we handle directives) I'm not sure it warrants 1.0.0 release...

src/test/kotlin/com/expedia/graphql/directives/KotlinSchemaDIrectiveWiringTest.kt
```diff
@@ -0,0 +1,135 @@
+// package com.expedia.graphql.directives
```
smyrick: @dkuc84 Can you uncomment this file?

PR #245, opened 2019-06-26
feat: support data and errors with DataFetcherResult

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/244

Kotlin functions can now return a `DataFetcherResult` instead of just their return type which allows you to modify the errors field with any extra data you need


dkuc84:
While this could work it would also kind of imply that you would want to explicitly return `DataFetcherResult` pretty much everywhere. Would it be possible to support it in more generic way?

smyrick:
@dkuc84 What would be that more generic way? Do you mean some additional option to add info into the errors field but without wrapping?

I briefly looked into this. Is there some argument that we can include in the data fetchers or from the `DataFetchingEnvironmnet` that allows you to modify the errors? I haven't seen anything so far: https://www.graphql-java.com/documentation/v13/data-fetching/

dkuc84:
I'm thinking we could do something around adding errors to the GraphQL context but that would also imply we would need to either add some instrumentation to populate errors (workaround) or re-implement execution logic so it would take that into account.

smyrick:
I am going to close this PR until we have a solid decision on how we want to support data and errors

dkuc84:
@smyrick is this going to work with monads (e.g. `CompletableFuture`)?

PR #251, opened 2019-07-08
Revert throwing exception on missing directive wiring

dkuc84:
There are use cases when directives could be just informational (e.g. @deprecated). Instead of defining no-op directive wiring we should just ignore directives without wiring.

### DIFF
--- a/src/main/kotlin/com/expedia/graphql/directives/KotlinDirectiveWiringFactory.kt
+++ b/src/main/kotlin/com/expedia/graphql/directives/KotlinDirectiveWiringFactory.kt
@@ -10,11 +10,6 @@ import graphql.schema.GraphQLFieldDefinition
 import graphql.schema.GraphQLType
 
 /**
- * Default no-op wiring for deprecated directive.
- */
-private val defaultDeprecatedWiring = object : KotlinSchemaDirectiveWiring {}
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/expedia/graphql/directives/KotlinDirectiveWiringFactory.kt
```diff
@@ -64,23 +59,15 @@ open class KotlinDirectiveWiringFactory(
             val directiveWiring = discoverWiringProvider(directive.name, env)
             if (directiveWiring != null) {
                 modifiedObject = directiveWiring.wireOnEnvironment(env)
-            } else {
-                throw InvalidSchemaDirectiveWiringException("No directive wiring provided for ${directive.name}")
```
smyrick: Is this exception used elsewhere? Can we delete it?

dkuc84: its used few lines above

smyrick: 🤦‍♂ Sorry, reviewing on mobile

PR #260, opened 2019-07-18
limit the branches for travis

dkuc84:


.travis.yml
```diff
@@ -5,6 +5,11 @@ jdk:
   - openjdk8
   - openjdk11
 
+branches:
+  only:
```
smyrick: Both of these changes will need to be pushed to the `0.6.x` branch since travis pulls the config from GH

PR #263, opened 2019-07-18
fix travis branch regex

dkuc84:


smyrick:
Nice we have PR checks now, so that's a good sign

PR #268, opened 2019-07-18
publish only the schema-generator module

smyrick:
Set the profiles for release and javadoc to only run on the schema-generator module as the example app doesn't need to be published

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,11 @@
         <url>https://github.com/ExpediaDotCom/graphql-kotlin</url>
     </scm>
 
+    <modules>
+        <module>graphql-kotlin-schema-generator</module>
+        <module>graphql-kotlin-spring-example</module>
+    </modules>
+
     <properties>
         <!-- Project Versions -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -337,6 +342,11 @@
                 </property>
             </activation>
 
+            <!-- Exclude the example app from releases -->
+            <modules>
+                <module>graphql-kotlin-schema-generator</module>
```
dkuc84: it should already be handled by -> https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/graphql-kotlin-spring-example/pom.xml#L81

pom.xml
```diff
@@ -411,6 +421,12 @@
             <activation>
                 <jdk>1.8</jdk>
             </activation>
+
+            <!-- Exclude the example app from javadoc packages -->
+            <modules>
+                <module>graphql-kotlin-schema-generator</module>
```
dkuc84: and this should be the same as -> https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/graphql-kotlin-spring-example/pom.xml#L89

pom.xml
dkuc84: nvm, release runs `nexus-staging-maven-plugin` during deploy phase which was not skipped.

PR #279, opened 2019-07-24
Add debug logging to update-versions.sh

smyrick:
Add some more logging to the script because running the steps locally performs the correct changes

### DIFF
--- a/.travis/update-version.sh
+++ b/.travis/update-version.sh
@@ -28,6 +28,9 @@ commit_files() {
     # Stage the modified files
     git add ./pom.xml ./*/pom.xml
 
+    # Print the current files we are going to commit
+    git status
+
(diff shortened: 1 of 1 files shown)


.travis/update-version.sh
```diff
@@ -1,47 +1,50 @@
...
     NEW_VERSION=$(mvn --settings .travis/settings.xml help:evaluate -Dexpression=project.version -q -DforceStdout)
 }
 
 commit_files() {
+    echo "Commiting the changes to git"
 
-  # Use the version as the branch name
-  git checkout -b ${NEW_VERSION}
+    # Use the version as the branch name
+    git checkout -b ${NEW_VERSION}
```
dkuc84: do we want to print git status afterwards?

smyrick: Done

Issue #238, opened 2019-06-10
Apollo Federation Support


de55:
**Is your feature request related to a problem? Please describe.**
Support for generating schema as federated service. Federated services can be used by an apollo gateway to stitch together schemas from multiple graphql endpoints.

**Describe the solution you'd like**
Support generation of features in federation spec: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

**Describe alternatives you've considered**
N/A

**Additional context**
I use this project at work and want the benefits of a federated service. I am finishing up a different project now, but can start working on this in a few weeks. I wanted to check to see if you thought these additions would be an appropriate fit for this project.

dkuc84:
Thanks for creating the issue. This is indeed on our backlog but unfortunately it is currently blocked by https://github.com/ExpediaDotCom/graphql-kotlin/issues/216

de55:
Looks like 216 is closed now. You mentioned it is on the backlog, is this something that is expected to be implemented soon?

de55:
@dkuc84 Looks like 216 is closed now. You mentioned it is on the backlog, is this something that is expected to be implemented soon?

dkuc84:
I briefly looked into it (my WIP branch -> https://github.com/dkuc84/graphql-kotlin/tree/federation) to see how much work it would take to support federation and unfortunately it requires larger effort. Currently it looks like I won't be able to work on it anytime soon so feel free to pick it up.

Apollo released sample [federation-jvm](https://github.com/apollographql/federation-jvm) project that is an extension to `graphql-java`. Logic there applies transformation to an existing schema and relies on users providing manual resolver logic for the federated types. It could technically work but what is currently missing is the "glueing" part to generate federated query objects using `graphql-kotlin` - adding support for directives is pretty simple but what is problematic is that those federated queries are NOT part of query graph (i.e. they are not reachable from root queries). 

I'm thinking we would probably need to extend `SchemaGenerator` from `graphql-kotlin-schema-generator` that would take as an input list of those federated query objects (what would be problematic though is how those objects get instantiated).

formanojhr:
Does this enhancement help with Grapql-java to support apollo federation support?

smyrick:
@formanojhr Our implementation is directed at being used with the schema generation of GraphQL Kotlin Schema Generator

If you want a simple addition to GraphQL Java you can use https://github.com/apollographql/federation-jvm/blob/master/README.md

PR #280, opened 2019-07-25
Implementing federation spec

dkuc84:
Create new `graphql-kotlin-federation` module that extends `graphql-kotlin-schema-generator` to generate federated GraphQL schemas. Federated GraphQL architecture is an alternative to schema stitching approach that relies on declarative programming composition of schemas.

`graphql-kotlin-federation` adds support for the following:
- new `@key`, `@extends`, `@external`, `@requires` and `@provides` directives
- new `_FieldSet` and `_Any` scalar types
- new `_Entity` union type of all federated types
- new `_service` query that returns `_Service` object with `sdl` field
- new `_entity` query that returns list of `_Entity` objects 

In order to generate federate schemas we have to annotate the underlying type with proper new directives and then use new `toFederatedSchema` function to build the schema (instead of `toSchema` function from `graphql-kotlin-schema-generator`). `FederatedSchemaGeneratorHooks` also require to specify `FederatedTypeRegistry` that holds a mapping between federated type names and their corresponding resolver. Those new `FederatedTypeResolver`s are used to instantiate the federated types based on the federated `_entity` query generated from the gateway.

See: https://www.apollographql.com/docs/apollo-server/federation/introduction/

Resolves: #238


graphql-kotlin-federation/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    <name>graphql-kotlin-federation</name>
+    <packaging>jar</packaging>
+
+    <build>
+        <sourceDirectory>${project.basedir}/src/main/kotlin</sourceDirectory>
+        <testSourceDirectory>${project.basedir}/src/test/kotlin</testSourceDirectory>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
```
pcarrier: Does this do something I cannot find?

graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederationSchemaGeneratorHooks.kt
```diff
@@ -0,0 +1,123 @@
...
+            .asSequence()
+            .filterIsInstance<GraphQLObjectType>()
+            .filter { type -> type.getDirective("key") != null }
+            .onEach { type -> type.validateFederatedType() }
+            .map { it.name }
+            .toSet()
+
+        // register new federated queries
+        if (entityTypeNames.isNotEmpty()) {
+            federatedQuery.field(SERVICE_FIELD_DEFINITION)
```
pcarrier: A service could be federated without exposing any entities, eg to provide new fields on the query or mutation root types.

graphql-kotlin-federation/pom.xml
dkuc84: It applies detekt and ktlint plugins from the parent (https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/pom.xml#L138). Cannot apply them at the root project as it doesn't have any source code and there are no official maven plugins that would work out of the box.

graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederationSchemaGeneratorHooks.kt
dkuc84: nice catch! thanks

graphql-kotlin-federation/src/test/kotlin/com/expedia/graphql/federation/FederatedSchemaGeneratorTest.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    @Test
+    fun `verify generator does not add federation queries for non-federated schemas`() {
+        val expectedSchema = """
+            schema {
+              query: Query
+            }
+
+            type Query {
+              _service: _Service
```
gscheibel: If the service is not "federate-able" should its SDL contain `_service`?

dkuc84: yes -> https://github.com/ExpediaDotCom/graphql-kotlin/pull/280#discussion_r309385781

graphql-kotlin-federation/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+```groovy
+compile(group: 'com.expedia', name: 'graphql-kotlin-federation', version: "$latestVersion")
+```
+
+## Usage
+
+In order to generate valid federated schemas, you will need to annotate your both base service and the one extending it. Federated Gateway (e.g. Apollo) will then combine the individual graphs to form single federated graph.
+
+#### Base Service
```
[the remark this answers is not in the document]
dkuc84: i don't think that name matters - base service is just a service that defines base type, i'll update it to be more specific

graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/directives/ProvidesDirective.kt
```diff
@@ -0,0 +1,59 @@
...
+ *   @ProvidesDirective(FieldSet("name"))
+ *   fun user(): User = // implementation goes here
+ * }
+ *
+ * @KeyDirective(FieldSet("userId"))
+ * @ExtendsDirective
+ * class User(
+ *   @property:ExternalDirective val userId: String,
+ *   @property:ExternalDirective val name: String
+ * )
```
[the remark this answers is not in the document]
dkuc84: it would result in invalid schema

PR #287, opened 2019-07-31
fix: directives with arguments should be created per declaration

dkuc84:
Directive definitions should only be added to the schema once - while building the schema we are caching the directive definitions and add them after processing all the objects. Our current logic relies on this cache to not-rebuild the directives per each one of its declarations. Directives can have arguments which means that we have to rebuild the directive definition if they accept any arguments.

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/SchemaGeneratorState.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/SchemaGeneratorState.kt
@@ -4,21 +4,24 @@ import com.expedia.graphql.directives.DeprecatedDirective
 import graphql.Directives
 import graphql.schema.GraphQLDirective
 import graphql.schema.GraphQLType
+import java.util.concurrent.ConcurrentHashMap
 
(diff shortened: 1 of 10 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/SchemaGeneratorState.kt
```diff
@@ -4,21 +4,24 @@ import com.expedia.graphql.directives.DeprecatedDirective
 import graphql.Directives
 import graphql.schema.GraphQLDirective
 import graphql.schema.GraphQLType
+import java.util.concurrent.ConcurrentHashMap
 
 internal class SchemaGeneratorState(supportedPackages: List<String>) {
     val cache = TypesCache(supportedPackages)
     val additionalTypes = mutableSetOf<GraphQLType>()
-    val directives = mutableSetOf<GraphQLDirective>()
+    val directives = ConcurrentHashMap<String, GraphQLDirective>()
```
dkuc84: updated to `ConcurrentHashMap` instead of just simple `HashMap` as by default maven surefire plugin runs test in parallel which lead to some concurrency issues

PR #288, opened 2019-08-01
Support GraphQLDataFetcherPrefix Annotation

huangciyin:
support add prefix to datafeter.

example:

```
@GraphQLDataFetcherPrefix("example_prefix_")
class QueryObjectWithPrefix: Query {
   fun query1(value: Int) :Int{
        ...
    }

    fun query2(value: Int) :Int{
        ...
    }

    fun query3(value: Int) :Int{
        ...
    }
}
```

The above query would produce the following GraphQL schema:
```
schema {
  query: Query
}

type Query {
  example_prefix_query1(id: Int!): Int
  example_prefix_query2(id: Int!): Int
  example_prefix_query3(id: Int!): Int
}
```

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/annotations/GraphQLDataFetcherPrefix.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/annotations/GraphQLDataFetcherPrefix.kt
@@ -1,0 +1,7 @@
+package com.expedia.graphql.annotations
+
+/**
+ * Set the GraphQL DataFetcher Prefix to be picked up by the schema generator.
+ */
+@Target(AnnotationTarget.CLASS)
+annotation class GraphQLDataFetcherPrefix(val value: String)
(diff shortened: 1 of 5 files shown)


smyrick:
@huangciyin Thank you for providing a contribution. It seems what you are trying to accomplish is change the name of the GraphQL fields from the Kotlin functions. We have an annotation today to already do this, `@GraphQLName`: https://github.com/ExpediaDotCom/graphql-kotlin/wiki/Fields#documenting-fields

While this changes the entire name, and does not prepend a prefix like your changes, I don't see the need to include this feature. Ideally you shouldn't even need to use `@GraphQLName` as that is now changing the code from what developers see. If you want to have a naming convention, you can enforce that with code linters or just naming the functions that way.

We are going to close this PR. If you would like to start a discussion on a new feature request you can create a new issue with the feature tag: https://github.com/ExpediaDotCom/graphql-kotlin/issues/new/choose

huangciyin:
This pr support namespace through prefix on datafetcher/resolver function.
If we add an prefix annotation on Query/Mutation Class , all the data fetcher function in this class will be renamed through the prefix, which can support namespace on graphql schema.

https://github.com/graphql/graphql-spec/issues/163

huangciyin:
issue: https://github.com/ExpediaDotCom/graphql-kotlin/issues/289

PR #290, opened 2019-08-03
fix: remove example app tests

smyrick:
Remove the tests folder from the example app as this is causing coverage reports to drop in codecov

Screen Shot 2019-08-02 at 5 07 11 PM


smyrick:
☝️ Back up to `96.85%`

dkuc84:
Instead of removing tests from example app. I'd rather add some integration tests that ensure that we can start the app, generate the schema, execute some queries and verify their responses.

smyrick:
@dkuc84 I'm fine with that but I don't think the sample app needs to be included in the code coverage report. So maybe we can instead change the codecov config

smyrick:
Closing in favor of https://github.com/ExpediaDotCom/graphql-kotlin/pull/297

PR #293, opened 2019-08-05
fix: allow modifying GraphQLInterfaceType in the willAddGraphQLTypeToSchema hook

dkuc84:
InterfaceBuilder attempts to create all the implementing `GraphQLObjectType`s after building an interface but before executing the `willAddGraphQLTypeToSchema` hook. This means that if we attempt to update the interface in that hook there could be some objects implementing that interface and referencing the old one. Since graphql-java schema objects don't implement hashcode we end up with type conflict of two interfaces being defined in the schema.

Fix is to use `GraphQLTypeReference` when referencing `GraphQLInterfaceType` from the `GraphQLObjectType` builder. Reference is just by name and they are resolved as the last step of building a schema. This allows us to modify the interface in the hook.

Resolves: https://github.com/ExpediaDotCom/graphql-kotlin/issues/281

### DIFF
--- a/graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt
+++ b/graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt
@@ -113,7 +113,8 @@ class SchemaGeneratorHooksTest {
                 hookCalled = true
                 return when {
                     generatedType is GraphQLObjectType && generatedType.name == "SomeData" -> GraphQLObjectType.newObject(generatedType).description("My custom description").build()
-                    generatedType is GraphQLInterfaceType -> GraphQLInterfaceType.newInterface(generatedType).description("My custom interface description").build()
+                    generatedType is GraphQLInterfaceType && generatedType.name == "RandomData" ->
+                        GraphQLInterfaceType.newInterface(generatedType).description("My custom interface description").build()
                     else -> generatedType
                 }
             }
@@ -129,6 +130,10 @@ class SchemaGeneratorHooksTest {
         val type = schema.getObjectType("SomeData")
         assertNotNull(type)
         assertEquals(expected = "My custom description", actual = type.description)
+
+        val interfaceType = schema.getType("RandomData") as? GraphQLInterfaceType
+        assertNotNull(interfaceType)
+        assertEquals(expected = "My custom interface description", actual = interfaceType.description)
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt
```diff
@@ -120,8 +98,10 @@ class SchemaGeneratorHooksTest {
             queries = listOf(TopLevelObject(TestQuery())),
             config = getTestSchemaConfigWithHooks(hooks)
         )
-        assertEquals(SomeData::class.createType(), hooks.lastSeenType)
-        assertEquals("SomeData!", hooks.lastSeenGeneratedType?.deepName)
+        assertTrue(hooks.seenTypes.contains(RandomData::class.createType()))
+        assertTrue(hooks.seenTypes.contains(SomeData::class.createType()))
+        // TODO bug: didGenerateGraphQLType hook is not applied on the object types build from the interface
```
gscheibel: TODO intended here?

graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt
```diff
@@ -132,7 +112,8 @@ class SchemaGeneratorHooksTest {
             override fun willAddGraphQLTypeToSchema(type: KType, generatedType: GraphQLType): GraphQLType {
                 hookCalled = true
                 return when {
-                    generatedType is GraphQLObjectType && generatedType.name == "SomeData" -> GraphQLObjectType.Builder(generatedType).description("My custom description").build()
+                    generatedType is GraphQLObjectType && generatedType.name == "SomeData" -> GraphQLObjectType.newObject(generatedType).description("My custom description").build()
+                    generatedType is GraphQLInterfaceType -> GraphQLInterfaceType.newInterface(generatedType).description("My custom interface description").build()
```
gscheibel: shouldn't  we be testing that `My custom interface description` is actually applied?

graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt
dkuc84: yes -> it is a bug in how we apply the hooks... :(

graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt
dkuc84: done

PR #297, opened 2019-08-12
chore: exclude the example app from code coverage

smyrick:
Instead of deleting the tests, we can just exclude this project from the reports

This is the previous PR https://github.com/ExpediaDotCom/graphql-kotlin/pull/290


smyrick:
🎉 `+26.12%`

Issue #294, opened 2019-08-05
Hooks are not applied on objects build from InterfaceBuilder and UnionBuilder


dkuc84:
**Library Version**
1.0.0-RC2

**Describe the bug**
`willAddGraphQLTypeToSchema ` and `didGenerateGraphQLType ` schema generator hooks are not applied on `GraphQLObjectType`s built from `InterfaceBuilder` and `UnionBuilder`. 

Hooks are only applied if GraphQL types are build using generic `TypeBuilder#graphQLTypeOf` method. Both interface and union builders attempt to build their implementing objects by invoking object builder directly which does not apply the hooks.

**To Reproduce**
See [SchemaGeneratorHooksTest.`calls hook after generating object type`](https://github.com/ExpediaDotCom/graphql-kotlin/blob/4c2652f46cd4813b217b185ff06dfe9b47497895/graphql-kotlin-schema-generator/src/test/kotlin/com/expedia/graphql/hooks/SchemaGeneratorHooksTest.kt#L87-L105)

**Expected behavior**
`willAddGraphQLTypeToSchema ` and `didGenerateGraphQLType ` hooks are consistently applied on `GraphQLObjectType`s generated from `InterfaceBuilder` and `UnionBuilder`.

PR #303, opened 2019-08-14
fix: *generateGraphQLType hooks are now called on interfaces

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/294

Interfaces now go through the internal TypeBuilder method 'graphQLTypeOf' instead of 'objectFromReflection'. This means we can make 'objectFromReflection' private. But the type may be (if not always) wrapped in a GraphQLNonNull which means we have to cast it back to the interface type before adding it to the ObjectBuilder.

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/SchemaGenerator.kt
@@ -18,7 +18,6 @@ import com.expedia.graphql.generator.types.SubscriptionBuilder
 import com.expedia.graphql.generator.types.UnionBuilder
 import graphql.schema.GraphQLCodeRegistry
 import graphql.schema.GraphQLDirective
-import graphql.schema.GraphQLInterfaceType
 import graphql.schema.GraphQLSchema
 import java.lang.reflect.Field
 import kotlin.reflect.KAnnotatedElement
@@ -76,8 +75,8 @@ open class SchemaGenerator(val config: SchemaGeneratorConfig) {
     open fun listType(type: KType, inputType: Boolean) =
         listTypeBuilder.listType(type, inputType)
 
-    open fun objectType(kClass: KClass<*>, interfaceType: GraphQLInterfaceType? = null) =
-        objectTypeBuilder.objectType(kClass, interfaceType)
+    open fun objectType(kClass: KClass<*>) =
+        objectTypeBuilder.objectType(kClass)
 
     open fun inputObjectType(kClass: KClass<*>) =
         inputObjectTypeBuilder.inputObjectType(kClass)
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/TypesCache.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/TypesCache.kt
@@ -39,8 +39,8 @@ internal class TypesCache(private val supportedPackages: List<String>) {
         val cacheKey = getCacheKeyString(key)
 
         if (cacheKey != null) {
-            typeUnderConstruction.remove(kGraphQLType.kClass)
-            return cache.put(cacheKey, kGraphQLType)
+            cache[cacheKey] = kGraphQLType
+            return kGraphQLType
         }
 
         return null
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/ObjectBuilder.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/ObjectBuilder.kt
@@ -31,7 +31,7 @@ internal class ObjectBuilder(generator: SchemaGenerator) : TypeBuilder(generator
             }
 
             kClass.getValidSuperclasses(config.hooks)
-                .map { graphQLTypeOf(type = it.createType(), inputType = false, annotatedAsID = false) }
+                .map { graphQLTypeOf(it.createType()) }
                 .forEach {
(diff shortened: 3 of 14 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/TypesCache.kt
```diff
@@ -35,12 +35,12 @@ internal class TypesCache(private val supportedPackages: List<String>) {
...
     }
 
-    fun put(key: TypesCacheKey, kGraphQLType: KGraphQLType): KGraphQLType? {
+    fun saveIntoCache(key: TypesCacheKey, kGraphQLType: KGraphQLType): KGraphQLType? {
         val cacheKey = getCacheKeyString(key)
 
         if (cacheKey != null) {
-            typeUnderConstruction.remove(kGraphQLType.kClass)
-            return cache.put(cacheKey, kGraphQLType)
+            cache[cacheKey] = kGraphQLType
```
smyrick: We don't need to remove from `typeUnderConstruction` since the only place we add is in the `buildIfNotUnderConstruction` method and we also remove it there.

graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/ObjectBuilder.kt
```diff
@@ -34,11 +35,10 @@ internal class ObjectBuilder(generator: SchemaGenerator) : TypeBuilder(generator
                 builder.withInterface(GraphQLTypeReference(interfaceType.name))
             } else {
                 kClass.getValidSuperclasses(config.hooks)
-                    .map { objectFromReflection(it.createType(), false) }
+                    .map { graphQLTypeOf(type = it.createType(), inputType = false, annotatedAsID = false) }
                     .forEach {
-                        when (it) {
-                            is GraphQLInterfaceType -> builder.withInterface(it)
```
smyrick: This was the bug. The type would sometimes come back as a GraphQLNonNull if we are now calling `objectFromReflection ` instead of `graphQLTypeOf`. That means this conditional would fail and it would never be added.

Now that we are calling `objectFromReflection`, all the hooks will be called as well

graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/TypesCache.kt
```diff
@@ -35,12 +35,12 @@ internal class TypesCache(private val supportedPackages: List<String>) {
         return null
     }
 
-    fun put(key: TypesCacheKey, kGraphQLType: KGraphQLType): KGraphQLType? {
+    fun saveIntoCache(key: TypesCacheKey, kGraphQLType: KGraphQLType): KGraphQLType? {
```
dkuc84: i kind of like the simple `put` better

graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/ObjectBuilder.kt
dkuc84: i dont think the comment is correct -> when building super classes it would build the interface correctly and add it to the type BUT since it was skipping the `graphQLTypeOf` and instead of doing the direct call to `objectFromReflection` it would also skip all the hooks that are defined in `graphQLTypeOf` method

smyrick:
@dkuc84 So I found the actual issue with this https://github.com/ExpediaDotCom/graphql-kotlin/issues/294

This code change doesn’t fix the test you commented in previously. If I simplify the unit test and have a query class that only returns field which the return type is the interface then the schema never directly contains all the implementations of the interface, they are all in the additional types. That means to fix the issue we just need to add the hooks where we call the additional types.

dkuc84:
@smyrick actual reason behind #294 are those lines -> https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/InterfaceBuilder.kt#L37 and https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/UnionBuilder.kt#L30

They are calling object builder directly instead of using the generic type builder method.

graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/SchemaGenerator.kt
```diff
@@ -76,8 +75,8 @@ open class SchemaGenerator(val config: SchemaGeneratorConfig) {
     open fun listType(type: KType, inputType: Boolean) =
         listTypeBuilder.listType(type, inputType)
 
-    open fun objectType(kClass: KClass<*>, interfaceType: GraphQLInterfaceType? = null) =
-        objectTypeBuilder.objectType(kClass, interfaceType)
+    open fun objectType(kClass: KClass<*>) =
```
smyrick: Since we no longer directly call the object type, we never call the method passing in an interface so we can clean this up

smyrick:
Ready for review

graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/types/ObjectBuilder.kt
smyrick: Updated with the new logic

graphql-kotlin-schema-generator/src/main/kotlin/com/expedia/graphql/generator/state/TypesCache.kt
smyrick: Done

PR #304, opened 2019-08-16
Changes to sdl field of _service object to make it compliant with the…

tobias-f:
… federation specification

- if the Query type is present, it should have the @extends directive, since it extends the query type
- the default schema definition should not be includeded in the sdl. This causes problems for example with the @apollo/gateway, because if the default schema definition is present, no queries are picked up from the federated schema.

### DIFF
--- a/graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederatedSchemaGeneratorHooks.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederatedSchemaGeneratorHooks.kt
@@ -59,9 +59,11 @@ open class FederatedSchemaGeneratorHooks(private val federatedTypeRegistry: Fede
             federatedQuery.field(entityField)
 
             // SDL returned by _service query should not contain directives or new scalars
-            val sdl = originalSchema.print()
+            val sdl = originalSchema
+                .print(includeDefaultSchemaDefinition = false)
                 .replace(directiveRegex, "")
                 .replace(scalarRegex, "")
+                .replace("type Query", "type Query @extends")
                 .trim()
             federatedCodeRegistry.dataFetcher(FieldCoordinates.coordinates(originalQuery.name, SERVICE_FIELD_DEFINITION.name), DataFetcher { _Service(sdl) })
             federatedCodeRegistry.dataFetcher(FieldCoordinates.coordinates(originalQuery.name, entityField.name), DataFetcher {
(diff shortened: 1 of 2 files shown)


graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederatedSchemaGeneratorHooks.kt
```diff
@@ -59,9 +59,11 @@ open class FederatedSchemaGeneratorHooks(private val federatedTypeRegistry: Fede
             federatedQuery.field(entityField)
 
             // SDL returned by _service query should not contain directives or new scalars
-            val sdl = originalSchema.print()
+            val sdl = originalSchema
+                .print(includeDefaultSchemaDefinition = false)
                 .replace(directiveRegex, "")
                 .replace(scalarRegex, "")
+                .replace("type Query", "type Query @extends")
```
dkuc84: I don't think this makes much sense, each service could define number of root queries that are not federated. Since this would apply to all federated schemas then there wouldn't be a base Query type anywhere, e.g.

service A - defines root query to resolve base type with @key 
service B - extends base type from service A

In the above situation both `Query` types would be marked as `extended` which is wrong

dkuc84:
> * if the Query type is present, it should have the @extends directive, since it extends the query type

if all `Query` types are extended then which service defines the base?

> * the default schema definition should not be included in the sdl. This causes problems for example with the @apollo/gateway, because if the default schema definition is present, no queries are picked up from the federated schema.

Sounds like a bug in `@apollo/gateway` - @pcarrier can you take a look?

dkuc84:
Documenting for completeness

>> * if the Query type is present, it should have the @extends directive, since it extends the query type
> if all `Query` types are extended then which service defines the base?

Apollo gateway defines the base `Query` type.  Each service that is part of the federated schema should be extending the `Query` type to signal that its queries are part of a larger federated schema.
 
> > * the default schema definition should not be included in the sdl. This causes problems for example with the @apollo/gateway, because if the default schema definition is present, no queries are picked up from the federated schema.
> 
> Sounds like a bug in `@apollo/gateway` - @pcarrier can you take a look?

This still does sound like a bug but indeed there is no need to include those defaults.

graphql-kotlin-federation/src/test/kotlin/com/expedia/graphql/federation/execution/ServiceQueryResolverTest.kt
```diff
@@ -28,7 +25,7 @@ type Book implements Product @extends @key(fields : "id") {
   weight: Float! @external
 }
 
-type Query {
+type Query @extends {
```
dkuc84: we should be excluding empty `Query` types as well

dkuc84: unfortunately `graphql-java` requires query object to be present

graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederatedSchemaGeneratorHooks.kt
dkuc84: in order to exclude empty query types, lets add this one as well

```
private val emptyQuery = "^type Query \\{$\\s+^\\}$\\s+".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))

// SDL returned by _service query should NOT contain
// - default schema definition
// - empty Query type
// - directives
// - new scalars
val sdl = originalSchema.print(includeDefaultSchemaDefinition = false)
    .replace(directiveRegex, "")
    .replace(scalarRegex, "")
    .replace(emptyQuery, "")
    .replace("type ${originalQuery.name}", "type ${originalQuery.name} @extends")
    .trim()
```

dkuc84:
Thanks for the fix!

PR #313, opened 2019-08-31
Fix wrapped federation

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/310

While I don't think it is possible to have an input type extended we should still be able to have nested types that are not part of the federated entities be generated properly. This can happen when there are lists of nullalbes or double wrapped elements

Screen Shot 2019-08-30 at 5 09 10 PM


dkuc84:
Instead of adding federation to the example app, lets keep it simple as is and create separate example for federation (https://github.com/ExpediaDotCom/graphql-kotlin/issues/300)

smyrick:
Let's merge this first then I can update this PR to resolve the bug fix alone: https://github.com/ExpediaDotCom/graphql-kotlin/pull/315

PR #315, opened 2019-09-03
Federation example

smyrick:
Add a federation example with two running spring boot apps and a Node.js app running `@apollo/gateway`

Here is a query sent to the gateways which resolves data directly from both apps, and also runs a query from `base-app` (`getWidgetById`) by then is extended with data from `extend-app`

Screen Shot 2019-09-04 at 12 08 00 PM


dkuc84:
@smyrick as per the federation docs workaround around `graphql-java` limitation is to use `@extends` directive instead ([link](https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/directives/ExtendsDirective.kt))

smyrick:
@dkuc84 Ah I missed that, ok that means `graphql-kotlin-federation` has a bug right now because the `type Query` does not have this directive on it

dkuc84:
@smyrick SDL used by federation is correct but yes there is a mismatch between generated `Query` type and the SDL returned by the `_service { sdl }` (https://github.com/ExpediaDotCom/graphql-kotlin/blob/master/graphql-kotlin-federation/src/main/kotlin/com/expedia/graphql/federation/FederatedSchemaGeneratorHooks.kt#L69).

dkuc84:
Maybe we should put all examples under single example folder?, e.g.

```
graphql-kotlin
- graphql-kotlin-federation
- graphql-kotlin-schema-generator
- examples
|- spring
|- federation
  |- base-app
  |- extend-app
  |- gateway 
```

thoughts?

smyrick:
Yea, I was thinking about that. It would make the top level easier to look at. I will update

smyrick:
@dkuc84 I would love to start using actions, but we need the org admins to opt-in to the feature.

PR #316, opened 2019-09-05
Fix nested federation

smyrick:
Fixes https://github.com/ExpediaDotCom/graphql-kotlin/issues/310

While I don't think it is possible to have an input type extended we should still be able to have nested types that are not part of the federated entities be generated properly. This can happen when there are lists of nullalbes or double wrapped elements

### DIFF
--- a/examples/spring/pom.xml
+++ b/examples/spring/pom.xml
@@ -8,7 +8,7 @@
         <relativePath>..</relativePath>
     </parent>
 
-    <artifactId>spring</artifactId>
+    <artifactId>spring-example</artifactId>
     <description>Example SpringBoot app using graphql-kotlin</description>
     <url>https://github.com/ExpediaDotCom/graphql-kotlin</url>
     <packaging>jar</packaging>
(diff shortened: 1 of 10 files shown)


examples/spring/pom.xml
```diff
@@ -8,7 +8,7 @@
         <relativePath>..</relativePath>
     </parent>
 
-    <artifactId>spring-example</artifactId>
+    <artifactId>spring</artifactId>
```
dkuc84: lets keep the previous name as it is more meaningful

smyrick: @dkuc84 Do you want to change the name of the folders all together? This is just the maven module name and i think it makes sense that it should match the directory name

If we want to change that I would do that in a separate PR

dkuc84: artifact ID often match the folder name but by no means it is a requirement. `spring-example` is much more meaningful artifact name than just the `spring`. Up to you whether you want to rename the folder to match the artifact ID but lets keep the module name as `spring-example`.

smyrick: I would prefer changing the name of the directory, I will do that as another PR

smyrick: I will undo my change here any if we need to make updates later we can

PR #329, opened 2019-09-07
Add new module graphql-kotlin-spring-server

dkuc84:
Create a new module `graphql-kotlin-spring-server` which is a Spring Boot auto-configuration library for creating GraphQL web application.


smyrick:
Nice! This is exactly what I was talking about this week. I think it would be good even just for us to move more of this basic spring code to open source since everyone will have to do this same setup

I propose we call it `graphql-kotlin-spring-server` though since Spring Boot is just a library on top of Spring

examples/federation/base-app/src/main/kotlin/com/expediagroup/graphql/sample/RoutesConfiguration.kt
```diff
@@ -16,69 +16,23 @@
...
-    @Bean
-    fun sdlRoute() = router {
-        GET("/sdl") {
-            ok().contentType(MediaType.TEXT_PLAIN).syncBody(schema.print())
-        }
-    }
-
     @Bean
-    fun graphQLToolRoutes() = router {
+    fun playgroundRoute() = coRouter {
```
smyrick: Do we want to include the playground in the library? Everyone is going to include graphiql or the playground if we don't so we could just add this in.

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/SchemaAutoConfiguration.kt
```diff
@@ -0,0 +1,75 @@
...
+@ConditionalOnProperty(value = ["graphql.federation.enabled"], havingValue = "false", matchIfMissing = true)
+@Configuration
+class SchemaAutoConfiguration {
+
+    private val logger = LoggerFactory.getLogger(SchemaAutoConfiguration::class.java)
+
+    @Bean
+    @ConditionalOnMissingBean
+    fun schemaConfig(config: GraphQLConfigurationProperties): SchemaGeneratorConfig = SchemaGeneratorConfig(
+            supportedPackages = config.packages
```
smyrick: How do we specify other config options? Do we override this bean or do we need to make the spring config more verbose

examples/federation/base-app/src/main/kotlin/com/expediagroup/graphql/sample/RoutesConfiguration.kt
dkuc84: while certainly useful, `playground` is optional and you don't need it to run a GraphQL server. I removed it from `spring-server` due to the above and also to simplify our logic in how to support custom `/graphql` (and `subscriptions` endpoint) as this value is hard coded in the HTML.

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/SchemaAutoConfiguration.kt
dkuc84: I though the simplest route would be to override the bean. All the auto configuration beans will back off if one provides their own.

smyrick:
Looks good to me, ready for review by others?

graphql-kotlin-federation/README.md
```diff
@@ -163,6 +163,6 @@ type Query {
 
 ## Documentation
 
-There are more examples and documentation in our [Wiki](https://github.com/ExpediaGroup/graphql-kotlin/wiki) or you can view the [javadocs](https://www.javadoc.io/doc/com.expedia/graphql-kotlin) for all published versions.
+There are more examples and documentation in our [Wiki](https://github.com/ExpediaGroup/graphql-kotlin/wiki) or you can view the [javadocs](https://www.javadoc.io/doc/com.expediagroup/graphql-kotlin) for all published versions.
```
smyrick: Can we update the javadocs link to `graphql-kotlin-federation`

smyrick:
@dkuc84 It looks like you need to pull from master and update `DataAndErrors.kt`

PR #335, opened 2019-09-09
Github action for PR check

smyrick:
Try a simple Github action for PR checks

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,20 @@
+name: Maven CI
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Build with Maven
+        run: mvn clean install --file pom.xml
```
mykevinjung: Do we need `install`?  For Jenkins jobs, we do not need to install the package into Jenkins local repo. `package` would be better?

smyrick: No, we don't, good catch

PR #337, opened 2019-09-10
Update .travis.yml

smyrick:
Remove simple install step and just use default travis commands

https://docs.travis-ci.com/user/languages/java/

Testing this out to figure out why JDK 8 does not include all modules...


dkuc84:
I think we could simply just try listing the active profiles, e.g. add `mvn help:active-profiles` to [update-version.sh](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/.travis/update-version.sh#L9) script. Default profiles should be automatically picked up unless some profile is explicitly specified.

PR #340, opened 2019-09-11
Mark junit as test scoped dependency

larslorenzen:
Hi, we just noticed that the junit dependency was not marked as test dependency.

What do you think?

best regards
Lars


dkuc84:
Great catch! Thanks!

Issue #317, opened 2019-09-05
Change @GraphQLID to only work on strings


smyrick:
**Library Version**
1.0.0-RC5

**Describe the issue**
Our current implementation allows for multiple types to be a `ID`:
* `String`, `Int`, `Long` or `UUID`

https://github.com/ExpediaDotCom/graphql-kotlin/wiki/Scalars#id

**Expected behavior**
According the to GraphQL spec the `ID` scalar should always serialize to a String

https://graphql.github.io/graphql-spec/June2018/#sec-ID

Do we need to change this behaviour?

dkuc84:
Our current behavior is based on `graphql-java` implementation which supports implicit conversions from `String`, `Int`, `Long`, `UUID` or `BigInteger` ([link](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/Scalars.java#L266)). As per the spec it will always serialize as a String.

If we update our logic to only support `String` then I believe it will make it much easier to resolve https://github.com/ExpediaDotCom/graphql-kotlin/issues/249 (otherwise deserialization part might be tricky).

smyrick:
@dkuc84 I resolved #249 here: https://github.com/ExpediaDotCom/graphql-kotlin/pull/319

So I think since `graphql-java` is doing the conversion of those types back to Strings we are still spec compliant, and it does make it easier to use types like `UUID`. 

I am going to close this then as discussion is not needed

dkuc84:
I'm actually leaning towards explicitly supporting only Strings - given an `ID` argument client has no idea whether it should be `Int`, `UUID` or whatever. If they specify data in wrong format (e.g. trying to pass `Int` when we expect `UUID`) it will cause runtime exceptions. It will be safer if we explicitly support only `String` and let the devs decide whether they need to validate it.

smyrick:
If we only had to support strings we could create a new simple class `ID` which can be used instead to be more explicit in the Kotlin code about the resulting schema which means we could remove this annotation

smyrick:
Another option is instead of having the annotation we can have a more explicit type that is a wrapper of a type parameter, and then do the wrapped type checks as we normally do.

```kotlin
data class ID<T>(val value: T)
```

This would be a minor change though and not sure if it's worth the effort to go through
```kotlin
class User(val id: ID<UUID>, val name: String)

// vs

class User(@GraphQLID val id: UUID, val name: String)
```

dkuc84:
[...]

PR #345, opened 2019-09-17
BREAKING CHANGE: Change GraphQLID to only use strings

smyrick:
### :pencil: Description
Due to deserialization issues that can happen with UUID, we are only going to support Strings as the explicit type for GraphQLID. You can still use UUID in your data fetchers but you will have to do the parsing yourself and handle the exceptions if it is not a valid UUID.

Example of client input

Screen Shot 2019-09-17 at 11 37 24 AM

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/317

### DIFF
--- a/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/SchemaGeneratorTest.kt
+++ b/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/SchemaGeneratorTest.kt
@@ -323,7 +323,7 @@ class SchemaGeneratorTest {
         val bobChildren = res?.get("children") as? List<Map<String, Any>>
         assertNotNull(bobChildren)
 
-        val firstChild = bobChildren.get(0)
+        val firstChild = bobChildren.first()
         assertEquals("Alice", firstChild["name"])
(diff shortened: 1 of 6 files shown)


graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/SchemaGeneratorTest.kt
```diff
@@ -507,16 +505,16 @@ class SchemaGeneratorTest {
     data class Person(val name: String, val children: List<Person>? = null)
 
     data class PlaceOfIds(
-        @GraphQLID val intId: Int,
-        @GraphQLID val longId: Long,
+        @GraphQLID val intId: String,
```
dkuc84: should we drop the other fields from test?

smyrick: Sure

smyrick:
https://github.com/ExpediaGroup/graphql-kotlin/issues/317#issuecomment-532352305

PR #351, opened 2019-09-18
Fix travis script to update versions

smyrick:
### :pencil: Description
Fix the script to update all pom files and check in all the files to git. Also update to the correct versions that was not done in the last build

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/pull/350


.travis/update-version.sh
```diff
@@ -13,7 +13,7 @@ update_version() {
     mvn --settings .travis/settings.xml org.codehaus.mojo:versions-maven-plugin:2.7:set -DnewVersion="${TRAVIS_TAG}-SNAPSHOT"
```
dariuszkuc: do we also need to specify the default build profile here?

smyrick: No that runs on all of them but I can just to be safe

PR #356, opened 2019-09-19
another attempt to fix update-version.sh script

dariuszkuc:
just add everything as it publishes to new branch that has to be still manually merged

### DIFF
--- a/.travis/update-version.sh
+++ b/.travis/update-version.sh
@@ -26,7 +26,7 @@ commit_files() {
     git checkout -b ${NEW_VERSION}
 
     # Stage the modified files
-    git add ./**/pom.xml
+    git add ./pom.xml ./*/pom.xml ./*/*/pom.xml
 
     # Print the current files we are going to commit
     git status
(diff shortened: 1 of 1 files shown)


.travis/update-version.sh
```diff
@@ -26,7 +26,7 @@ commit_files() {
     git checkout -b ${NEW_VERSION}
 
     # Stage the modified files
-    git add ./**/pom.xml
+    git add ./pom.xml ./*/pom.xml ./*/*/pom.xml
```
smyrick: We can just do a `git add -A`. We are only changing the pom files in this script and we still have to manually merge the PR

dariuszkuc: I thought that we wanted to explicitly limit it to just `pom.xml` files (I guess another safeguard?). If that is not the case then yeah lets just do `git add .` / `git add -A`

smyrick: Yea let's just add all since we have to merge a PR manually still

PR #368, opened 2019-09-22
fix non nullable nested object marked as nullable

### LINKED ISSUE #367: Non Nullable Nested Objects marked as Nullable
**Library Version**
1.0.0-RC9

**Describe the bug**
Non nullable nested objects are marked as null in some cases. If this object, then implements an interface that relies on a non nullable field, the following error will be thrown:

```
Caused by: graphql.schema.validation.InvalidSchemaException: invalid schema:
object type 'Widget' does not implement interface 'HoldsNestedWidget' because field 'nested' is defined as 'Widget' type and not as 'Widget!' type
```

**To Reproduce**
The following schema will cause the above error:

```kotlin
interface HoldsNestedWidget{
    fun nested(): Widget
}
class Widget(val id: Int): HoldsNestedWidget{
    override fun nested(): Widget = Widget(Random.nextInt())
}
@Component
class WidgetQuery: Query {
    fun nestedWidget(): HoldsNestedWidget = Widget(1)
}
```

If you remove the interface, you can see that the graphql property nested on Widget is nullable even though it is non nullable in Kotlin.

**Expected behavior**
No error thrown and nested field above is labelled as non null in schema.

de55:
### :pencil: Description
Fix non nullable nested object marked as nullable and add/update tests.

### :link: Related Issues
This resolves #367


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/TypeBuilder.kt
```diff
@@ -47,12 +47,12 @@ internal open class TypeBuilder constructor(protected val generator: SchemaGener
 
         // Do not call the hook on GraphQLTypeReference as we have not generated the type yet
         val unwrappedType = GraphQLTypeUtil.unwrapType(graphQLType).lastElement()
+        val typeWithNullability = graphQLType.wrapInNonNull(type)
```
smyrick: This would throw an exception if I had a type returned from the hooks that correctly had it wrapped in a `GraphQLNonNull` right? I think we can refactor the `wrapInNonNull` method to have the logic to check if it is already non-null. If it is just return it instead of throwing an exception

@dariuszkuc What do you think?

de55: Not sure what the original purpose of throwing that error was, but it would have thrown an error if it a GraphQLNonNull type wrapped anything besides a GraphQLTypeReference. I think the only ways for a GraphQLNonNull type to get into that method are to create a scalar type that is wrapped in GraphQLNonNull or implement onRewireGraphQLType to wrap a type in GraphQLNonNull. Idk if there is a valid use case for either one of those.

smyrick: I don't think there is. Let's just change the logic and remove the exception and all its test cases

de55: ok done.

Issue #370, opened 2019-09-23
Subscription support graphql-ws protocol


smyrick:
**Library Version**
1.0.0-RC9

**Describe the bug**
Subscriptions from the Apollo client do not work (also from GraphQLPlayground) since it uses a different protocol

https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md

**To Reproduce**
Send a request from the GraphQL playground in the `spring-example` module

**Expected behavior**
Subscriptions can work from Apollo Client sending the `graphql-ws` protocol

PR #371, opened 2019-09-23
Fix Apollo subscription message handling

smyrick:
### :pencil: Description
Support the messages that come from the Apollo client subscriptions.

https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md

We need to wrap the GraphQLRequest and GraphQLResponse in a special message before we send it over the web socket. That message structure is implemented in the handler for the `graphql-ws` protocol

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/370

### DIFF
--- a/examples/spring/src/main/resources/application.yml
+++ b/examples/spring/src/main/resources/application.yml
@@ -1,6 +1,4 @@
 graphql:
-  packages:
-    - "com.expediagroup.graphql.sample"
   subscriptions:
     # Send a ka message every 1000 ms (1 second)
     keepAliveInterval: 1000
--- a/examples/spring/src/main/resources/application.yml
+++ b/examples/spring/src/main/resources/application.yml
@@ -1,0 +1,3 @@
+graphql:
+  packages:
+    - "com.expediagroup.graphql.sample"
--- a/examples/spring/src/main/resources/application.yml
+++ b/examples/spring/src/main/resources/application.yml
@@ -1,3 +1,6 @@
 graphql:
   packages:
     - "com.expediagroup.graphql.sample"
+  subscriptions:
+    # Send a ka message every 1000 ms (1 second)
+    keepAliveInterval: 1000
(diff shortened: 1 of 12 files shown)


smyrick:
Two examples of sending the Keep alive message

https://github.com/leangen/graphql-spqr-spring-boot-starter/blob/0f5d66d478233d3d8401c1e5e4736ac3ed59d793/graphql-spqr-spring-boot-autoconfigure/src/main/java/io/leangen/graphql/spqr/spring/web/apollo/ApolloProtocolHandler.java#L51

and

https://github.com/graphql-java-kickstart/graphql-java-servlet/blob/373869b773bf280c1d745d8deb77665defd84929/src/main/java/graphql/servlet/core/internal/ApolloSubscriptionProtocolHandler.java#L74

examples/spring/src/main/resources/application.yml
```diff
@@ -0,0 +1,3 @@
+graphql:
+  packages:
+    - "com.expediagroup.graphql.sample"
```
dariuszkuc: this will be ignored because we override the config bean here -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/spring/src/main/kotlin/com/expediagroup/graphql/sample/Application.kt#L49

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/GraphQLConfigurationProperties.kt
```diff
@@ -45,4 +45,6 @@ class FederationConfigurationProperties {
 class SubscriptionConfigurationProperties {
     /** GraphQL subscriptions endpoint, defaults to 'subscriptions' */
     var endpoint: String = "subscriptions"
+    /** Keep the websocket alive and send a message to the client every interval in ms. Default to not sending messages */
+    var keepAliveInterval: Long? = null
```
dariuszkuc: unsure if it is needed but can you also add it in to https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-spring-server/src/main/resources/META-INF/spring-configuration-metadata.json

smyrick:
Here is an example of the subscriptions working on the playground

Screen Shot 2019-09-24 at 2 25 39 PM

smyrick:
And here is a `keep alive` example where the client stops the connection

Screen Shot 2019-09-24 at 2 36 26 PM

PR #376, opened 2019-09-25
Increase code coverage

smyrick:
### :pencil: Description
Update the code coverage levels with more unit tests and update the maven settings to exclude the examples in the Jacoco reports

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,6 +200,11 @@
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>${jacoco-maven-plugin.version}</version>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -200,6 +200,11 @@
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>${jacoco-maven-plugin.version}</version>
+                    <configuration>
+                        <excludes>
+                            <exclude>**/*com/expediagroup/graphql/sample/**</exclude>
```
dariuszkuc: outside of this PR but maybe we should rename `sample` package to `example`?

PR #386, opened 2019-09-27
spring-server: context web filter should only apply on graphql routes

dariuszkuc:
### :pencil: Description

`ContextWebFilter` should only be applicable on `/graphql` and `/subscription` routes.


graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ContextWebFilter.kt
```diff
@@ -32,14 +33,23 @@ const val GRAPHQL_CONTEXT_FILTER_ODER = 0
...
+                chain.filter(exchange).subscriberContext { it.put(GRAPHQL_CONTEXT_KEY, graphQLContext) }
+            }
+        } else {
+            chain.filter(exchange)
+        }
 
     override fun getOrder(): Int = GRAPHQL_CONTEXT_FILTER_ODER
+
+    internal fun isApplicable(exchange: ServerWebExchange): Boolean =
+        contextualPathsRegex.matches(exchange.request.uri.path)
```
tapaderster: what's the impact of this regex match on every request?

graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/spring/context/GraphQLContextFactoryIT.kt
```diff
@@ -0,0 +1,93 @@
...
+                val firstValue = coroutineContext[ReactorContext]?.context?.get<String>("firstFilterValue")
+                return CustomContext(
+                    first = firstValue,
+                    second = request.headers.getFirst("X-Second-Header") ?: "DEFAULT_SECOND"
+                )
+            }
+        }
+
+        @Bean
+        @Order(GRAPHQL_CONTEXT_FILTER_ODER - 10)
```
tapaderster: where does this `10` come from?

dariuszkuc: `10` doesn't matter, all what we care about is that final order value is < `GRAPHQL_CONTEXT_FILTER_ORDER`

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ContextWebFilter.kt
dariuszkuc: should be negligible. I could change it to string equality check unsure whether it will be any different.

tapaderster: think this should be fine.

graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/spring/context/GraphQLContextFactoryIT.kt
tapaderster: can we just do a `-1` then. Justifies the intent much clearly. `10` gives me an impression of a magic number if I am looking at the code out of the blue.

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ContextWebFilter.kt
dariuszkuc: updated to string comparison as it seemed to be slightly faster in local testing (100_000 executions took ~25ms)

Issue #381, opened 2019-09-26
DataLoaders - how to use?


neetkee:
Is it possible to use data loaders to perform batch queries? I can't find a way to use it with spring-server, I suppose we need to set dataLoaderRegistry somewhere in here?
https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/model/GraphQLRequest.kt#L28

neetkee:
I see, it's possible to extend SimpleQueryHandler, inject my own DataLoaderRegistry in it, and use it in DataFetcher. Is it right thing to do? 

And one more question, is there a reason why ExecutionResult.toGraphQLResponse() is internal here: [GraphQLResponse](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/model/GraphQLResponse.kt)

dariuszkuc:
Disclaimer: I haven't tried using data loader with this library

By default `graphql-kotlin` will invoke target underlying functions using [FunctionDataFetcher](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt). Within the target function itself you could easily add some data loading/batching logic.

You definitely can also override default query handler and try configuring your data loader registry there. If you get it working it would be great if you could contribute a PR to the example spring app.

----
re: `ExecutionResult.toGraphQLResponse() being internal

There wasn't a specific reason for it, it is just a convenience function so there wasn't a use case to make it pubic. Feel free to open up a PR to change it.

smyrick:
Example checked in here: https://github.com/ExpediaGroup/graphql-kotlin/pull/396

PR #396, opened 2019-09-28
Example of DataLoaders.

neetkee:
### :pencil: Description
I've asked about using dataLoaders in the related issue, so here is my example of its possible implementation, using an extension of a query handler. What do you think?

Note that I also made ExecutionResult.toGraphQLResponse() function public in this PR.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/381

### DIFF
--- a/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/OrganizationQuery.kt
+++ b/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/OrganizationQuery.kt
@@ -1,0 +1,54 @@
+package com.expediagroup.graphql.examples.query
+
+import com.expediagroup.graphql.annotations.GraphQLDescription
+import com.expediagroup.graphql.examples.model.Company
+import com.expediagroup.graphql.examples.model.Employee
+import com.expediagroup.graphql.spring.operations.Query
+import graphql.schema.DataFetcher
+import graphql.schema.DataFetchingEnvironment
+import org.springframework.beans.factory.BeanFactory
+import org.springframework.beans.factory.BeanFactoryAware
+import org.springframework.context.annotation.Scope
+import org.springframework.stereotype.Component
+import java.util.concurrent.CompletableFuture
+
+@Component
+class CompanyService {
+    private val companies = listOf(
+        Company(id = 1, name = "FirstCompany"),
+        Company(id = 2, name = "SecondCompany")
+    )
+
+    fun getCompanies(ids: List<Int>): List<Company> = companies
+}
+
+@Component
+class EmployeeQuery : Query {
+    private val employees = listOf(
+        Employee(name = "Mike", companyId = 1),
+        Employee(name = "John", companyId = 1),
+        Employee(name = "Steve", companyId = 2)
+    )
+
+    @GraphQLDescription("Get all employees")
+    fun employees(): List<Employee> {
+        return employees
+    }
+}
+
+@Component("CompanyDataFetcher")
+@Scope("prototype")
+class CompanyDataFetcher : DataFetcher<CompletableFuture<Company>>, BeanFactoryAware {
+    private lateinit var beanFactory: BeanFactory
+
+    override fun setBeanFactory(beanFactory: BeanFactory) {
+        this.beanFactory = beanFactory
+    }
+
+    override fun get(environment: DataFetchingEnvironment): CompletableFuture<Company> {
+        val companyId = environment.getSource<Employee>().companyId
+        return environment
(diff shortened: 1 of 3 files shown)


dariuszkuc:
Thanks for providing the example! I've added default support to `SimpleQueryHandler` in https://github.com/ExpediaGroup/graphql-kotlin/pull/400.

Would you mind updating the example to use latest code?

neetkee:
I guess we need to create DataLoaderRegistry per request, to avoid passing cached values between requests? I'm not sure how to do it with https://github.com/ExpediaGroup/graphql-kotlin/pull/400. Maybe we need to autowire some factory, that creates dataLoaderRegistry, I saw a similar way in SPQR lib: [DataLoaderRegistryFactory](https://github.com/leangen/graphql-spqr-spring-boot-starter/blob/0c2c956a0a9aea15345b210d02353c944126ae6e/graphql-spqr-spring-boot-autoconfigure/src/main/java/io/leangen/graphql/spqr/spring/autoconfigure/DataLoaderRegistryFactory.java)

dariuszkuc:
Interesting! So data loader registry should be created per request? I've added support for this in [#410](https://github.com/ExpediaGroup/graphql-kotlin/pull/410) by creating factory that allows developers to create new data loader registry instance per query execution. Unsure how useful it is as within `graphql-kotlin` we generate schemas directly from the code based on the underlying functions and classes (where devs can implement whatever) but I think it is good to have an option to use DataLoader.

smyrick:
👍 Plus one to this:
> Unsure how useful it is as within graphql-kotlin we generate schemas directly from the code based on the underlying functions and classes 

In graphql-java land, I can see how this is a helpful caching mechanism since you don't have a class that is run when you execute a resolver/data fetcher. But with our library you will be executing a function that can be in any scope and fetching data from a downstream service can be cached in any way you want. This feature will be something we can support through the factory but we will defer handling of everything else to the server devs.

neetkee:
@smyrick 
> But with our library you will be executing a function that can be in any scope and fetching data from a downstream service can be cached in any way you want.

Could you give me some direction on how it can be implemented?

smyrick:
@neetkee 


If you have this class and function as your schema object

```kotlin
class Query {
  fun getUserById(id: Int): User? {
    val result = getUserFromSource(id)
    return result
  }
}
```

the `getUserFromSource` could be any function that loads data from an API, database, or a cache system. This means that you could handle the caching layer yourself based on more inputs or configs and not solely based on the `DataLoader` implementation.

neetkee:
If I understand correctly, we can use it to load from cache/service/etc by id, one by one, but if we want to batch our request, we need to gather all needed ids, and then execute our query.

dariuszkuc:
> If I understand correctly, we can use it to load from cache/service/etc by id, one by one, but if we want to batch our request, we need to gather all needed ids, and then execute our query.

Yep. It all drills down to how you gather those IDs -> you need to know all of them to batch regardless whether you use data loader or not.

examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/OrganizationQuery.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @GraphQLDescription("Get all employees")
+    fun employees(): List<Employee> {
+        return employees
+    }
+}
+
+@Component("CompanyDataFetcher")
+@Scope("prototype")
+class CompanyDataFetcher : DataFetcher<CompletableFuture<Company>>, BeanFactoryAware {
```
RIP21, line 41: @smyrick hey. Am I understand that correctly that this thing will be called anytime somebody queries the` company` field `company`
`lateinit var Company` on Employee and ONLY the Employee, and not other classes?
Or I have to branch here or whatever?

Maybe some drawing or ASCII or draw.io graph will help me understand the connection between beans and the whole resolve cycle as I'm unsure :)

smyrick: @RIP21 The DataFetcher is the name of the "resolver" for graphql-java. You can see the docs here: https://www.graphql-java.com/documentation/v16/data-fetching/

This is a data fetcher for the entire `Company` class. GraphQL Kotlin normally would handle creating the different data fetchers for you, but in this case the user wanted have a `DataLoader` which is a separate GraphQL specific caching mechanism. The `DataFetcher` is invoked when ever there is a field that returns a `Company` and, using `DataLoader`s, they are cache the company info for each employee so they don't have to make multiple calls to get the the duplicate company data for all the employees

PR #404, opened 2019-09-30
Create publish docs action

smyrick:
### :pencil: Description
Publish the docusaurus build to the `gh-pages` branch on checkin


dariuszkuc:
Do we want to restrict it to only run on `docs` update?

PR #415, opened 2019-10-02
switch state from internal to public

de55:
### :pencil: Description
Switch state in `SchemaGenerator` (along with corresponding classes) from internal to public. I was running into the following error while trying to override the `objectType` function because an object was being used while being built:

Caused by: java.lang.ClassCastException: class graphql.schema.GraphQLTypeReference cannot be cast to class graphql.schema.GraphQLUnmodifiedType (graphql.schema.GraphQLTypeReference and graphql.schema.GraphQLUnmodifiedType are in unnamed module of loader 'app')

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -49,7 +49,7 @@ import kotlin.reflect.KType
 @Suppress("LeakingThis")
 open class SchemaGenerator(val config: SchemaGeneratorConfig) {
 
-    internal val state = SchemaGeneratorState(config.supportedPackages)
+    val state = SchemaGeneratorState(config.supportedPackages)
     internal val subTypeMapper = SubTypeMapper(config.supportedPackages)
     internal val codeRegistry = GraphQLCodeRegistry.newCodeRegistry()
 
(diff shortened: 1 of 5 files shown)


smyrick:
@de55 Can you provide the branch of the code that is throwing this error? This looks like a schema generation issue, that we can fix without exposing internal implementations.

If there is also something you need to override we should do that with hooks instead of the extending the class

smyrick:
Marking DNM until we can confirm this is needed

de55:
This is related to  #384 and #385.

Specifically this comment:

> It turns out I was not able to implement it through the hooks because I needed access to the generator, which creates a circular reference. I instead implemented it by inheriting from `SchemaGenerator` and overriding the `objectType` and `interfaceType` functions.
> 
> The only issue with this approach is that `FederatedSchemaGenerator` is final, so I had to copy/paste the code from that to my custom generator. Would you mind if that is made an open class?

After using the extension function in a different package, I ran into the issue.  However, I believe anyone that extends the schema generator (which has already been made open) would run into similar issues. Here is the override function:

```
    override fun objectType(kClass: KClass<*>): GraphQLType {
        val type = super.objectType(kClass)
        return state.cache.buildIfNotUnderConstruction(kClass) {
            val unwrappedType = GraphQLTypeUtil.unwrapNonNull(type)
            if (unwrappedType is GraphQLObjectType) addExtensionsToObject(unwrappedType, kClass, type) else type
        }
    }

    private fun addExtensionsToObject(unwrappedType: GraphQLObjectType, kClass: KClass<*>, origType: GraphQLType): GraphQLType {
        val builder = GraphQLObjectType.newObject(unwrappedType)
        val name = kClass.getSimpleName()
       //this method just gets the extension function for this kClass
        extensionFunctionMapper.getValidExtensionFunctions(kClass)
            .forEach { builder.field(super.function(it, name, null, false)) }
        val type = builder.build()
        return if(origType is GraphQLNonNull) GraphQLNonNull.nonNull(type) else type
    }
```

de55:
@smyrick any update on this? Do you think there’s a different way I should implement?

de55:
I was able to figure it out without exposing internals.

Issue #419, opened 2019-10-02
Interfaces of Interfaces is invalid GraphQL


smyrick:
**Library Version**
1.0.0

**Describe the issue**
According to the spec, an interface can not implement another interface. This is not caught by our schema generator but instead just returns the first level interface as the type in the schema.

However there is an outstand RFC into the GraphQL spec to allow this: https://github.com/graphql/graphql-spec/pull/373

Until this is approved and updated we should fail with an exception to better inform schema developers of the limitations when an interface has a supertype that is not `@GraphQLIgnored`

**To Reproduce**

This test passes, do we want to make it not pass?

```kotlin

import com.expediagroup.graphql.TopLevelObject
import com.expediagroup.graphql.testSchemaConfig
import com.expediagroup.graphql.toSchema
import org.junit.jupiter.api.Test
import kotlin.test.assertEquals
import kotlin.test.assertNotNull
import kotlin.test.assertNull

class InterfaceOfInterfaceTest {

    @Test
    fun `interface of interface`() {
        val queries = listOf(TopLevelObject(InterfaceOfInterfaceQuery()))

        val schema = toSchema(queries = queries, config = testSchemaConfig)

        assertEquals(expected = 1, actual = schema.queryType.fieldDefinitions.size)
        assertEquals(expected = "getClass", actual = schema.queryType.fieldDefinitions.first().name)

        assertNotNull(schema.typeMap["MyClass"])
        assertNotNull(schema.typeMap["SecondLevel"])
        assertNull(schema.typeMap["FirstLevel"])
    }

    interface FirstLevel {
        val id: String
    }

    interface SecondLevel : FirstLevel {
        val name: String
    }

    class MyClass(override val id: String, override val name: String) : SecondLevel

    class InterfaceOfInterfaceQuery {
        fun getClass() = MyClass(id = "1", name = "fooBar")
    }
}
```

produces

```graphql
schema {
  query: Query
}

interface SecondLevel {
  id: String!
  name: String!
}

type MyClass implements SecondLevel {
  id: String!
  name: String!
}

type Query {
  getClass: MyClass!
}
```

smyrick:
Changing this to an enhancement since we still technically generate a valid schema, just in the Kotlin world we are losing some information

PR #420, opened 2019-10-02
Throw exception on multi-level interface

smyrick:
### :pencil: Description
According to the spec, an interface can not implement another interface. This is not caught by our schema generator but instead just returns the first level interface as the type in the schema. We are now throwing an exception to help translate this understanding of GraphQL through the Kotlin code

Is this a breaking change? Someone may be using 1.0.0 today and just by updating their version they will have to fix something.

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/419

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/exceptions/InvalidInterfaceException.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/exceptions/InvalidInterfaceException.kt
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2019 Expedia, Inc
+ *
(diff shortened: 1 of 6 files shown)


smyrick:
I would also be fine not merging this and just keeping the implementation today that only includes the first level

smyrick:
Question for the team? Do we need to look multiple levels deep when resolving interfaces. Currently we only look one level from the class

dariuszkuc:
> Question for the team? Do we need to look multiple levels deep when resolving interfaces. Currently we only look one level from the class

I think supporting generation of the schema based on `class -> common abstract (ignored) -> interface` pattern might be useful. Currently we don't support that as we only look at direct ancestors.

smyrick:
@dariuszkuc Updated to return multiple levels deep now

PR #421, opened 2019-10-03
[spring-server] update to use Spring Boot RC1

dariuszkuc:


dariuszkuc:
Changing to minor version change as there was a breaking change in router functions between SpringBoot M6 and RC1 (`bodyAndAwait` -> `bodyValueAndAwait`)

PR #428, opened 2019-10-15
Fixed link to Gateway Readme

sundeep-expedia:
### :pencil: Description
The current link was causing a 404 as it was using an absolute path

### :link: Related Issues
None


smyrick:
Thank you for the contribution!

Issue #429, opened 2019-10-16
Spring Sample: Could not find artifact com.expediagroup:graphql-kotlin-spring-server:jar:1.1.1-SNAPSHOT


martinbonnin:
From current master (d6e5718dfeade09c46a18468cf7a4e97ec1dbbbb), it is not possible to build the spring sample:

```bash
~/git/graphql-kotlin/examples/spring$ mvn clean install
[INFO] Scanning for projects...
[INFO] Inspecting build with total of 1 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 1 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] 
[INFO] ------------------< com.expediagroup:spring-example >-------------------
[INFO] Building spring-example 1.1.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for com.expediagroup:graphql-kotlin-spring-server:jar:1.1.1-SNAPSHOT is missing, no dependency information available
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.614 s
[INFO] Finished at: 2019-10-16T10:47:21+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project spring-example: Could not resolve dependencies for project com.expediagroup:spring-example:jar:1.1.1-SNAPSHOT: Could not find artifact com.expediagroup:graphql-kotlin-spring-server:jar:1.1.1-SNAPSHOT -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
```

gscheibel:
Hello @martinbonnin,

We do not publish the snapshot version `1.1.1-SNAPSHOT` and but the `1.1.1` is available in Maven Central: https://search.maven.org/artifact/com.expediagroup/graphql-kotlin-spring-server/1.1.0/jar

martinbonnin:
[...]

PR #430, opened 2019-10-16
build dependencies before building the app

martinbonnin:
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/429


smyrick:
I updated the docs in this PR with a little more detail: https://github.com/ExpediaGroup/graphql-kotlin/pull/431

PR #441, opened 2019-10-17
[chore] force Github CI Action to use default maven build profile

dariuszkuc:


dariuszkuc:
Unsure why Github Actions ignore the `activeByDefault` setting.

PR #445, opened 2019-10-18
[spring-server] use data classes for configuration properties

dariuszkuc:
### :pencil: Description

SpringBoot 2.2 supports using immutable data classes for `ConfigurationProperties` by specifying `@ConstructorBinding`.


dariuszkuc:
Also cleaned up example app configs due to https://github.com/ExpediaGroup/graphql-kotlin/pull/433

PR #451, opened 2019-10-22
Fix GH actions PR check

smyrick:
### :pencil: Description
Migrate the actions Maven check to java 11

Also per this issue, we should be running ScanResult::close if we no longer need to use the reflection results: classgraph/classgraph#331


smyrick:
Per this issue, we should be running `ScanResult::close` if we no longer need to use the reflection results: https://github.com/classgraph/classgraph/issues/331

PR #456, opened 2019-10-25
[docs] spring-server documentation updates

dariuszkuc:


docs/server/spring-properties.md
```diff
@@ -5,15 +5,17 @@ title: Configuration Properties
 
 `graphql-kotlin-spring-server` relies on [GraphQLConfigurationProperties](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/GraphQLConfigurationProperties.kt) 
 to provide various customizations of the auto-configuration library. All applicable configuration properties expose [configuration
-metadata](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html) that provides
+metadata](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html) that provide
```
EXPErdelfavero: "metadata provides" is idiomatic usage, if not technically correct. I would only change it here if you're going to change similar phrasing everywhere.

smyrick: I would be fine on either here. I am not the professional writer though

smyrick:
If fixed the GH actions build here: https://github.com/ExpediaGroup/graphql-kotlin/pull/451

docs/server/spring-properties.md
EXPErdelfavero: Point taken, Shane. I've been avoiding doing copy edits on these docs because your team is actively maintaining them, using their authentic style, and they're clear enough. I don't want to be That Guy.  However Derek wants to write this is fine, because he's actually doing the work. I probably shouldn't have said anything.

smyrick: @EXPErdelfavero No, please continue providing feedback if you have time! We encourage all contributions to make the documentation and library easy to use and understand

EXPErdelfavero: Don't worry, I'll still be here. I'll just be thinking harder about commenting when there's something I think is important. This was a minor nit, not the sort of thing I should be taking up my time or yours with.

Issue #472, opened 2019-11-14
Pass context to FederatedTypeResolver


hrkfdn:
It seems that it is currently not possible to access the context in `FederatedTypeResolver.resolve()`. For queries, the `@GraphQLContext` annotation can be added to access the context, but federated type resolvers need to follow the interface specifications.

As the context can contain headers that are relevant for authentication, it would be very useful to have in the federated type resolvers. Alternatively, the `DataFetchingEnvironment` could be passed.

PR #473, opened 2019-11-14
[federation] federated type resolvers should have access to environment

dariuszkuc:
### :pencil: Description

When resolving the federated types we currently don't have access to GraphQLContext. GraphQLContext can contain some important information (e.g. related to auth) that might be needed for type resolution so the resolvers should be able to access it. This PR updates the `FederatedTypeResolver` interface to also accept `DataFetchingEnvironment` when batch resolving the types.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/472


smyrick:
Overall the changes look good to me. Do we want to wait until we have the other feature request that is a major change ready to go before we merge so we can release 2.0.0 fairly quickly?

dariuszkuc:
Yeah, maybe we should merge it into some 2.0.0-SNAPSHOT branch or something like that.

PR #475, opened 2019-11-15
[schema-generator] support GraphQLInputNullable

huangciyin:
### :pencil: Description
support @GraphQLInputNullable which can override the default nullability from kotlin type.

example:
```
class SomeInput {
    @GraphQLInputNullable(true)
    val someNumber: Int = 0
}

class TestQuery1 {
    fun query(@GraphQLInputNullable(false) input: SomeInput? = null): Int=0
}
```
will generate schema:
```
schema:type Query {
  query(input: SomeInput!): Int!
}

input SomeInput {
  someNumber: Int
}
```
@GraphQLInputNullable support Argument and InputProperty.

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/annotations/GraphQLInputNullable.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/annotations/GraphQLInputNullable.kt
@@ -1,0 +1,6 @@
+package com.expediagroup.graphql.annotations
+
+/**
+ * Set the GraphQL Input nullability to be picked up by the schema generator.
+ */
+annotation class GraphQLInputNullable(val value: Boolean)
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/TypeBuilder.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/TypeBuilder.kt
@@ -22,7 +22,7 @@ import com.expediagroup.graphql.generator.extensions.isEnum
 import com.expediagroup.graphql.generator.extensions.isInterface
 import com.expediagroup.graphql.generator.extensions.isListType
 import com.expediagroup.graphql.generator.extensions.isUnion
-import com.expediagroup.graphql.generator.extensions.wrapInNonNull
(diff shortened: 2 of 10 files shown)


smyrick:
@huangciyin What is the intent behind providing two ways to make things nullable? If you change the schema with out changing the Kotlin code then you could have a NPE as you won't be required to handles nulls. That doesn't seem save to me. I would rather have the nulls explicitly in the type system

huangciyin:
sometimes we have default value for kotlin prop, so it won't be null.  
example:
```
class SomeInput {
    @GraphQLInputNullable(true)
    val someNumber: Int = 0
}
```
we define the prop `someNumber` as  `Int`, not `Int?`, because it has default vaule, and we do not want check null in kotlin logic code.  
But when we use `Int`, the schema generated from kotlin type can not be null, which is not we want.

smyrick:
But if the value can be null from the client you will have to check for null anyway.

Instead of having default values in the arguments what if have them in the code. Currently we don't have a good way of getting default values in reflection.

```kotlin
class SomeInput {  
    val someNumber: Int?
}

class MyQuery {

fun getValue(input: Some input): Int {
    return input.someNumber ?: 0
}

}
```

huangciyin:
@smyrick If the prop is a simple primitive , we can use get function to handle this.
But consider what if the prop is a embed class, and we need set sub prop of the class? 
And also when we define Input And Output in a single kotlin class, if Input and Output has differrent nullability requirement, we need some way to recognize it.

dariuszkuc:
Your proposed change break the schema null type safety guaranteed by Kotlin which as Shane mentioned causes more issues than it solves. 

I believe what you are looking for is actually the ability to define default values from code -> https://github.com/ExpediaGroup/graphql-kotlin/issues/53. Unfortunately we are bound by limitations of Java so default value information cannot be retrieved through a reflection.

> And also when we define Input And Output in a single kotlin class, if Input and Output has differrent nullability requirement, we need some way to recognize it.

This means that input and output should be defined as different Kotlin classes.

smyrick:
As mentioned this breaks one of the core concepts of this library and since we have the same opinion from multiple maintainers I am going to close this pull request.

@huangciyin If you wanted to add customer annotations and schema building behavior you can do that with our hooks system but again, we highly recommend NOT dynamically changing nullability

Issue #359, opened 2019-09-20
validate subscription function return type when generating schema


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
Per [graphql-java docs](https://www.graphql-java.com/documentation/v13/subscriptions/) subscription data fetchers should return some `Publisher` (e.g. Reactor `Flux`). Currently we don't perform any validations to verify that functions specified for subscriptions return valid types.

**Describe the solution you'd like**
Schema generator should fail generation process if subscription function returns unsupported type.

**Describe alternatives you've considered**
N/A

**Additional context**
Related issue -> https://github.com/ExpediaGroup/graphql-kotlin/issues/358

JithendraD:
@dariuszkuc Can I pick this issue to work for Hacktoberfest.

smyrick:
Have a branch I am working on

PR #476, opened 2019-11-15
Validate subscription return type

smyrick:
### :pencil: Description
Subscriptions must return a `org.reactivestreams.Publisher` for graphql-java to work properly. An exception will now be thrown if any return type of a top level subscription is not a subclass of Publisher

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/359

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/exceptions/InvalidSubscriptionTypeException.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/exceptions/InvalidSubscriptionTypeException.kt
@@ -16,7 +16,13 @@
 
 package com.expediagroup.graphql.exceptions
 
+import com.expediagroup.graphql.generator.extensions.getSimpleName
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
 
-class InvalidSubscriptionTypeException(kClass: KClass<*>) :
-    GraphQLKotlinException("Schema requires all subscriptions to be public, ${kClass.simpleName} has ${kClass.visibility} visibility modifier")
+class InvalidSubscriptionTypeException(kClass: KClass<*>, kFunction: KFunction<*>? = null) :
+    GraphQLKotlinException(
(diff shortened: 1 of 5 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/exceptions/InvalidSubscriptionTypeException.kt
```diff
@@ -16,7 +16,13 @@
 
 package com.expediagroup.graphql.exceptions
 
+import com.expediagroup.graphql.generator.extensions.getSimpleName
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
 
-class InvalidSubscriptionTypeException(kClass: KClass<*>) :
-    GraphQLKotlinException("Schema requires all subscriptions to be public, ${kClass.simpleName} has ${kClass.visibility} visibility modifier")
+class InvalidSubscriptionTypeException(kClass: KClass<*>, kFunction: KFunction<*>? = null) :
```
dariuszkuc: Is there any use case when we would throw this without having access to the `KFunction`? If not then lets make it not nullable.

smyrick: We do throw it before we check the response type when we only check the visibility so I thought the error message might make more sense if we didn't include the non-relevant info but technically we could log it for all errors

smyrick: Actually, no we don't have access to the KFunction in the first place we throw the exception when we can the entire class visibility so we do have to keep it nullable

https://github.com/ExpediaGroup/graphql-kotlin/blob/26a25015cdee58504205f49cf23840c270d156d7/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/SubscriptionBuilder.kt#L41-L43

Issue #418, opened 2019-10-02
[federation] federated traces support


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
`graphql-kotlin-federation` should provide support for capturing traces for federated requests.

**Describe the solution you'd like**
Instrumentation that can be included in GraphQL configuration to capture federated trace information.

**Describe alternatives you've considered**
N/A

**Additional context**
https://www.apollographql.com/docs/apollo-server/federation/metrics/

tk26:
I'd like to take this issue up :)

smyrick:
@tk26 I have assigned you. Feel free to start a discussion here on your ideas or start a draft PR to show some examples!

Goregius:
Is it possible to use Apollo's federation-jvm library to use with this? 

```
@Bean
fun federatedTracingInstrumentation(): Instrumentation = FederatedTracingInstrumentation()
```

This is giving me a `DOWNSTREAM_SERVICE_ERROR` only for the gateway though.

dariuszkuc:
> Is it possible to use Apollo's federation-jvm library to use with this?

Haven't tried it myself but technically it should be possible as `FederatedTracingInstrumentation` from `federation-jvm` is just another instrumentation. I believe thought it requires custom `GraphQLContext` to work.

Goregius:
For some reason, it only works if I set the JDK version to 1.8, I was using JDK 11 before and it gave a ConcurrentModificationException when graphql-java was trying to execute the query

smyrick:
Is this issue still valid? Including custom instrumentation through this package seems to have worked for others so I don't think it is required for the 2.0 release

https://github.com/apollographql/federation-jvm/blob/master/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java

However maybe we can include this library in our federation lib so others don't have to? That could be a minor change/new feature release

dariuszkuc:
@smyrick `federation-jvm` is another implementation of Apollo Federation model for `graphql-java`. It seems weird to suggest to include whole library (that basically does same thing as we do but following SDL first approach) to just get the tracing. 

If we are to support Apollo Federation we should make it a complete package including tracing support that nicely integrates with our `graphql-kotlin-spring-server`. This should be part of `2.0.0` release.

smyrick:
Closed in https://github.com/ExpediaGroup/graphql-kotlin/pull/960

PR #477, opened 2019-11-17
Federated tracing

tk26:
### :pencil: Description
This is to add Federated Tracing Instrumentation to `graphql-kotlin-federation`
Apollo's `federation-jvm` is taken as the reference (https://github.com/apollographql/federation-jvm/pull/25/)

### :link: Related Issues
Draft PR for https://github.com/ExpediaGroup/graphql-kotlin/issues/418

### DIFF
--- a/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/tracing/FederatedTracingInstrumentation.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/tracing/FederatedTracingInstrumentation.kt
@@ -16,16 +16,40 @@
 
 package com.expediagroup.graphql.federation.tracing
 
+import com.expediagroup.graphql.exceptions.GraphQLKotlinException
 import com.google.protobuf.Timestamp
+import graphql.ExecutionResult
+import graphql.ExecutionResultImpl
+import graphql.GraphQLError
+import graphql.GraphqlErrorBuilder
+import graphql.execution.DataFetcherResult
 import graphql.execution.ExecutionPath
+import graphql.execution.ExecutionStepInfo
+import graphql.execution.instrumentation.InstrumentationContext
 import graphql.execution.instrumentation.InstrumentationState
 import graphql.execution.instrumentation.SimpleInstrumentation
+import graphql.execution.instrumentation.SimpleInstrumentationContext.whenCompleted
 import graphql.execution.instrumentation.parameters.InstrumentationCreateStateParameters
+import graphql.execution.instrumentation.parameters.InstrumentationExecutionParameters
+import graphql.execution.instrumentation.parameters.InstrumentationFieldFetchParameters
+import graphql.language.Document
+import graphql.language.SourceLocation
+import graphql.parser.InvalidSyntaxException
+import graphql.schema.GraphQLTypeUtil.simplePrint
 
 import mdg.engine.proto.Reports
 import java.time.Instant
+import java.util.ArrayList
+import java.util.Base64
+import java.util.concurrent.CompletableFuture
+import kotlin.collections.LinkedHashMap
 
 class FederatedTracingInstrumentation : SimpleInstrumentation() {
+    companion object {
+        const val HEADER_NAME: String = "apollo-federation-include-trace"
+        const val HEADER_VALUE: String = "ftv1"
+        const val EXTENSION_KEY: String = "ftv1"
+    }
 
     override fun createState(parameters: InstrumentationCreateStateParameters): InstrumentationState? {
         val context = parameters.executionInput.context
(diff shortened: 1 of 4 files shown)


tk26:
@dariuszkuc @smyrick Would love your feedback on this draft PR. As mentioned in the Description, this change refers implementation from `https://github.com/apollographql/federation-jvm` heavily (might as well be called it's Kotlin conversion)

If things look good, here's my TODO -
1. Give necessary credits to `federation-jvm`
2. Add tests

graphql-kotlin-federation/src/main/protobuf/reports.proto
```diff
@@ -0,0 +1,493 @@
+// This file is copied over from the `apollo-engine-reporting` package, which, in turn, derives its information from
+// Apollo's cloud infrastructure. It is expected that as the protobuf definition changes within the underlying system
+// we are reporting to, these libraries will need to be updated in sync.
+// https://github.com/apollographql/apollo-server/blob/master/packages/apollo-engine-reporting-protobuf/src/reports.proto
```
smyrick: Is there someway we can pull this file from NPM so we don't have to manage it here? https://www.npmjs.com/package/apollo-engine-reporting-protobuf

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/tracing/FederatedTracingInstrumentation.kt
```diff
@@ -0,0 +1,251 @@
...
+class FederatedTracingInstrumentation : SimpleInstrumentation() {
+    companion object {
+        const val HEADER_NAME: String = "apollo-federation-include-trace"
+        const val HEADER_VALUE: String = "ftv1"
+        const val EXTENSION_KEY: String = "ftv1"
+    }
+
+    override fun createState(parameters: InstrumentationCreateStateParameters): InstrumentationState? {
+        val context = parameters.executionInput.context
+        if (context is HttpRequestHeaders) {
```
dariuszkuc: This puts a requirement on the context object types. If we are going to go with this route we should create something very generic that has access to request and response objects instead.

dariuszkuc:
Wondering whether we need all of that logic - haven't looked much into the actual code but technically once we create the underlying federated objects using federated type resolver then individual methods should be traceable by using default https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/tracing/TracingInstrumentation.java.

graphql-kotlin-federation/src/main/protobuf/reports.proto
tk26: Hmmm! Interesting. I can take a look, good thought 👍

tk26:
> Wondering whether we need all of that logic - haven't looked much into the actual code but technically once we create the underlying federated objects using federated type resolver then individual methods should be traceable by using default [graphql-java/graphql-java:src/main/java/graphql/execution/instrumentation/tracing/TracingInstrumentation.java@`master`](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/tracing/TracingInstrumentation.java) [](/graphql-java/graphql-java/blob/HEAD@{2019-11-18T04:20:28Z}/src/main/java/graphql/execution/instrumentation/tracing/TracingInstrumentation.java).

Great point 👍 `TracingInstrumentation` seems to be able to do most of the job. We can make `FederatedTracingInstrumentation` extend `TracingInstrumentation` with a couple more changes - 
1. override `createState` to check if `apollo-federation-include-trace` header has `ftv1` value (Apollo Gateway's enforcement- https://www.apollographql.com/docs/apollo-server/federation/metrics/#how-tracing-data-is-exposed-from-a-federated-service)
2. Put the errors in `ftv1` extension (https://www.apollographql.com/docs/apollo-server/federation/metrics/#how-errors-work - paragraph 3)

The above 2 changes are necessary to make it work with Apollo Gateway - the question I have here is should we rather rename `FederatedTracingInstrumentation` as `ApolloFederatedTracingInstrumentation` for clarity?

smyrick:
I think `FederatedTracingInstrumentation` since federation is created by apollo but has a spec that is not Apollo specific

tk26:
> I think `FederatedTracingInstrumentation` since federation is created by apollo but has a spec that is not Apollo specific

The Apollo docsite references -
https://www.apollographql.com/docs/apollo-server/federation/metrics/#how-tracing-data-is-exposed-from-a-federated-service and https://www.apollographql.com/docs/apollo-server/federation/metrics/#how-errors-work

^ these indicate its very apollo-centric 🤔

smyrick:
@tk26 I am going to close the draft until this is ready for review again. This is just to clean up our open PRs. Feel free to reopen when you need

Issue #432, opened 2019-10-16
[schema-generator] remove DataFetcherExecutionPredicate from SchemaGeneratorHooks 


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
`SchemaGeneratorHooks` are used to modify the behavior of schema generation. `DataFetcherExecutionPredicate` is used at runtime by `FunctionDataFetcher` to allow modification of parameters based on the predicates (e.g. apply JSR-303 bean validations, see [spring example](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/ValidatedQuery.kt)).  

**Describe the solution you'd like**
`SchemaGeneratorHooks` should not contain the predicate. Maybe it should be part of the `SchemaGeneratorConfig` instead?

**Describe alternatives you've considered**
Do we even need to support the predicate evaluation in data fetcher? Directives could be used instead.

smyrick:
I think we can just delete the predicate all together. If someone wants custom behaviours they can now provide their own `DataFetcherFactory`

PR #478, opened 2019-11-18
BREAKING CHANGE: Remove DataFetcherExecutionPredicate

smyrick:
### :pencil: Description
Remove the predicate and all usages to simplify the execution code. We already have a way of specifying a custom DataFetcherFactory which requires that developers still pass in a predicate when instead they can just override all the behaviour in the predicate in the custom DataFetcher.

This is a breaking change so it will require a release of 2.0.0. We may also want to verify our own internal usage of this change in Expedia first with a 2.0.0 RC release. While it should be fine it will be good to verify that a major change like this doesn't drastically impact any existing users

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/432


dariuszkuc:
LGTM. We might want to provide some common way of applying execution predicates but I'm fine dropping it altogether for now.

PR #480, opened 2019-11-19
update to next major release

dariuszkuc:
While we don't have timeline for releasing 2.0.0 this will allow us to start merging breaking API changes.


smyrick:
Let's wait to merge until we have the releases fixed

PR #481, opened 2019-11-19
[chore] move dokka configuration to release profile

dariuszkuc:
### :pencil: Description

In order to release to nexus we need to publish Javadocs. Currently dokka only works for JDK8 so we had it explicitly gated behind a profile that would only be activated when run using Java 8. Since we are explicitly gating Travis release job to run on Java 8 only so we don't need this separate profile.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -385,16 +385,7 @@
                             </execution>
                         </executions>
                     </plugin>
-                </plugins>
-            </build>
-        </profile>
-        <profile>
-            <id>jdk8</id>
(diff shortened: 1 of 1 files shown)


dariuszkuc:
Unsure what changed on Travis recently but it finally started to behave correctly - i.e. when maven profile is explicitly set (https://github.com/ExpediaGroup/graphql-kotlin/blob/master/.travis/deploy.sh#L26) no other profiles should be activated.

PR #487, opened 2019-11-21
Allow for subscription request on the same endpoint

### LINKED ISSUE #484: Allow graphql.subscriptions.endpoint to be equals to graphql.endpoint = graphql
I want my graphql subscriptions endpoint to be `graphql` and not the default value `subscriptions`.
When I set `graphql.subscriptions.endpoint=graphql` the application started without problem but it seems that the graphql endpoint stop working.
playground can't connect to the server and when I run a query I get the following response from server:

```
{
  "error": {
    "timestamp": "2019-11-19T19:52:42.530+0000",
    "path": "/graphql",
    "status": 405,
    "error": "Method Not Allowed",
    "message": "Request method 'POST' not supported",
    "requestId": "b32bd99c"
  }
}

```
It should be possible to handle the websocket and http endpoint with the same path `/graphql`. 

N/A

**Additional context**
On many frontend app using apollo client the usual way to setup the client is as following, we just replace the http(s) by ws(s) to have the websocket URI.

```
const httpUri = process.env.APP_SERVER_URL + '/graphql';
const wsUri = httpUri.replace(/^https?/, 'ws');

const httpLink = new HttpLink({
  uri: httpUri,
  credentials: 'include',
});

const wsLink = new WebSocketLink({
  uri: wsUri,
  options: {
    // Automatic reconnect in case of connection error
    reconnect: true,
  },
});
```

smyrick:
### :pencil: Description
Filter out request over http that contain websocket headers and lower the order for when the ws:// protocol is handled so you can have both `http://app/graphql` and `ws://app/graphql`

### :link: Related Issues
Fixes #484

### DIFF
--- a/examples/spring/src/main/resources/application.yml
+++ b/examples/spring/src/main/resources/application.yml
@@ -3,3 +3,4 @@ graphql:
   subscriptions:
     # Send a ka message every 1000 ms (1 second)
     keepAliveInterval: 1000
+    endpoint: graphql
--- a/examples/spring/src/main/resources/application.yml
+++ b/examples/spring/src/main/resources/application.yml
@@ -3,4 +3,3 @@ graphql:
   subscriptions:
     # Send a ka message every 1000 ms (1 second)
     keepAliveInterval: 1000
-    endpoint: graphql
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/RoutesConfiguration.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/RoutesConfiguration.kt
@@ -52,8 +52,10 @@ class RoutesConfiguration(
     fun graphQLRoutes() = coRouter {
         val isEndpointRequest = POST(config.endpoint) or GET(config.endpoint)
         val isNotWebsocketRequest = headers {
-            it.header("Connection").contains("Upgrade").not()
-        }
+            // These headers are defined in the HTTP Protocol upgrade mechanism
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
+            it.header("Connection").contains("Upgrade").and(it.header("Upgrade").contains("websocket"))
+        }.not()
 
         (isEndpointRequest and isNotWebsocketRequest).invoke { serverRequest ->
             val graphQLRequest = createGraphQLRequest(serverRequest)
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/RoutesConfiguration.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/RoutesConfiguration.kt
@@ -50,11 +50,10 @@ class RoutesConfiguration(
 
     @Bean
     fun graphQLRoutes() = coRouter {
-        val isNotWebsocketRequest = headers {
-            it.asHttpHeaders()["Upgrade"]?.contains("websocket")?.not() == true
-        }
-
         val isEndpointRequest = POST(config.endpoint) or GET(config.endpoint)
+        val isNotWebsocketRequest = headers {
+            it.header("Connection").contains("Upgrade").not()
+        }
 
         (isEndpointRequest and isNotWebsocketRequest).invoke { serverRequest ->
             val graphQLRequest = createGraphQLRequest(serverRequest)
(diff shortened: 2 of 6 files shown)


smyrick:
I need to add an integration test so let's wait to merge until then

examples/spring/src/main/resources/application.yml
```diff
@@ -3,3 +3,4 @@ graphql:
   subscriptions:
     # Send a ka message every 1000 ms (1 second)
     keepAliveInterval: 1000
+    endpoint: graphql
```
dariuszkuc: i assume this is just from testing -> lets revert back to default

smyrick: Done

smyrick:
Ready for review

EXPErdelfavero:
Any documentation changes to be made, to go with the added functionality?

smyrick:
@EXPErdelfavero Since this is a bug fix there are no new feature to document.

Our current documentation is correct, we just now handle the proper mapping for the `graphql.subscription.endpoint` property: https://expediagroup.github.io/graphql-kotlin/docs/server/spring-properties

If we didn't add this fix in, we should have updated our docs with the assumption that the subscriptions URL and standard URL can't be the same

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/SubscriptionAutoConfiguration.kt
```diff
@@ -57,5 +56,5 @@ class SubscriptionAutoConfiguration {
 
     @Bean
     fun subscriptionHandlerMapping(config: GraphQLConfigurationProperties, subscriptionWebSocketHandler: SubscriptionWebSocketHandler): HandlerMapping =
-        SimpleUrlHandlerMapping(mapOf(config.subscriptions.endpoint to subscriptionWebSocketHandler), Ordered.HIGHEST_PRECEDENCE)
+        SimpleUrlHandlerMapping(mapOf(config.subscriptions.endpoint to subscriptionWebSocketHandler), 1)
```
dariuszkuc: maybe extract 1 to a some constant that talks about ordering?

smyrick:
@dariuszkuc Updated with test

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/RoutesConfiguration.kt
```diff
@@ -50,7 +50,16 @@ class RoutesConfiguration(
...
     @Bean
     fun graphQLRoutes() = coRouter {
-        (POST(config.endpoint) or GET(config.endpoint)).invoke { serverRequest ->
+        val isEndpointRequest = POST(config.endpoint) or GET(config.endpoint)
+        val isNotWebsocketRequest = headers {
+            // These headers are defined in the HTTP Protocol upgrade mechanism
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
+            val isUpgrade = requestContainsHeader(it, "Connection", "Upgrade")
+            val isWebSocket = requestContainsHeader(it, "Upgrade", "websocket")
+            isUpgrade.and(isWebSocket)
```
dariuszkuc: can we use `infix` notation here?

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/SubscriptionAutoConfiguration.kt
```diff
@@ -39,6 +38,16 @@ import org.springframework.web.reactive.socket.server.support.WebSocketHandlerAd
...
 
+    /**
+     * This value is needed so that this url handler is run without a drastically different order
+     * to the graphql routes in [RoutesConfiguration]. If we use [org.springframework.core.Ordered] to set as extreme
+     * high or low, then the requests are not handled properly.
+     *
+     * Hopefully we can eventually move the url handler to the same router DSL.
+     * https://github.com/spring-projects/spring-framework/issues/19476
+     */
+    private val urlHandlerOrder = 0
```
dariuszkuc: should this be `const`?

smyrick: const can only be used in a companion object, do we want to move it to an object?

dariuszkuc: instead of companion object i'd move it to top level field

smyrick: Done

PR #494, opened 2019-11-30
WIP migrating to gradle

dariuszkuc:


dariuszkuc:
Some additional changes to consider:

* separate unit tests and integration tests -> currently test plugin executes both in the same task, this is different from Maven where it runs both surefire and failsafe plugins
* cleanup dependencies between the modules - Gradle supports both `api` (equivalent of Maven `compile`) and `implementation` (equivalent of Maven `runtime`). We should double check what should be exposed in transitive dependencies.
* dokka gradle plugin generates some extra empty classes that contain only internal functions - didn't drill down into why it behaves differently than maven plugin, currently neither maven nor gradle generate javadocs for annotation classes
* is there any value in publishing root `pom.xml` (i.e. `graphql-kotlin`)?

dariuszkuc:
NOTE: source code changes were from `ktlint` formatting fixes

dariuszkuc:
Closing it for now. We'll revisit later on once a need to migrate materialize.

PR #496, opened 2019-12-02
Failing test to validate GH Actions check

smyrick:
### :pencil: Description
Change a unit test to fail to validate the new github actions PR check

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/pull/495


smyrick:
Checks run now, all good. I made a wrong change to the test file which is why it didn't fail

PR #503, opened 2019-12-05
Update interface docs

smyrick:
### :pencil: Description
Update the interface docs on multiple level interfaces

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/419


docs/writing-schemas/interfaces.md
```diff
@@ -76,3 +76,6 @@ type TopLevelQuery {
 }
 
 ```
+
+### Known Issues
+> NOTE: Due to a feature added in 1.0.0, we no longer support multiple levels of interfaces in a schema because the GraphQL spec does not support this feature.  [See 419](https://github.com/ExpediaGroup/graphql-kotlin/issues/419). If you do have multiple interfaces you will have to either combine them into a single interface or ignore all the parent interfaces.
```
EXPErdelfavero: Is there a test or sample code we can point them to for an example of what to do if you have multiple interfaces?

Proofread of this text is clean.

smyrick: Added code samples

PR #505, opened 2019-12-05
Update documenting-fields.md

smyrick:
### :pencil: Description
Update docs to include mentionion of markdown syntax


smyrick:
Also using this to trigger a docs publish

PR #508, opened 2019-12-06
Add new request exception handler method

smyrick:
### :pencil: Description
Allow for overriding of just the exception handling in SimpleQueryHandler. This will allow us to keep the default behaviour of execution, and provide custom error handling


dariuszkuc:
`SimpleQueryHandler` was supposed to provide some basic implementation - if developers need some custom logic they can easily create their own implementation. Is this really useful?

smyrick:
This is going to help some teams internally at Expedia Group override their own custom handling logic so we don't have to add any specific features to the open source library.

@dariuszkuc I will send you more details directly. If you can think of a better way to handle this request I am open for ideas.

smyrick:
We are going to move this extension into our implementation internally at EG so we don't have to modify the library just for our needs.

Issue #499, opened 2019-12-04
Subscription stops working doing automated tests


gklijs:
**Library Version**
1.4.2

**Describe the bug**
When running a load test, I very early on get time-outs. I have a comparable implementation using Java, and it runs a whole lot longer. It's about 2 minutes vs. 30 minutes. I don't see any error in the logs. What will happen is that every second a couple of times a subscription will be started, a message will be received, through Kafka and Webflux, and then the subscription will be stopped again.
A preliminary comparison can be found [here](https://graphql.gklijs.tech/results/graphql-servers/99-latency).

**To Reproduce**
At [kafka-graphql-examples](https://github.com/openweb-nl/kafka-graphql-examples/tree/ge-kotlin/graphql-endpoint) the implementation can be found. It's possible on mac/linux to run the whole test using the scripts.
I know this is quite a lot, and the test is doing a lot more than needed. I might try to create a test in Kotlin, using the example project that produces a similar result.

**Expected behavior**
To be at least on par with the java version, or be faster because of the use of coroutines.

I saw there was already an open issue for changing the subscription implementation, https://github.com/ExpediaGroup/graphql-kotlin/issues/358. It might fix the problem.

smyrick:
@gklijs Thank you for running tests against our library! It is helpful to see performance metrics outside of @ExpediaGroup

Do you think you could run the same perf test against this application? https://github.com/graphql-java-kickstart/graphql-spring-boot/tree/master/example-graphql-subscription

This is the graphql-java subscriptions example also using spring boot. If this is showing similar numbers than this issue really should be opened on graphql-java instead as we are not implementing much logic on top of the subscription execution.

gklijs:
The Java test is already based on that, it's in [this branch] (https://github.com/openweb-nl/kafka-graphql-examples/tree/ge-java). It's performing much better than this one, so the problems seems not entirely with graphql-java.

smyrick:
Ah ok, I apologise. I did not see it was a separate branch. This definitely is worth further investigation then. I don't have time to investigate this week but this is definitely high priority. Of course any community members are open to help!

gklijs:
Tried some things, using Flow instead of WebFlux to pass as publisher.  But it keeps the same. Shat after the first generator is created it stopped.
[...]

PR #510, opened 2019-12-08
fix subscriptions by not removing the ak subscription when one of the…

gklijs:
### :pencil: Description
Prevents stopping ka when ending a subscription

### :link: Related Issues
#499 I hoped, but is not helped by it.

### DIFF
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
@@ -75,11 +75,11 @@ class ApolloSubscriptionProtocolHandler(
                 }
                 operationMessage.type == GQL_START.type -> startSubscription(operationMessage, session)
                 operationMessage.type == GQL_STOP.type -> {
-                    stopSubscription(operationMessage, session)
+                    stopSubscription(operationMessage, session, false)
                     Flux.empty()
                 }
                 operationMessage.type == GQL_CONNECTION_TERMINATE.type -> {
-                    stopSubscription(operationMessage, session)
+                    stopSubscription(operationMessage, session, true)
                     session.close()
                     Flux.empty()
                 }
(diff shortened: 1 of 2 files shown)


smyrick:
@gklijs I tried checking out your performance benchmarks but the step up was taking a little bit longer than expected since I have not used clojure before

Is it possible that you can update the `SubscriptionWebSocketHandler.handle` method to this.

```kotlin
override fun handle(session: WebSocketSession): Mono<Void> {
        val response = session.receive()
            .map { it.payloadAsText }
            .flatMap { subscriptionHandler.handle(it, session) }
            .map { objectMapper.writeValueAsString(it) }
            .map { session.textMessage(it) }
            .doFinally { session.close() }

        return session.send(response)
    }
```

While the handler should close the session when receiving a `GQL_CONNECTION_TERMINATE` message, I just want to make sure that the session is in fact closing and not eating up the memory.

smyrick:
On a side note, I tried using sending the data over as bytes instead of converting to a string to see if that improved performance, but I my quick change did not seem to work so I will need to look into that more

```kotlin
val response = session.receive()
            .map { it.payloadAsText }
            .flatMap { subscriptionHandler.handle(it, session) }
            .map { objectMapper.writeValueAsBytes(it) }
            .map { session.binaryMessage { dataBufferFactory -> dataBufferFactory.wrap(it) } }
            .doFinally { session.close() }
```

gklijs:
I'll give it a try tomorrow. You indeed need leiningen (build tool for Clojure) and sassc installed. Now that I better know the problem I might try create a test in the project itself trough.
Binary data is handled differently as string values in a websocket so that might be reason it fails. I don't that's a good idea because clients might break because of it.

gklijs:
The mentioned fix https://github.com/ExpediaGroup/graphql-kotlin/pull/510#issuecomment-563370716 didn't help. It's still stops after around 120 subscriptions being started and stopped.

smyrick:
@gklijs Can you try adding in these changes: https://github.com/ExpediaGroup/graphql-kotlin/pull/512

We don't need to save the session connections, because to cancel them later on we reference by session id, so we can just directly cancel the session on `GQL_CONNECTION_TERMINATE`. The open client streams can be cancelled if the client then sends `GQL_STOP`

gklijs:
Finally kind of done updating the results (not all variants ran 10 times, would improve accuracy a bit, but not much) https://graphql.gklijs.tech/results/graphql-servers/transactions-per-second

PR #514, opened 2019-12-13
Add perf tests to validate subscription performance

smyrick:
### :pencil: Description
Add a simple perf test runner, https://artillery.io/. This allows us to easily run the perf tests and debug the server in the same project

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/499


smyrick:
I ran the perf tests locally then manually triggered a single request and I can see that the cached values in `activeSessions` and `activeOperations` or correctly being cleaned up if the client is sending the proper messages. See the size is zero.

Screen Shot 2019-12-13 at 11 24 09 AM

smyrick:
However, in an unrelated note, GraphQL Playground does not terminate the session on stop. It only sends the `STOP` message which should stop sending data but keep the session and the flux active.

So this may be an issue. I will try to debug this but if @gklijs you want to use this perf test setup to verify other libraries that could help

smyrick:
This is ready to merge as is. If there is a bug fix I can make that a separate PR.

PR #515, opened 2019-12-13
Fix subscription caching logic

smyrick:
### :pencil: Description
Please merge this first: https://github.com/ExpediaGroup/graphql-kotlin/pull/514

Move the saving of subscriptions to a separate class so we can verify the logic with unit tests and simplify the`ApolloSubscriptionProtocolHandler`. This also exposed a bug that we were not saving the operation subscriptions to be stopped properly. This is now covered by the unit tests

### :link: Related Issues
* Builds on top of https://github.com/ExpediaGroup/graphql-kotlin/pull/514 for testing purposes
* Hopefully will help with subscription performance: #499


gklijs:
Just build the source branch, and then the test related to #499. Still the same result after about 120 subscriptions that have been started and stopped the websocket becomes unresponsive.

Issue #141, opened 2019-01-08
Issue #136 (graphql.AssertException: interfaceType can't be null) isn't fixed


Togrias:
Please revisit issue #136. Union types are fixed but interface types remain broken.

```
interface SomeUnion
class A: SomeUnion { fun getB(): B }
class B: SomeUnion { fun getA(): A }
```

The above builds properly.

```
interface SomeInterface {
  val id: String
}
class C: SomeInterface {
  fun getD(): D 
  override val id: String { ... }
}
class D: SomeInterface { 
  fun getC(): C 
    override val id: String { ... }
}
```

This doesn't.

PR #520, opened 2019-12-17
Prevent starting second subscription with the same id

gklijs:
#141 # :pencil: Description
Prevent starting second subscription with the same id, remove he complete message as it's not needed, and probably preventing flux from closing. Validated this will solve the problems, as it finally ran longer than 2 minutes.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/499

### DIFF
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
@@ -65,10 +65,7 @@ class ApolloSubscriptionProtocolHandler(
                     return ackowledgeMessageFlux.concatWith(keepAliveFlux)
                 }
                 GQL_START.type -> return startSubscription(operationMessage, session)
-                GQL_STOP.type -> {
-                    sessionState.stopOperation(session, operationMessage)
-                    return Flux.empty()
-                }
+                GQL_STOP.type -> return sessionState.stopOperation(session, operationMessage)
                 GQL_CONNECTION_TERMINATE.type -> {
                     sessionState.terminateSession(session)
                     return Flux.empty()
@@ -107,6 +104,11 @@ class ApolloSubscriptionProtocolHandler(
             return Flux.just(basicConnectionErrorMessage)
         }
 
+        if (sessionState.operationExists(session, operationMessage)) {
+            logger.info("Already subscribed to operation ${operationMessage.id} for session ${session.id}")
+            return Flux.empty()
+        }
+
         val payload = operationMessage.payload
 
         if (payload == null) {
@@ -125,7 +127,6 @@ class ApolloSubscriptionProtocolHandler(
                         SubscriptionOperationMessage(type = GQL_DATA.type, id = operationMessage.id, payload = it)
                     }
                 }
-                .concatWith(Flux.just(SubscriptionOperationMessage(type = SubscriptionOperationMessage.ServerMessages.GQL_COMPLETE.type, id = operationMessage.id)))
                 .doOnSubscribe { sessionState.saveOperation(session, operationMessage, it) }
         } catch (exception: Exception) {
             logger.error("Error running graphql subscription", exception)
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionSessionState.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionSessionState.kt
@@ -17,8 +17,10 @@
 package com.expediagroup.graphql.spring.execution
 
 import com.expediagroup.graphql.spring.model.SubscriptionOperationMessage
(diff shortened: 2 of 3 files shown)


graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/model/SubscriptionOperationMessage.kt
```diff
@@ -42,7 +42,6 @@ data class SubscriptionOperationMessage(
         GQL_CONNECTION_ERROR("connection_error"),
         GQL_DATA("data"),
         GQL_ERROR("error"),
-        GQL_COMPLETE("complete"),
```
smyrick: We still need to return this message to be compliant with the graphql-ws spec

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
```diff
@@ -107,6 +107,11 @@ class ApolloSubscriptionProtocolHandler(
             return Flux.just(basicConnectionErrorMessage)
         }
 
+        if (sessionState.operationExists(session, operationMessage)) {
```
smyrick: I was thinking about this earlier and I don't think we need to have this check. There is only every one session open per connection so we are not going to every have a client open a connection with duplicate session IDs and request ID.

If that some how did happen I imagine we would want to return the active subscription but I don't think we need to handle this

gklijs: We will overwrite the value in the map, so the privious one won't be closed, and you might get messages double.

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/model/SubscriptionOperationMessage.kt
gklijs: I wasn't aware of https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md I'll try to change so that on u subscribe it's still send.

gklijs: I return it now on a succesfull stop operation. Also added test for it.

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
gklijs: Another way to handle would be when we add the subscription, to first cancel any existing one.

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionSessionState.kt
```diff
@@ -53,16 +55,19 @@ internal class ApolloSubscriptionSessionState {
...
-
-            if (operationsForSession?.isEmpty() == true) {
-                activeOperations.remove(session.id)
+            operationsForSession?.get(operationMessage.id)?.let {
+                it.cancel()
+                operationsForSession.remove(operationMessage.id)
+                if (operationsForSession.isEmpty()) {
+                    activeOperations.remove(session.id)
+                }
+                return Flux.just(SubscriptionOperationMessage(type = GQL_COMPLETE.type, id = operationMessage.id))
```
smyrick: This behaviour is not correct. We need to send the complete event back to clients when the subscription is complete, not when the operation has an error or the client manually stops it

https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_complete

gklijs: For queries and mutation yes, for subscriptions this is different:
`For subscriptions:
App triggers PubSub's publication method, and the server publishes the event, passing it through the subscribe executor to create GraphQL execution result
Client receives GQL_DATA with the data, and handles it.
When client unsubscribe, the server triggers onOperationDone and sends GQL_COMPLETE message to the client.`
Is it the intent to also handle queries and mutation over websocket? Cause for graphql-java-kickstart this is not in scope.

smyrick: NVM from the spec as of December 2019

> Server sends GQL_COMPLETE if the operation is a query or mutation (for subscriptions, this sent when unsubscribing)

smyrick:
Thanks for your help on the subscriptions @gklijs. If anyone else has any comments or bugs on subscriptions please comment below. We are not currently using them at ExpediaGroup so we don't have the traffic to verify but we still want this library to support subscriptions 100%

PR #521, opened 2019-12-18
Allow to override FunctionDataFetcher objectmapper

max9599:
Currently it is not possible to provide my object mapper

(I use Vert.x and they provide core mapper instance, which I would like to use to handle  (de)serialization of dates/timestamps)


max9599:
open class could be overriden, my mistake

Issue #524, opened 2019-12-20
[spring-server] default data fetcher should use Jackson object mapper from spring context


dariuszkuc:
**Library Version**
2.0.0-SNAPSHOT

**Describe the bug**
Currently it is not possible to provide custom Jackson object mapper (e.g. one that includes java date time module) to be used by default function data fetcher (see code below). In order to use custom object mapper users have to create their own instance of data fetcher factory provider.

```kotlin
open class KotlinDataFetcherFactoryProvider {

    private val defaultObjectMapper = jacksonObjectMapper()

    /**
     * Retrieve instance of [DataFetcherFactory] that will be used to resolve target function.
     *
     * @param target target object that performs the data fetching or NULL if target object should be dynamically
     * retrieved during data fetcher execution from [graphql.schema.DataFetchingEnvironment]
     * @param kFunction Kotlin function being invoked
     */
    open fun functionDataFetcherFactory(target: Any?, kFunction: KFunction<*>): DataFetcherFactory<Any> =
            DataFetcherFactories.useDataFetcher(
                FunctionDataFetcher(
                        target = target,
                        fn = kFunction,
                        objectMapper = defaultObjectMapper)
            )
...
}
```

**To Reproduce**
Create custom `ObjectMapper` in spring context (or customize it by using `Jackson2ObjectMapperBuilderCustomizer`) and notice that it is ignored by `KotlinDataFetcherFactoryProvider` as it [creates its own instance anyway](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/KotlinDataFetcherFactoryProvider.kt#L33).

**Expected behavior**
`ObjectMapper` from spring application context should be used by default data fetcher.

Issue #518, opened 2019-12-17
Apollo Uploads don't work correctly


erikhofer:
**Library Version**
1.4.2

**Describe the bug**
Im using graphql-kotlin to build the schema for graphql-java (because I can't switch to WebFlux at the moment). I'd like to use Apollo uploads but they don't quite work. The schema is created correctly but the datafetcher throws at runtime.

<details>
  <summary>Stack trace</summary>

```
java.lang.IllegalArgumentException: No serializer found for class java.io.FileDescriptor and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: org.apache.catalina.core.ApplicationPart["inputStream"]->java.io.FileInputStream["fd"])
	at com.fasterxml.jackson.databind.ObjectMapper._convert(ObjectMapper.java:3922)
	at com.fasterxml.jackson.databind.ObjectMapper.convertValue(ObjectMapper.java:3853)
	at com.expediagroup.graphql.execution.FunctionDataFetcher.convertParameterValue(FunctionDataFetcher.kt:85)
	at com.expediagroup.graphql.execution.FunctionDataFetcher.mapParameterToValue(FunctionDataFetcher.kt:79)
	at com.expediagroup.graphql.execution.FunctionDataFetcher.get(FunctionDataFetcher.kt:58)
	at graphql.execution.instrumentation.dataloader.DataLoaderDispatcherInstrumentation.lambda$instrumentDataFetcher$0(DataLoaderDispatcherInstrumentation.java:86)
	at graphql.execution.ExecutionStrategy.fetchField(ExecutionStrategy.java:270)
	at graphql.execution.ExecutionStrategy.resolveFieldWithInfo(ExecutionStrategy.java:198)
	at graphql.execution.AsyncExecutionStrategy.execute(AsyncExecutionStrategy.java:74)
	at graphql.execution.Execution.executeOperation(Execution.java:161)
	at graphql.execution.Execution.execute(Execution.java:102)
	at graphql.GraphQL.execute(GraphQL.java:605)
	at graphql.GraphQL.parseValidateAndExecute(GraphQL.java:538)
	at graphql.GraphQL.executeAsync(GraphQL.java:502)
	at graphql.servlet.core.GraphQLQueryInvoker.query(GraphQLQueryInvoker.java:106)
[...]

PR #525, opened 2019-12-20
[spring-server] default data fetcher should use Jackson object mapper from spring context

dariuszkuc:
### :pencil: Description
Currently it is not possible to provide custom Jackson object mapper (e.g. one that includes java date time module) to be used by default function data fetcher. In order to use custom object mapper users have to create their own instance of data fetcher factory provider. This change updates the data fetcher factory provider to accept an instance of object mapper that will be used by the function data fetcher.

### :link: Related Issues
Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/524


smyrick:
LGTM

smyrick:
Is this ready for review? Was there some other feature you were looking to add?

dariuszkuc:
> Is this ready for review? Was there some other feature you were looking to add?

Looking into adding some unit test for this.

smyrick:
@dariuszkuc This look good to me. We have the single test. Is this good to merge?

dariuszkuc:
@smyrick this should be good to merge - was looking into whether we need to have that jackson object mapper in the first place and unfortunately it is required in order to process any complex input objects. 

On the side note - currently we do an extra conversion from primitive scalar value to the same scalar type (as noted [here](https://github.com/ExpediaGroup/graphql-kotlin/issues/518#issuecomment-566996194)). Unsure how big of a perf impact that is but we might want to optimize it as well.

smyrick:
Major version changes now that we changed from class to interface

PR #531, opened 2019-12-27
Fix recursive input schema

smyrick:
### :pencil: Description
If there was an input type used recursively or an output type that self-references on one of it's functions then the schema generator would fail. This PR fixes that by doing a couple things:

 * Refactor the schema generator to expose graphQLTypeOf and delete TypeBuilder.
 * Adds unit tests to verify all recursive inputs
 * Caches values and types under construction by considering if they are input types

There are a lot more changes with the schema generator refactor here that I will continue with in a second PR to clean up the code even more, but that will have no effect on features so I am making that a separate PR.

### :link: Related Issues
* Fixes #527
* Helps start resolving https://github.com/ExpediaGroup/graphql-kotlin/issues/84 (more to come on this)

Issue #84, opened 2018-12-04
Untangle cyclic dependency between SchemaGenerator and type builders


smyrick:
We have this due to the recursive nature of the code. But does anyone have a better approach? This is just to bring up the discussion. If we don't come to a good answer we can keep the code as is will do no harm

I prefer having the separate type builders and that requires they have at least one common function they can call back to

smyrick:
I don't see a good way to move away from this right now. If anyone wants to open a different PR feel free

PR #532, opened 2019-12-27
Refactor schema generator to remove cyclic dependencies

smyrick:
### :pencil: Description
To remove the cyclic structure of the TypeBuilder and SchemaGenerator, I moved all the individual type builder to be pure functions that accept the schema generator as an argument. This means we can remove the methods on the schema generator that allow us to call the different type builders

### :link: Related Issues
Please merge this first: https://github.com/ExpediaGroup/graphql-kotlin/pull/531
Related to old issue: https://github.com/ExpediaGroup/graphql-kotlin/issues/84

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/ScalarBuilder.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/ScalarBuilder.kt
@@ -30,11 +30,11 @@ import kotlin.reflect.full.isSubclassOf
 
 internal fun generateScalar(generator: SchemaGenerator, type: KType, annotatedAsID: Boolean): GraphQLScalarType? {
     val kClass = type.getKClass()
-
     val scalar = when {
         annotatedAsID -> getId(kClass)
-        else -> ScalarMap.defaultScalarsMap[kClass]
+        else -> defaultScalarsMap[kClass]
     }
+
     return scalar?.let {
         generator.config.hooks.onRewireGraphQLType(it).safeCast()
     }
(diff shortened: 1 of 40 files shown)


smyrick:
I can take this one step further by moving out `graphQLTypeOf` to its own internal function and call it `generateGraphQLType`, then I don't need to pass around the `SchemaGenerator` and instead can just pass the `SchemaGeneratorConfig`

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/ScalarBuilder.kt
```diff
@@ -28,41 +28,38 @@ import kotlin.reflect.KClass
...
+@Throws(InvalidIdTypeException::class)
+private fun getId(kClass: KClass<*>): GraphQLScalarType? {
+    return if (kClass.isSubclassOf(String::class)) {
+        Scalars.GraphQLID
+    } else {
+        throw InvalidIdTypeException(kClass)
     }
 }
+
+private object ScalarMap {
```
dariuszkuc: do we need the object wrapper in the first place? top level `private val defaultScalarsMap ` will be compiled to `private final static defaultScalarsMap `

dariuszkuc:
> I can take this one step further by moving out `graphQLTypeOf` to its own internal function and call it `generateGraphQLType`, then I don't need to pass around the `SchemaGenerator` and instead can just pass the `SchemaGeneratorConfig`

@smyrick we can do that in the subsequent PR if needed

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/ScalarBuilder.kt
smyrick: No we don't I can move this out

smyrick:
@dariuszkuc I think I will make a separate PR to move out `graphQLTypeOf` because it requires more work to fix the federation tests, so for now this is ready to review and merge.

PR #535, opened 2020-01-02
BREAKING CHANGE: Refactor graphQLTypeOf to top level function

smyrick:
### :pencil: Description
Move the function to generate a graphql type to a top level file function. This makes the function pure since it accepts the state of the SchemaGenerator as an argument and cleans up the methods that we need to expose in the class. It does however still make it public in graphql-kotlin-schema-generator as we still need to be able to it them from graphql-kotlin-federation

#### Other Changes
* Simplify the SchemaGenerator by removing `SchemaGeneratorState` and making all the properties internal, and adding a helper method `addAdditionalTypesWithAnnotation` to be used from federation only
* Rename internal class `SubTypeMapper` to `ClassScanner`
* Move `TopLevelNames` to its own file
* Refactor `FederatedSchemaGenerator` to extend with just one more internal method `addExtendedTypes` so that we can keep the `ClassScanner` internal to `graphql-kotlin-schema-generator`

### :link: Related Issues
Part 2 of https://github.com/ExpediaGroup/graphql-kotlin/pull/532


smyrick:
Please comment if you think this is not needed. I personally like the change but this does impact existing 1.0.0 users that they will no longer be able to extend the generator through the `generate` method and instead they will have to use the hooks.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/generateSchema.kt
```diff
@@ -0,0 +1,50 @@
...
+package com.expediagroup.graphql.generator
+
+import com.expediagroup.graphql.TopLevelObject
+import com.expediagroup.graphql.generator.types.generateMutations
+import com.expediagroup.graphql.generator.types.generateQueries
+import com.expediagroup.graphql.generator.types.generateSubscriptions
+import graphql.schema.GraphQLSchema
+
+/**
+ * Internal use only. Please use [com.expediagroup.graphql.toSchema] instead.
```
dariuszkuc: whats the point of having this function? wouldn't it make more sense to just add it in to the `toSchema` instead?

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/ClassScanner.kt
```diff
@@ -22,7 +22,7 @@ import io.github.classgraph.ClassInfoList
 import kotlin.reflect.KClass
 import kotlin.reflect.jvm.jvmName
 
-internal class SubTypeMapper(supportedPackages: List<String>) {
+internal class ClassScanner(supportedPackages: List<String>) {
```
dariuszkuc: if you are renaming the class you should rename all of the props below to reflect it as well to avoid the confusion

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/generateSchema.kt
dariuszkuc: guess you kept it separate to allow usage from `federation` but i think it is still pretty confusing

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -17,101 +17,37 @@
...
-import com.expediagroup.graphql.generator.types.generateUnion
 import graphql.schema.GraphQLCodeRegistry
-import graphql.schema.GraphQLSchema
 import graphql.schema.GraphQLType
-import graphql.schema.GraphQLTypeReference
-import graphql.schema.GraphQLTypeUtil
-import kotlin.reflect.KClass
-import kotlin.reflect.KType
 
 open class SchemaGenerator(val config: SchemaGeneratorConfig) {
```
dariuszkuc: this change make the class pretty useless -> it is called schema generator yet it doesn't generate schema at all

smyrick: Yea it is now just a schema generator state. But that is already taken. We could combine  the state and generator into one class?

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/generateSchema.kt
smyrick: I have now made this the only publicly exposed top level function. We need to have this public so we can call it from the federation module. I also don't want to add it onto the generator class as that is now just the state of the current generating schema.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/ClassScanner.kt
smyrick: Done

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
smyrick: I have deleted SchemaGeneratorState

dariuszkuc:
@smyrick with this change `SchemaGenerator` is no longer generating the schema and only holds a collection of states that implies functionality (e.g. keeping track of additional directives to add, etc). I think `generateSchema` function should be part of the generator.

smyrick:
> @smyrick with this change `SchemaGenerator` is no longer generating the schema and only holds a collection of states that implies functionality (e.g. keeping track of additional directives to add, etc). I think `generateSchema` function should be part of the generator.

I agree. Since it needs to be exposed as public anyway it is probably better to contain it in the class.

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/toFederatedSchema.kt
```diff
@@ -38,5 +38,8 @@ fun toFederatedSchema(
     subscriptions: List<TopLevelObject> = emptyList()
 ): GraphQLSchema {
     val generator = FederatedSchemaGenerator(config)
-    return generator.generate(queries, mutations, subscriptions)
+    generator.addExtendedTypes()
```
dariuszkuc: I believe this should be updated to move to `generateSchema` within the `FederatedSchemaGenerator`?

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -18,100 +18,92 @@ package com.expediagroup.graphql.generator
...
-        val graphQLType = state.cache.buildIfNotUnderConstruction(kClass, inputType) { getGraphQLType(kClass, inputType, type) }
-
-        return config.hooks.willAddGraphQLTypeToSchema(type, graphQLType)
+    /**
+     * Clean up the state and saved information after schema generation.
+     *
+     * Not required to call, as the state and subTypeMapper will be cleaned up by the garbage collector,
+     * but it can help clean up memory early if it not being used.
+     */
+    fun close() {
```
dariuszkuc: i think this should be called internally from the `generateSchema`? do we need to expose it?

smyrick: I can move this to generateSchema

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/toFederatedSchema.kt
smyrick: This one I can not move easily with our testing strategy as is.

All the tests in `com.expediagroup.graphql.federation.validation` build a schema to test the `FederatedSchemaValidator` if we add all the types with `@extends` then the schema is no longer valid and it errors out.

Ideally we would not have to call `FederatedSchemaGenerator.generateSchema` to test the validator but that requires changing all the test models to graphql-java fully built schema objects which is a lot more work.

dariuszkuc: If thats the case then I'd drop the `FederatedSchemaGenerator` and make `addExtendedTypes` extension method on the regular `SchemaGenerator`

smyrick: I can't do that because the `ExtendsDirective` is only part of the `graphql-kotlin-federation` package, so there needs to be code there that can reference that annotation. That is why I atleast made the `addAdditionalTypesWithAnnotation` part of the base `SchemaGenerator`

dariuszkuc: I meant the extension should be within the `graphql-kotlin-federation` package (potentially within the same `toFederatedSchema` file.

smyrick: I could do that. It would require though that we expose the `addAdditionalTypesWithAnnotation` as public so we can call it from the extension method. Do we want to to that?

dariuszkuc: so going back to the previous comment -> if I call `FederatedSchemaGenerator.generateSchema` I would expect it to build me a federated schema. With this change it won't be true and requires one to explicitly call the `addExtendedTypes` BEFORE calling `generateSchema`. I understand the reason for that was to make it more testable but it breaks the logic.

smyrick: I agree that it would makes more sense, so how about I make the change to make the testing easier in a separate PR and then we can come back to this one.

smyrick: @dariuszkuc With this change, I will be able to move the addExtendedTypes into the generateSchema function

https://github.com/ExpediaGroup/graphql-kotlin/pull/540

smyrick: Done, this is now in the `generateSchema` method

PR #538, opened 2020-01-06
Update docs about spring reactive

smyrick:
### :pencil: Description
Include a short sentence about spring reactive vs webmvc

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/537

Issue #537, opened 2020-01-06
Dependency on spring-boot-starter-web broke graphql-kotlin-spring-server. Maybe update docs?


viktorchukhantsev:
Hello. I'm tried to create example application(using start.spring.io) with playground and everything and found that dependency on spring-boot-starter-web broke graphql-kotlin-spring-server. Looks like because it uses Tomcat server. I'm newbie with Spring stack. This isn't covered in docs and take me several hours to understand what was broken. Maybe somebody with more experience should cover this situation in documentation?

smyrick:
@viktorchukhantsev I have updated the docs in the above linked PR. Feel free to contribute to the documentation yourself if something is unclear or reference our [examples](https://github.com/ExpediaGroup/graphql-kotlin/tree/master/examples)

viktorchukhantsev:
@smyrick Thank you 👍

dariuszkuc:
Updated docs are available at https://expediagroup.github.io/graphql-kotlin/docs/spring-server/spring-overview

PR #539, opened 2020-01-07
Update spring server overview page

smyrick:
### :pencil: Description
Add some images and more details, copying information from https://medium.com/expedia-group-tech/creating-a-reactive-graphql-server-with-spring-boot-and-kotlin-54aca7316470

### :link: Related Issues
Revamp of https://github.com/ExpediaGroup/graphql-kotlin/pull/538

### 🖼 New look
Screen Shot 2020-01-07 at 10 49 03 AM


EXPErdelfavero:
This change is an improvement on several fronts. I'm proposing an addition to specifically address #537.

There was text in #538 to address the specific problem encountered in #537, which I don't see in any of this. (Possibly my ignorance is showing.) Should there be specific mention of Spring WebMVC and/or Spring Servlet somewhere here, even if it's only pointing out that we're not giving examples of those technologies, and users of those technologies will encounter issues they'll have to resolve for themselves? It's not clear to me whether the proposed changes here would prevent someone else from running into the same issue as #537.

smyrick:
@EXPErdelfavero Added a back a full sentence about the dropping of Spring Web MVC. We did make some assumptions that anyone who has developed on a Spring server knows that "reactive" means they cannot use servlet. Hopefully the above sentence clarifies that without being such a flag. 

We want to save the yellow block `NOTE:` sections for something of higher priority.

EXPErdelfavero:
@smyrick Thanks. I'd review and approve the text, but hesitate because I've just demonstrated my lack of qualification to perform a technical review. :) The text looks fine from an editorial standpoint.

docs/spring-server/spring-overview.md
```diff
@@ -7,6 +7,36 @@ title: Spring Server Overview
 is a Spring Boot auto-configuration library that automatically configures beans required to start up a reactive GraphQL
 web server.
 
+This library is built on a [Spring WebFlux (reactive)](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) stack. By dropping dependencies like [Spring Web MVC (servlet)](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html) we can have a fully non-blocking server.
```
dariuszkuc:
```suggestion
This library is built on a [Spring WebFlux (reactive)](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) stack which is a non-blocking alternative to a traditional [Spring Web MVC (servlet)](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html) based stack. Since both frameworks utilize different threading models they cannot and should not be intermixed. When building a GraphQL server using `graphql-kotlin-spring-server` all your queries and mutations should follow one of the supported [asynchronous execution models](https://expediagroup.github.io/graphql-kotlin/docs/execution/async-models). 
```

PR #540, opened 2020-01-08
Refactor federation validation unit tests

smyrick:
### :pencil: Description
Create new unit tests that call the the federation validation functions directly so we don't have to create a entire schema. This will make it easier to clear out other areas now that we delete these tests

### :link: Related Issues
Helps makes this PR easier once merged to master: https://github.com/ExpediaGroup/graphql-kotlin/pull/535


dariuszkuc:
@smyrick while this performs same validations I think the original integration tests were more meaningful as they were testing out various scenarios using annotated classes that end user could create. By manually creating those scenarios it is not as easy to follow and understand what is being tested or how the actual class would look like in the target app code.

smyrick:
@dariuszkuc I know the existing tests are more end-to-end but they do not work if we want to make `generateGraphQLType` internal because that means we have to call the next method up that is public and that means that all the extended types get added to the schema for this package so the tests all fail. 

One way we would have to fix this is to split up the tests into their own packages and limit the reflection that occurs so that it doesn't pick up all the other invalid types in every test.

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+
+    /**
+     * type Foo @key {
+     *   bar: String
+     * }
+     */
+    @Test
+    fun `validate federated GraphQLObjectType with key directive`() {
```
dariuszkuc:
```suggestion
    fun `validate federated GraphQLObjectType with empty key directive`() {
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
```diff
@@ -0,0 +1,226 @@
...
+        assertEquals(expectedError, result.exceptionOrNull()?.message)
+    }
+
+    /**
+     * type Foo @key {
+     *   bar: String @provides
+     * }
+     */
+    @Test
+    fun `validate federated GraphQLObjectType with key directive and field with provides directive`() {
```
dariuszkuc: should this be testing valid `@key` directive and invalid `@provides` instead?

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
```diff
@@ -0,0 +1,226 @@
...
+        assertEquals(expectedError, result.exceptionOrNull()?.message)
+    }
+
+    /**
+     * type Foo @key {
+     *   bar: String @requires
+     * }
+     */
+    @Test
+    fun `validate federated GraphQLObjectType with key directive and field with requires directive`() {
```
dariuszkuc: should this be valid `@key` and invalid `@requires`?

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
```diff
@@ -0,0 +1,226 @@
...
+        assertEquals(expectedError, result.exceptionOrNull()?.message)
+    }
+
+    /**
+     * type Foo @key {
+     *   bar: String @external
+     * }
+     */
+    @Test
+    fun `validate federated GraphQLObjectType with key directive and field with external directive`() {
```
dariuszkuc: should this be valid `@key`?

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
smyrick: I can do that, it requires just a little moving around to get a value into the directive builder

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateRequiresDirectiveKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        .build()
+
+    private val validatedType: GraphQLObjectType = GraphQLObjectType.newObject()
+        .name("Foo")
+        .field(descriptionField)
+        .build()
+
+    /**
+     * type Foo {
+     *   description: String
```
dariuszkuc:
```suggestion
     *   description: String @requires
```

dariuszkuc: above would still be invalid schema as `@requires` specifies field set, e.g. from old IT
```
    type RequiresOnLocalType @key(fields : "id") {
      description: String!
      id: String!
      shippingCost: String! @requires(fields : "weight")
      weight: Float!
    }
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateRequiresDirectiveKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        assertEquals("base Foo type has fields marked with @requires directive, validatedField=description", errors.first())
+    }
+
+    /**
+     * type Foo @extends {
+     *   description: String
+     * }
+     */
+    @Test
+    fun `extended types run validation on the directive`() {
```
dariuszkuc: unsure what this test is supposed to verify - it checks that field without directive will fail check if directive is specified?

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateRequiresDirectiveKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+package com.expediagroup.graphql.federation.validation
+
+import graphql.Scalars.GraphQLString
+import graphql.schema.GraphQLFieldDefinition
+import graphql.schema.GraphQLObjectType
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+class ValidateRequiresDirectiveKtTest {
```
dariuszkuc: those tests are not as explicit as previous ones

- checks for valid schema
```
    type SimpleRequires @extends @key(fields : "id") {
      description: String!
      id: String! @external
      shippingCost: String! @requires(fields : "weight")
      weight: Float! @external
    }
```
- `@requires` references local field
```
    type RequiresLocalField @extends @key(fields : "id") {
      description: String!
      id: String! @external
      shippingCost: String! @requires(fields : "weight")
      weight: Float!
    }
```
- non-existent field referenced
```
    type RequiresNonExistentField @extends @key(fields : "id") {
      description: String!
      id: String! @external
      shippingCost: String! @requires(fields : "weight")
      weight: Float! @external
    }
```
- base type uses `@requires`
```
    type RequiresOnLocalType @key(fields : "id") {
      description: String!
      id: String!
      shippingCost: String! @requires(fields : "weight")
      weight: Float!
    }
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
smyrick: See above, done

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateRequiresDirectiveKtTest.kt
smyrick: Updating

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
```diff
@@ -0,0 +1,253 @@
...
+            Invalid federated schema:
+             - @provides directive is specified on a Foo.bar field but it does not return an object type
+        """.trimIndent()
+
+        assertEquals(expectedError, result.exceptionOrNull()?.message)
+    }
+
+    /**
+     * type Foo @key(fields: "bar") {
+     *   bar: String @requires
```
dariuszkuc: `@requires` should not be on a `@key` field

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateRequiresDirectiveKtTest.kt
dariuszkuc: while tests below now have proper `@requires` cases they still do not have valid types, e.g.`@extends` without `@key`

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaValidatorTest.kt
```diff
@@ -0,0 +1,253 @@
...
+            Invalid federated schema:
+             - @key directive on Foo is missing field information
+        """.trimIndent()
+
+        assertEquals(expectedError, result.exceptionOrNull()?.message)
+    }
+
+    /**
+     * type Foo @key(fields: "bar") {
+     *   bar: String @provides
```
dariuszkuc: `@provides` should not be a field from `@key`

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateRequiresDirectiveKtTest.kt
smyrick: @dariuszkuc I was only trying to cover all the logic in `validateRequiresDirective` which is really just one conditional. 

After that I test all the other method in their own unit tests so I didn't see the need to tests all possible edge cases through this method.

I can add them though if we want to make these some more integration level tests.

dariuszkuc: @smyrick without doing the proper setup those unit tests are just testing stuff in order to bump the code coverage which seems wrong to me

smyrick: Checking the key directive on the parent type doesn't happen anywhere underneath `validateRequiresDirective` as it only accepts the parent type name for logging purposes.

smyrick: I have moved the valid tests cases to run from `toFederatedSchema` in their own packages. We can only have passing schemas here though because as soon as we include an invalid type marked with `@extend` all test cases will fail

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
```diff
@@ -0,0 +1,202 @@
...
+        assertEquals("foo specifies invalid field set - field set specifies fields that do not exist", errors.first())
+    }
+
+    /**
+     * type Parent @extends {
+     *   foo: String
+     * }
+     */
+    @Test
+    fun `returns an error on extended type without external directive`() {
```
dariuszkuc: i think this test is somewhat weird - i.e. i believe the original test was verifying that key for extended type is referencing local value

```
type Parent @key(fields: "foo") @extends {
  foo: String
}

File: graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt [hunk 11]
```diff
@@ -0,0 +1,202 @@
...
+        assertEquals("foo specifies invalid field set - extended type incorrectly references local field=foo", errors.first())
+    }
+
+    /**
+     * type Parent {
+     *   foo: String @external
+     * }
+     */
+    @Test
+    fun `returns an error on a non extended type with external directive`() {
```
dariuszkuc:
```suggestion
    fun `returns an error on a non local type with external directive`() {
```

dariuszkuc: also missing `@key`

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
```diff
@@ -0,0 +1,202 @@
...
+            extendedType = false,
+            errors = errors,
+            validatedDirective = "foo")
+
+        assertEquals(1, errors.size)
+        assertEquals("foo specifies invalid field set - type incorrectly references external field=foo", errors.first())
+    }
+
+    /**
+     * type Parent @extends {
```
dariuszkuc:
```suggestion
     * type Parent @key(fields: "foo") @extends {
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
```diff
@@ -0,0 +1,202 @@
...
+
+        assertEquals(1, errors.size)
+        assertEquals("foo specifies invalid field set - field set references GraphQLList, field=foo", errors.first())
+    }
+
+    /**
+     * interface MyInterface {
+     *   bar: String
+     * }
+     * type Parent @extends {
```
dariuszkuc:
```suggestion
     * type Parent @key(fields: "foo") @extends {
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
```diff
@@ -0,0 +1,202 @@
...
+            validatedDirective = "foo")
+
+        assertEquals(1, errors.size)
+        assertEquals("foo specifies invalid field set - field set references GraphQLInterfaceType, field=foo", errors.first())
+    }
+
+    /**
+     * union MyUnion = MyType
+     *
+     * type Parent @extends {
```
dariuszkuc:
```suggestion
     * type Parent @key(fields: "foo") @extends {
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
```diff
@@ -0,0 +1,202 @@
...
+            extendedType = true,
+            errors = errors,
+            validatedDirective = "foo")
+
+        assertEquals(1, errors.size)
+        assertEquals("foo specifies invalid field set - field set references GraphQLUnionType, field=foo", errors.first())
+    }
+
+    /**
+     * type Parent @extends {
```
dariuszkuc:
```suggestion
     * type Parent @key(fields: "foo") @extends {
```

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
smyrick: This is validating this conditional check: https://github.com/ExpediaGroup/graphql-kotlin/pull/540/files#diff-25f40334e67210df324a96883dcffa2cR29

graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/validation/ValidateKeySetFieldKtTest.kt
```diff
@@ -0,0 +1,198 @@
...
+        assertEquals("foo specifies invalid field set - extended type incorrectly references local field=foo", errors.first())
+    }
+
+    /**
+     * type Parent @key(fields: "foo") {
+     *   foo: String @external
+     * }
+     */
+    @Test
+    fun `returns an error on a non local type with external directive`() {
```
dariuszkuc:
```suggestion
    fun `returns an error on a local type with external directive`() {
```

PR #541, opened 2020-01-09
BREAKING CHANGE: empty complex types should fail schema generation

dariuszkuc:
### :pencil: Description

GraphQL specification requires Query type to be present as it is required to run introspection query. Per specification it also shouldn't be possible to generate empty complex types (objects, input objects or interfaces) and they should expose at least a single field. Since root Query type is a special GraphQLObjectType it also has to expose at least a single field.

Breaking changes:
* at least single Query is required when generating the schema
* split `didGenerateQueryType` hook (and corresponding `Mutation` and `Subscription` hooks) into `didGenerateQueryFieldType` (previous functionality) and `didGenerateQueryObjectType` hooks to allow more granular control when generating the schema
* default `SchemaGeneratorHooks` now performs validation of generated object types (including special query, mutation and subscription types), input object types and interfaces to ensure we don't generate empty complex object types

see: https://github.com/graphql/graphql-spec/issues/490 and https://github.com/graphql/graphql-spec/issues/568

### :link: Related Issues
None

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
@@ -112,42 +112,42 @@ interface SchemaGeneratorHooks {
     }
 
     /**
-     * Called after converting the function to a field definition but before adding to the schema to allow customization
+     * Called after converting the function to a field definition but before adding to the query object to allow customization
      */
-    fun didGenerateQueryFieldType(kClass: KClass<*>, function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition = fieldDefinition
+    fun didGenerateQueryField(kClass: KClass<*>, function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition = fieldDefinition
 
     /**
-     * Called after converting the function to a field definition but before adding to the schema to allow customization
+     * Called after converting the function to a field definition but before adding to the mutation object to allow customization
      */
-    fun didGenerateMutationFieldType(kClass: KClass<*>, function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition = fieldDefinition
+    fun didGenerateMutationField(kClass: KClass<*>, function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition = fieldDefinition
 
     /**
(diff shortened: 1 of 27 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
```diff
@@ -90,22 +100,58 @@ interface SchemaGeneratorHooks {
     /**
      * Called after wrapping the type based on nullity but before adding the generated type to the schema
      */
-    fun didGenerateGraphQLType(type: KType, generatedType: GraphQLType) = generatedType
+    @Suppress("Detekt.ThrowsCount")
+    fun didGenerateGraphQLType(type: KType, generatedType: GraphQLType): GraphQLType {
```
dariuszkuc: we are validating in `didGenerateGraphQLType` instead of `willAddGraphQLTypeToSchema` hook to allow custom modifications in the latter (including adding new fields).

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
```diff
@@ -90,22 +100,58 @@ interface SchemaGeneratorHooks {
...
+            unwrapped is GraphQLInputObjectType && unwrapped.fieldDefinitions.isEmpty() -> throw EmptyInputObjectTypeException(ktype = type)
+            else -> generatedType
+        }
+    }
 
     /**
      * Called after converting the function to a field definition but before adding to the schema to allow customization
      */
-    fun didGenerateQueryType(kClass: KClass<*>, function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition = fieldDefinition
+    fun didGenerateQueryFieldType(kClass: KClass<*>, function: KFunction<*>, fieldDefinition: GraphQLFieldDefinition): GraphQLFieldDefinition = fieldDefinition
```
smyrick: Do we need to have `type` as a suffix for all of these hooks, or do we just want to call them `didGenerate*Field` `didGenerate*Object`?

dariuszkuc: I just followed the naming of other hooks but can change if it needed

smyrick: If we are going to make a breaking change, I think we should drop the `Type` from this names. It makes sense on the other ones because it referencing the `GraphQLType`, but here `QueryObjectType` is a bit redundant to `QueryObject`

Issue #543, opened 2020-01-11
Authentication with subscriptions


de55:
How would you recommend hooking into the onConnect event for subscriptions, similar to here: 

https://www.apollographql.com/docs/graphql-subscriptions/authentication/

dariuszkuc:
I'd assume it should be handled the same way as for the regular queries, i.e. through GraphQLContext populated by a webfilter that is executed before any query/mutation/subscription execution.

Goregius:
Maybe there could be an example project with authentication (webfilter/directive based) in the examples folder at some point?

de55:
Unlike an http request, the websocket request auth information doesn't come through the headers. Instead, it comes through the initial payload. Right now, I don't see any easy way to access the initial payload information, as the available beans to override in the spring server are:
- ```SubscriptionHandler ```
- ```WebSocketHandlerAdapter```

Neither of these locations allows access to the payload (as far as I'm aware).

What would you think about adding hooks within the ```ApolloSubscriptionProtocolHandler```? We could either just add one in the ```GQL_CONNECTION_INIT.type``` statement (this would be similar to the Apollo OnConnect property) or you could take it a step further and add a hook into every section.
```
   @Suppress("Detekt.TooGenericExceptionCaught")
    fun handle(payload: String, session: WebSocketSession): Flux<SubscriptionOperationMessage> {
        try {
            val operationMessage: SubscriptionOperationMessage = objectMapper.readValue(payload)

            logger.debug("GraphQL subscription client message, sessionId=${session.id} operationMessage=$operationMessage")

            when (operationMessage.type) {
                GQL_CONNECTION_INIT.type -> {
                    // add a hook here to access the message before acknowledging it
                    // the payload property is what we need access to
                    val ackowledgeMessageFlux = Flux.just(acknowledgeMessage)
                    val keepAliveFlux = getKeepAliveFlux(session)
                    return ackowledgeMessageFlux.concatWith(keepAliveFlux)
                }
                GQL_START.type -> return startSubscription(operationMessage, session)
                GQL_STOP.type -> return sessionState.stopOperation(session, operationMessage)
                GQL_CONNECTION_TERMINATE.type -> {
                    sessionState.terminateSession(session)
                    return Flux.empty()
                }
                else -> {
                    logger.error("Unknown subscription operation $operationMessage")
                    sessionState.stopOperation(session, operationMessage)
[...]

PR #544, opened 2020-01-12
Add Apollo Subscription Hooks

de55:
### :pencil: Description
Following the Message Flow section of the spec closer, there should be the following callbacks:

onConnect (this would occur on GQL_CONNECTION_INIT and result should be added to the context for subscriptions)
onOperation (this would occur on GQL_START before running the subscription)
onOperationDone (this would occur on GQL_STOP before? stopping the subscription)
In the Apollo Subscription Server, they have two differences from above:

They call onOperationDone onOperationComplete
They added an onDisconnect event (this would occur on GQL_CONNECTION_TERMINATE before? terminating the session)

### :link: Related Issues
#543


de55:
I wanted to show you the path I am going down and get your thoughts. Need to add/fix tests. Do not merge.

smyrick:
Overall the concept looks good to me. Do we want to call them `ApolloSubscriptionHooks` just to match the hooks concept we have with the schema generator?

de55:
@smyrick I switched the name to ApolloSubscriptionHooks per recommendation.

To ensure that the GraphqlContext is updated in onConnect prior to being supplied to the subscription function, I needed to add the mono for onConnect to the cache.

I added a functional example of the hooks to the spring-example project.

I updated the tests to reflect the changes that I made, but I have not yet added new tests. Prior to that, please review and let me know if there are any additional changes you'd recommend.

@dariuszkuc in the comments of the original issue, you mentioned not wanting to use Mono in the GraphqlContext. The changes I made to the cache, made it so that could be avoided. Jowever, Mono's are exposed as a part of the ApolloSubscriptionHooks interface. The reactor library is already used in other public subscription interfaces, like SubscriptionHandler. But the Monos could instead be switched to a suspend function and then have the calls wrapped in a mono internally, if that is preferred.

examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/execution/MySubscriptionHooks.kt
```diff
@@ -0,0 +1,27 @@
...
+    override fun onConnect(
+        connectionParams: Map<String, String>,
+        session: WebSocketSession,
+        graphQLContext: Any?
+    ): Mono<Unit> = mono {
+        val bearer = connectionParams["Authorization"] ?: "none"
+        val context = graphQLContext as? MyGraphQLContext
+        context?.subscriptionValue = bearer
+//        val jwt =  validator.validate(bearer)
+//        return jwt.map {
```
smyrick: Can we remove commented code if it is not used

graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandlerTest.kt
```diff
@@ -218,20 +220,11 @@ class ApolloSubscriptionProtocolHandlerTest {
...
         val subscriptionHandler: SubscriptionHandler = mockk()
-        val handler = ApolloSubscriptionProtocolHandler(config, subscriptionHandler, objectMapper)
+        val handler = ApolloSubscriptionProtocolHandler(config, subscriptionHandler, objectMapper, subscriptionHooks)
 
-        val initMessage = SubscriptionOperationMessage(GQL_CONNECTION_INIT.type).toJson()
         val stopMessage = SubscriptionOperationMessage(GQL_STOP.type).toJson()
-        val initFlux = handler.handle(initMessage, session)
         val stopFlux = handler.handle(stopMessage, session)
 
-        StepVerifier.create(initFlux)
```
smyrick: This test should not change if we are not using different hooks

smyrick:
@de55 I think exposing the interface with Monos is fine. As you mentioned this is part of the Spring server library where we are using Reactor already. 

As for my other comments, we shouldn't needed to change existing behavior to add hooks so all the valuations we have in our tests should ideally stay the same. If you add some other tests just to validate the hooks are called this overall looks good

de55:
@smyrick It has been a busy week and was just able to come back to this. I added tests for the hooks, and I added documentation.

I am still running into issues with the current tests. Basically, what is happening:

```
        val flux = handler.handle(operationMessage, session)
        val message = flux.blockFirst(Duration.ofSeconds(2)) //the message is successfully read here
        assertEquals(expected = 1, actual = flux.count().block()) // returns 0 and throws error since messsage was read above
```

I've tried unsuccessfully to fix the issue. If I do not make any changes to the prior code, except wrapping the return like so:

```
return Mono.subscriberContext().flatMapMany<SubscriptionOperationMessage> { reactorContext ->
   ...prior code
}
```
I run into the same issue. The code still runs fine using a subscription on a server, as it doesn't need to ever replay. Is there a situation where it would need to replay? Or would you know how to fix the issue above?

smyrick:
@de55 Instead of asserting the flux values by reading can you instead use `reactor.test.StepVerifier`?

de55:
@smyrick do those changes work? Is that what you meant?

smyrick:
Nice, thank you for your contribtutions @de55. These will be included in the 2.0.0 release

PR #547, opened 2020-01-14
[build] migrate to Gradle

dariuszkuc:
### :pencil: Description

Updates build to use gradle composite build - one for core libraries and one for examples. Gradle is much more configurable and allows for dynamic build configuration (i.e. scripting) vs static configuration in Maven. Ability to dynamically to configure various build tasks allows us to build much more powerfull plugins. Gradle also makes few optimizations over Maven with build caches and smarter build tasks dependency graph which should improve the build speed as well.

In order to build library, instead of running
> ./mvnw clean install

now we simply need to run
> ./gradlew clean build

You can see a list of all available tasks with their descriptions by running `gradle tasks`.

Additional changes:
* build is now enforcing JSR305 (Java non-null annotations) in strict mode.
* ktlint is now applied on all projects (including examples)
* both unit and integration tests are run in single test goal, if needed can separate it
* example projects no longer build jars and since we are using composite build we don't need any custom logic to filter those while publishing
* all library dependencies are currently set to API (i.e. maven compile scope), we might update those in the future to IMPLEMENTATION so we don't leak unnecessary dependencies (e.g. classgraph)

This PR sets up the project for the future GraphQL client package that will be provided with both Maven and Gradle plugin. While we technically could still build both plugins using Maven it would not be possible to apply the Gradle plugin on Maven example projects. It is not possible to apply library plugin to a sibling project - we have to use composite build for that (i.e. examples are dependent on built libraries).

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/pull/494


smyrick:
As mentioned directly, we still need to resolve the release scripts in travis

smyrick:
Actually @dariuszkuc do we need to update the master build travis script first?

.github/workflows/continous-integration.yml
```diff
@@ -32,7 +42,8 @@ jobs:
         working-directory: examples
         run: ./gradlew clean build
       - name: Upload coverage to Codecov
-        # currently this doesn't work for forks
+        # currently this doesn't work for PR from forks
+        # see: https://github.com/codecov/codecov-action/issues/29
         if: matrix.java == 8
```
smyrick: Can we add a conditional on if this a fork somehow?

dariuszkuc: I am unaware of any available check that would verify that

dariuszkuc: I am unaware of any available check that would verify it.

EXPErdelfavero: Would this be of any help, or is it at the wrong level of granualarity?
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

EXPErdelfavero: Oh, never mind. I misread the question to be about branches, not forks.

smyrick: What about checking `github.repository` or `github.base_ref`?

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#github-context

EXPErdelfavero: I did read that actions are disabled on forks by default, though. Fourth paragraph down in this section: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#about-github-actions

.github/workflows/continous-integration.yml
```diff
@@ -0,0 +1,51 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - 'docs/**'
+      - 'website/**'
+  pull_request:
```
smyrick: Can this be the following?

```yml
on:
  [push, pull_request]:
    branches:
      - master
    paths-ignore:
      - 'docs/**'
      - 'website/**'
```

EXPErdelfavero:
If someone submits a PR that includes documentation, does the PR process you're running with this change build those docs? This would be useful to catch errors in the doc changes that cause the doc build process to fail. Maybe that's out of scope for this PR, but I thought of it while I was looking at the changes. (I thought it would be a good idea to learn a little about how Github Actions work, and you're the best working example I know of.)

dariuszkuc:
@EXPErdelfavero docs are build using [separate workflow](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/.github/workflows/publish-docs.yml). I'm thinking we could do a PR check that runs [docusaurus-build](https://docusaurus.io/docs/en/commands#docusaurus-build).

.github/workflows/continous-integration.yml
dariuszkuc: Actually this doesn't work.

```
If you need to specify activity types or configuration for an event, you must configure each event separately. You must append a colon (:) to all events, including events without configuration.
```

.github/workflows/continous-integration.yml
dariuszkuc: no luck, tried few combinations in https://github.com/ExpediaGroup/graphql-kotlin/pull/550

Following https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#github-context, tried
- github.repository = ExpediaGroup/graphql-kotlin
- github.ref = `refs/pull/<PR_NUM>/merge`
- github.head_ref = fork branch name (without fork info)
- github.base_ref = master

PR #556, opened 2020-01-16
Restructured configuration properties to work properly with kapt

Goregius:
### :pencil: Description
Properties data classes were moved into GraphQLConfigurationProperties so that kapt can detect the properties and generate the spring metadata file correctly.


Goregius:
Since the structure of the property classes have changed, is this too much of a breaking change?

smyrick:
Do we expect library users to consume these properties? We make them public spring beans but do we expect others to override those beans and not just set the config values from the properties files?

If that is the case then this is only an internal patch change.

Goregius:
The change does mean that to access the inner properties you would have to request for the GraphQLConfigurationProperties class then access the properties within that.

An alternative I found which is to use ConstructorBinding and ConfigurationProperties for all the classes which would keep the structure the same.

```kotlin
/**
 * GraphQL subscription configuration properties.
 */
@ConstructorBinding
@ConfigurationProperties("graphql.subscriptions")
data class SubscriptionConfigurationProperties(
    /** GraphQL subscriptions endpoint, defaults to 'subscriptions' */
    val endpoint: String = "subscriptions",
    /** Keep the websocket alive and send a message to the client every interval in ms. Default to not sending messages */
    val keepAliveInterval: Long? = null
)
```

This would mean having to specify all of them in the spring.factories file:

```properties
org.springframework.boot.autoconfigure.EnableAutoConfiguration=com.expediagroup.graphql.spring.GraphQLAutoConfiguration,com.expediagroup.graphql.spring.SubscriptionConfigurationProperties,com.expediagroup.graphql.spring.PlaygroundConfigurationProperties,com.expediagroup.graphql.spring.SDLConfigurationProperties,com.expediagroup.graphql.spring.FederationConfigurationProperties
```

Would this be preferred?

smyrick:
No, I think the nested structure is better, so we can keep this as is. It will be included as part of the 2.0.0 release

Issue #565, opened 2020-01-17
Update to graphql-java 14


smyrick:
**Describe the solution you'd like**
GraphQL Java released a new major version, 14.0. There are some breaking changes which look like they will impact us and it will not be a simple version update. However there are lots of bug fixes and other issues that we were hoping to be fixed that have been closed in this release.

We should try to get this out in 2.0.0. There shouldn't be any additional breaking changes for our library users but it will be a great finisher to add into this release

**Additional context**
https://github.com/graphql-java/graphql-java/releases/tag/v14.0

smyrick:
I will start a branch on this

Issue #533, opened 2019-12-29
Support for quote descriptions (Docstrings)


Goregius:
**Is your feature request related to a problem? Please describe.**
Currently, the schema generator can only produce descriptions for functions, types, etc. with # comments instead of quote descriptions (with " or """). This is problematic as Apollo servers with Federation strip these # descriptions away but allows quoted descriptions, this is intentional as # comments for descriptions are depreciated.

**Describe the solution you'd like**
Graphql-java has already fixed this but the fix isn't out as a major release yet. Either wait for that to come out, use a daily release version that has this feature (unlikely), or find a way to support it within this library (I don't see how that is possible without making a custom schema printer).

**Additional context**
The changes graphql-java made can be seen in this PR: https://github.com/graphql-java/graphql-java/pull/1609.

smyrick:
This should be fixed with a graphql-java upgrade https://github.com/ExpediaGroup/graphql-kotlin/issues/565

PR #568, opened 2020-01-17
Upgrade to graphql-java 14

smyrick:
### :pencil: Description
Upgrading to graphql-java 14 requires some breaking changes: https://github.com/graphql-java/graphql-java/releases/tag/v14.0

The biggest change is that the SDL descriptions are now printed with `"Comments"` instead of `#Comment`. So we needed to update a few integration tests and our regex in federation that inspects the schema. This PR would make it a lot simpler, but it can be merged later: https://github.com/graphql-java/graphql-java/pull/1756

The other big change is in the underlying schema code and how a `GraphQLType` has been split up into `GraphQLType` and `GraphQLNamedType`, and how some elements of execution use the new type `GraphQLSchemaElement` instead of `GraphQLType`. We were able to hide most of this behind the generator but there are some breaking changes to our library users in the hooks.

### :link: Related Issues
* Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/565
* Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/533

### DIFF
--- a/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
@@ -51,8 +51,8 @@ import kotlin.reflect.full.findAnnotation
  * Hooks for generating federated GraphQL schema.
  */
 open class FederatedSchemaGeneratorHooks(private val federatedTypeRegistry: FederatedTypeRegistry) : SchemaGeneratorHooks {
-    private val directiveDefinitionRegex = "(^#.+$[\\r\\n])?^directive @\\w+.+$[\\r\\n]*".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
-    private val scalarDefinitionRegex = "(^#.+$[\\r\\n])?^scalar (_FieldSet|_Any)$[\\r\\n]*".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
+    private val directiveDefinitionRegex = "(^\".+\"$[\\r\\n])?^directive @\\w+.+\$[\\r\\n]*".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
+    private val scalarDefinitionRegex = "(^\".+\"$[\\r\\n])?^scalar (_FieldSet|_Any)$[\\r\\n]*".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
     private val emptyQueryRegex = "^type Query \\{$\\s+^\\}$\\s+".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
     private val validator = FederatedSchemaValidator()
 
--- a/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
@@ -17,9 +17,14 @@
 package com.expediagroup.graphql.federation
 
 import com.expediagroup.graphql.annotations.GraphQLName
+import com.expediagroup.graphql.directives.DEPRECATED_DIRECTIVE_NAME
 import com.expediagroup.graphql.extensions.print
+import com.expediagroup.graphql.federation.directives.EXTENDS_DIRECTIVE_NAME
+import com.expediagroup.graphql.federation.directives.EXTERNAL_DIRECTIVE_NAME
 import com.expediagroup.graphql.federation.directives.FieldSet
 import com.expediagroup.graphql.federation.directives.KEY_DIRECTIVE_NAME
+import com.expediagroup.graphql.federation.directives.PROVIDES_DIRECTIVE_NAME
+import com.expediagroup.graphql.federation.directives.REQUIRES_DIRECTIVE_NAME
 import com.expediagroup.graphql.federation.directives.extendsDirectiveType
 import com.expediagroup.graphql.federation.execution.EntityResolver
 import com.expediagroup.graphql.federation.execution.FederatedTypeRegistry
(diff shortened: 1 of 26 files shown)


graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
```diff
@@ -44,12 +51,13 @@ import kotlin.reflect.full.findAnnotation
  */
 open class FederatedSchemaGeneratorHooks(private val federatedTypeRegistry: FederatedTypeRegistry) : SchemaGeneratorHooks {
     private val directiveDefinitionRegex = "(^#.+$[\\r\\n])?^directive @\\w+.+$[\\r\\n]*".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
-    private val customDirectivesRegex = "\\s*@(?!extends)(?!external)(?!key)(?!provides)(?!requires)(?!deprecated)\\w+(?:\\(.*(?=\\))\\))?".toRegex(
-        setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
     private val scalarDefinitionRegex = "(^#.+$[\\r\\n])?^scalar (_FieldSet|_Any)$[\\r\\n]*".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
     private val emptyQueryRegex = "^type Query \\{$\\s+^\\}$\\s+".toRegex(setOf(RegexOption.MULTILINE, RegexOption.IGNORE_CASE))
```
smyrick: We just need to update the `directiveDefinitionRegex` and `scalarDefinitionRegex` regex to get a passing build so that they work with the new comment format to use `"Comment"` instead of `#Comment`

This PR would allow us to remove `directiveDefinitionRegex` all together https://github.com/graphql-java/graphql-java/pull/1756

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
```diff
@@ -94,7 +95,7 @@ interface SchemaGeneratorHooks {
      * Called after `willGenerateGraphQLType` and before `didGenerateGraphQLType`.
      * Enables you to change the wiring, e.g. apply directives to alter the target type.
      */
-    fun onRewireGraphQLType(generatedType: GraphQLType, coordinates: FieldCoordinates? = null, codeRegistry: GraphQLCodeRegistry.Builder? = null): GraphQLType =
+    fun onRewireGraphQLType(generatedType: GraphQLSchemaElement, coordinates: FieldCoordinates? = null, codeRegistry: GraphQLCodeRegistry.Builder? = null): GraphQLSchemaElement =
```
smyrick: This needs to be updated due to graphql-java changes and is a breaking change

.gitignore
```diff
@@ -7,6 +7,8 @@ pom.xml.versionsBackup
 # gradle ignore
 .gradle
 build/
+out/
```
dariuszkuc: i dont think it matters but i think it's coming from intellij missing some config - it should use build directory (also no need for `examples/**/out/` as this one should cover it

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
dariuszkuc: since you are using directive filter do we still need the `directiveDefinitionRegex`? I think this would have to handle both single and multiline comments

smyrick: Correct, there is no need for the directive regex now

dariuszkuc: If it is no longer needed please remove it (declaration at line 54 and referenced and line 94).

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/extensions/GraphQLSchemaExtensions.kt
```diff
@@ -34,46 +31,28 @@ import graphql.schema.idl.SchemaPrinter
...
             .includeExtendedScalarTypes(includeExtendedScalarTypes)
-            .includeSchemaDefintion(includeDefaultSchemaDefinition)
+            .includeSchemaDefinition(includeDefaultSchemaDefinition)
             .includeDirectives(includeDirectives)
+            .includeDirectives(includeDirectivesFilter)
+            .descriptionsAsHashComments(descriptionsAsHashComments)
     )
 
-    var schemaString = schemaPrinter.print(this)
-    if (includeDirectives) {
```
dariuszkuc: nice that we could get rid off this nasty workaround

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
smyrick: Actually, I misspoke. We can not remove this regex unless this PR is merged and released. https://github.com/graphql-java/graphql-java/pull/1756

smyrick:
> With the change to comment printing - can we add a test to verify multi line comments?

Done here https://github.com/ExpediaGroup/graphql-kotlin/pull/568/commits/4217ee7ae3f8aca5845e0a9c2100a55d92336766

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/extensions/GraphQLSchemaExtensions.kt
```diff
@@ -34,46 +31,28 @@ import graphql.schema.idl.SchemaPrinter
...
  */
 fun GraphQLSchema.print(
     includeIntrospectionTypes: Boolean = false,
     includeScalarTypes: Boolean = true,
     includeExtendedScalarTypes: Boolean = true,
     includeDefaultSchemaDefinition: Boolean = true,
-    includeDirectives: Boolean = true
+    includeDirectives: Boolean = true,
+    includeDirectivesFilter: Predicate<GraphQLDirective> = Predicate { includeDirectives },
+    descriptionsAsHashComments: Boolean = false
```
dariuszkuc: should we even expose this since this is deprecated?

smyrick: Hmm, good point. That is less we would have to change later. I don't see the need for us to use it at Expedia Group. If someone else needs this to be backwards compatible later after 2.0 we can add it in until GraphQL Java removes it

dariuszkuc: lets remove it then

PR #573, opened 2020-01-23
[test] Add perf test for different data fetchers

smyrick:
### :pencil: Description
Add a perf test to verify different data fetcher performance.

We can have `PropertyDataFetcher` or a `FunctionDataFetcher`. But more specifically on a `FunctionDataFetcher` we may have to parse the arguments with jackson


perf-tests/.nvmrc
```diff
@@ -1 +1 @@
-12
+13
```
dariuszkuc: i believe 12 is the LTS and the next LTS is 14, lets stick to LTS releases

dariuszkuc:
unsure what this is supposed to test...

smyrick:
I was using it to validate the performance difference of data fetchers, more specifically parsing arguments in a function.

We have a investigation in EG of a query running slowly when it has lots of fields that are resolved with functions.

I also updated the script to use npx so yarn is not needed

dariuszkuc:
> I was using it to validate the performance difference of data fetchers, more specifically parsing arguments in a function.
> 
> We have a investigation in EG of a query running slowly when it has lots of fields that are resolved with functions.

Each query (i.e. function) will run on some dispatcher (default to `default` with size limited to number of CPU cores) so unless they are suspending and mostly waiting for the data they will hit the limit of how many things can be processed in parallel. I'd look at the instrumentation details to check how long it takes to resolve each field.

smyrick:
Closing in favor of https://github.com/ExpediaGroup/graphql-kotlin/pull/574

We will do testing locally and don't need to check this in for now.

PR #574, opened 2020-01-23
[test] Update perf test to use npx over yarn

smyrick:
## :pencil: Description
Drop the need for developers to have yarn installed and instead use npx which is included with npm

### DIFF
--- a/.github/workflows/pr-check-docs.yml
+++ b/.github/workflows/pr-check-docs.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - 'docs/**'
       - 'website/**'
+      - 'perf-tests/**'
 
 jobs:
   check-docs:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-check-docs.yml
```diff
@@ -21,9 +22,27 @@ jobs:
...
 
-    - name: Run npm commands
+    - name: Build docs
       if: github.repository == 'ExpediaGroup/graphql-kotlin'
       run: |
         cd website
         npm install
         npm run build
+
+  check-perf-tests:
```
dariuszkuc: why would we want to run perf tests if someone updates docs?

dariuszkuc: maybe we should split those up?

smyrick: We may just be able to delete the perf tests here that use npm. Instead we may want some other client. This was an add in to debug our subscriptions.

smyrick:
Closing in favor of https://github.com/ExpediaGroup/graphql-kotlin/pull/576

PR #579, opened 2020-01-24
Set gradle prop to fix release process

smyrick:
### :pencil: Description
There is a bug with Gradle 6 and Nexus 2 currently that we can fix by setting a systemProp. 

### :link: Related Issues
See here for more info: https://github.com/gradle/gradle/issues/11308


smyrick:
I found this issue from here: https://github.com/Codearte/gradle-nexus-staging-plugin/issues/110

If this fixes it I will comment back as well saying this solved for us.

PR #583, opened 2020-01-30
Add example stress query

smyrick:
## :pencil: Description
Add a `StressQuery` that will create a lot of nodes so we can stress test data fetchers with a single query instead of lots of requests. 

This will help us debug performance overhead around `PropertyDataFetcher` vs `FunctionDataFetcher` vs `suspend FunctionDataFetcher`

## Test Results

I used YourKit locally to get a snapshot of each query. The files are too big to upload but I am looking at the results right now

### Property
```graphql
query stress {
  stressNode(count:100000) {
    valueId
  }
}
```

### Functions
```graphql
query stress {
  stressNode(count:100000) {
    functionId
  }
}
```

### Suspend Functions
```graphql
query stress {
  stressNode(count:100000) {
    suspendId
  }
}
```

### DIFF
--- a/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/directives/TrackTimesInvoked.kt
+++ b/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/directives/TrackTimesInvoked.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2020 Expedia, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


smyrick:
I looked at all the snapshots for the time to execute `SimpleQueryHandler::executeQuery` since this is run for each request and ignores all the started up time and the serialization/server time.

Here are the times in `ms`

| valueId | functionId | suspendId |
| ------- | ---------- | ----------- |
| 2681 | 2781 | 3400 |
| 0% | +4% | +27% |
 

The difference between a `FunctionDataFetcher` and `PropertyDataFetcher` does exist but it is not the major concern. There is significant overhead to run suspending functions when we create a lot of them though from the baseline of a just returning properties.

smyrick:
## Instrumentation
I added two more queries which add the `TrackTimesInvokedInstrumentation`. This track the total number of times each field that is marked with the directive is invoked for every request.

Running a query like this logs the following on the server `TrackTimesInvokedInstrumentation fields invoked: {loggingFunctionId=10, suspendLoggingFunctionId=10}`
```graphql
query stress {
  stressNode(count:10) {
    valueId
    loggingFunctionId
    suspendLoggingFunctionId
  }
}
```
## Queries
#### Function Instrumentation
```graphql
query stress {
  stressNode(count:100000) {
    suspendLoggingFunctionId
  }
}
```

#### Suspend Function Instrumentation
```graphql
query stress {
  stressNode(count:100000) {
    loggingFunctionId
  }
}
```

## Performance

Issue #584, opened 2020-01-31
Customizable list classes


de55:
I'd like to be able to make it so that ```Collection<T>``` is treated equivalently to ```List<T>```. Is there any possible way to do that now with the hooks? And if not, would we be able to make this piece configurable:

```
internal fun KClass<*>.isListType(): Boolean = this.isSubclassOf(List::class) || this.java.isArray
```

de55:
On a related note, I think this would be possible with the hooks if there was a way to add extra types to be generated. Is that something you would be open to adding (or that already exists)? The issue right now is that if I were to add ```Collection::class``` to 
```
override fun willGenerateGraphQLType(type: KType): GraphQLType? 
```
then I have to make sure that ```<T>``` exists, which I don't think I can do if the only place it is added to the schema is as ```Collection<T>```

de55:
> On a related note, I think this would be possible with the hooks if there was a way to add extra types to be generated. Is that something you would be open to adding (or that already exists)? The issue right now is that if I were to add `Collection::class` to
> 
> ```
> override fun willGenerateGraphQLType(type: KType): GraphQLType? 
> ```
> 
> then I have to make sure that `<T>` exists, which I don't think I can do if the only place it is added to the schema is as `Collection<T>`

I confirmed that I would be able to accomplish what I am trying to if I could add a list of types to be generated. That is because I could then do this:

```
    override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier) {
        Collection::class -> GraphQLList.list(GraphQLTypeReference.typeRef(type.getTypeOfFirstArgument().getSimpleName()))
        else -> super.willGenerateGraphQLType(type)
    }
    override fun isValidSuperclass(kClass: KClass<*>): Boolean {
        return if(kClass == Collection::class) false else  super.isValidSuperclass(kClass)
    }
```
That code works for sets/collections/etc. right now, as long as the underlying object within the collection will eventually be built.

dariuszkuc:
See https://github.com/ExpediaGroup/graphql-kotlin/issues/201

TLDR reason we only support list is due to deserialization issues, i.e. it is not possible to enforce set properties in JSON.

de55:
I understand why you don't broadly support it, but I'd still rather set it up. I also have other reasons that I'd like to add types, like adding extension functions. I had the extension functions working, but some of the 2.0 changes broke my current implementation.

The change to be able to add additional types would be simple see #585.

PR #585, opened 2020-01-31
Add the ability to add additional types to the schema

de55:
### :pencil: Description
Add the ability to add additional types to the schema

### :link: Related Issues
#584

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
@@ -154,6 +154,11 @@ interface SchemaGeneratorHooks {
         type
     }
 
+    /**
+     * Called after generating the all objects but before adding directives.
+     */
+    fun onAddAdditionalTypes(): Set<KType> = setOf()
+
     val wiringFactory: KotlinDirectiveWiringFactory
         get() = KotlinDirectiveWiringFactory()
 }
--- a/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/extension/CustomSchemaGeneratorHooks.kt
+++ b/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/extension/CustomSchemaGeneratorHooks.kt
@@ -16,46 +16,69 @@
 
 package com.expediagroup.graphql.examples.extension
 
+import com.expediagroup.graphql.annotations.GraphQLName
 import com.expediagroup.graphql.directives.KotlinDirectiveWiringFactory
+import com.expediagroup.graphql.examples.model.AdditionalObject
 import com.expediagroup.graphql.hooks.SchemaGeneratorHooks
 import graphql.language.StringValue
 import graphql.schema.Coercing
+import graphql.schema.GraphQLList
 import graphql.schema.GraphQLScalarType
 import graphql.schema.GraphQLType
+import graphql.schema.GraphQLTypeReference
 import java.util.UUID
+import kotlin.reflect.KClass
 import kotlin.reflect.KType
+import kotlin.reflect.full.findAnnotation
+import kotlin.reflect.full.starProjectedType
+import kotlin.reflect.jvm.jvmErasure
 
 /**
  * Schema generator hook that adds additional scalar types.
  */
 class CustomSchemaGeneratorHooks(override val wiringFactory: KotlinDirectiveWiringFactory) : SchemaGeneratorHooks {
 
+    private val additionalTypes = mutableSetOf<KType>()
     /**
-     * Register additional GraphQL scalar types.
+     * Register additional GraphQL types.
      */
     override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier) {
         UUID::class -> graphqlUUIDType
+        Set::class -> {
+            val nestedType = getNestedType(type)
+            additionalTypes.add(nestedType)
+            GraphQLList.list(GraphQLTypeReference.typeRef(getSimpleName(nestedType.jvmErasure)))
+        }
         else -> null
     }
-}
-
-internal val graphqlUUIDType = GraphQLScalarType.newScalar()
-    .name("UUID")
-    .description("A type representing a formatted java.util.UUID")
-    .coercing(UUIDCoercing)
-    .build()
-
-private object UUIDCoercing : Coercing<UUID, String> {
-    override fun parseValue(input: Any?): UUID = UUID.fromString(
-            serialize(
-                    input
-            )
-    )
-
-    override fun parseLiteral(input: Any?): UUID? {
-        val uuidString = (input as? StringValue)?.value
-        return UUID.fromString(uuidString)
+    override fun isValidSuperclass(kClass: KClass<*>): Boolean {
+        return if(kClass == Set::class) false else  super.isValidSuperclass(kClass)
     }
+    override fun onAddAdditionalTypes() = setOf(AdditionalObject::class.starProjectedType)
 
-    override fun serialize(dataFetcherResult: Any?): String = dataFetcherResult.toString()
-}
+    private fun getNestedType(type: KType): KType = type.arguments.firstOrNull()?.type ?: throw IllegalArgumentException("Type is not a list")
+
+    private fun getSimpleName(kClass: KClass<*>): String = kClass.findAnnotation<GraphQLName>()?.value
+        ?: kClass.simpleName
+        ?: throw IllegalArgumentException("Could not get graphql name")
+
+    internal val graphqlUUIDType = GraphQLScalarType.newScalar()
+        .name("UUID")
+        .description("A type representing a formatted java.util.UUID")
+        .coercing(UUIDCoercing)
+        .build()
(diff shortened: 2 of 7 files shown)


de55:
The three purposes I have for this are:
- Adding extension functions to the schema
- Adding different types of lists to the schema
- Automatically converting Map into List<KeyValuePair>

I understand that you don't want extension functions or Set<T>/Collection<T> as a part of the general library, but this enhancement would allow me to implement them through the hooks.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt
```diff
@@ -154,6 +154,11 @@ interface SchemaGeneratorHooks {
         type
     }
 
+    /**
+     * Called after generating the all objects but before adding directives.
+     */
+    fun onAddAdditionalTypes(): Set<KType> = setOf()
```
smyrick: This really isn't a hook. Maybe it should be something we can pass in through the config. If we merged this as a hook it would only run once and just returns the value it has.

@de55 Would adding a new value to `SchemaGeneratorConfig` for the additional types work for you?

de55: @smyrick I agree with you on it not really being a hook. The reason I didn't initially put it on the hooks is that the config depends on the hooks and (at least for me) the additional types depend on the hooks. So to pass the additional types into the config, you end up with a bit of a circular dependency.

That being said I think something like this would work as I could pass the provider into both the config & the hooks, and then dynamically build it before provider.additionalTypes() is called:

```
open class SchemaGeneratorConfig(
    open val supportedPackages: List<String>,
    open val topLevelNames: TopLevelNames = TopLevelNames(),
    open val hooks: SchemaGeneratorHooks = NoopSchemaGeneratorHooks,
    open val dataFetcherFactoryProvider: KotlinDataFetcherFactoryProvider = SimpleKotlinDataFetcherFactoryProvider(),
    open val additionalTypesProvider: AdditionalTypesProvider
)

interface AdditionalTypesProvider{
    fun additionalTypes() = setOf<KType>()
}
```

smyrick: Ah, ok I see the issue now. 

Here you have to manually unwrap the Set and if you don't add the wrapped type to the schema it won't work, so you need to dynamically update the additional types. That means we can have it passed as a static value in config

https://github.com/ExpediaGroup/graphql-kotlin/pull/585/files#diff-6ea5b1ef9b830b4dcad7a6d5ddf4b426R49

This seems like the best approach then. Looking at the implementation we have for `SchemaGenerator::additionalTypes`, we have it as a `List<GraphQLType>` and you have to call `generateGraphQLType` again for these. I wonder if it would be easier if we made the a `List<KType>` as we are generating the schema, and then just generate all the additional types, custom and normal all together at the end.

Let me throw together a quick branch...

smyrick: @de55 With this PR here we can change the call to merge the additional types saved and the ones from hooks and just call the private method

```kotlin
builder.additionalTypes(generateAdditionalTypes(additionalTypes.addAll(hookAddtionalTypes)))
```

https://github.com/ExpediaGroup/graphql-kotlin/pull/587

dariuszkuc: Additional types should not be part of the hooks but rather a config.

de55: @dariuszkuc @smyrick I think for my use case, things look cleaner from the hooks, but I'm not sure how others would use so I can go either way. If it is set up similar to above with ```AdditionalTypeProvider```, I can pass the the ```AdditionalTypeProvider``` into the ```SchemaGeneratorHooks``` & the ```SchemaGeneratorConfig``` and add to/update it from the ```SchemaGeneratorHooks```. Just let me know which method you want (hooks or config) and I'll update the PR.

de55: Another potential option would be implementing a hook that is something like 
```fun onCustomUnwrap(type: KType): List<KType>```.

You could call it right after:

```
internal fun generateGraphQLType(generator: SchemaGenerator, type: KType, inputType: Boolean = false, annotatedAsID: Boolean = false): GraphQLType {
    val hookGraphQLType = generator.config.hooks.willGenerateGraphQLType(type)
    val graphQLType = hookGraphQLType
        ?: generateScalar(generator, type, annotatedAsID)
        ?: objectFromReflection(generator, type, inputType)
```

The hook would then take the returned types and call ```internal fun generateGraphQLType`` for each of them.

Comment (smyrick) [reply graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt]: @de55 What if with my change here I made `generateAdditionalTypes` protected?

Then you could implement a custom `SchemaGenerator` and override the `generateSchema` method to do all the additional things you want.

Comment (de55) [reply graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt]: @smyrick That works.

Comment (de55) [reply graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt]: @smyrick in #587 you said:

"I have undone making the new method open for calling in protected scope and instead just made it private. We are still discussing on the team if we do actually want developers to use this custom behaviour or are we going against GraphQL best practices.

If we wanted to open it to protected scope later that is a simple PR."

Does that mean you landed against this change and any type of change related to this topic?

Comment (smyrick) [reply graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt]: @de55 After further discussion with some of the other maintainers, the changes on our end are fine. We just have some hesitation on what you are doing with your schema. 

At Expedia Group, we have found it is better to have the schema be declared as a matching Kotlin object to exactly what GraphQL can validate with its type system. Sets and Maps are not going to be guaranteed by your clients. If you want to see a talk from me about schema design and why and this library was created, see this talk from Apollo Day SF 2019: https://www.youtube.com/watch?v=MuD3TAP0D9Y

That being said, opening up our generator to add more additional types is not a bad thing and does allow for other uses cases. So I will create that new PR to make this protected and you can update you changes.

cc @dariuszkuc

Comment (de55) [reply graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/hooks/SchemaGeneratorHooks.kt]: Thanks @smyrick I appreciate that. The biggest use case we have is adding extensions functions and that is because we automatically generate parts of our schema.

On adding types for sets/maps, that is something we newly discussed doing, and I'd be interested to hear what your recommended approach is when the data truly is a set or a map.

On our team we've found that we are using the following patterns frequently:

```
class MySetClass(@GraphQLIgnore val pUniqueSet: Set<String>){
    val uniqueSet by lazy {pUniqueSet.toList()}
}
```

```
data class DateDecimalKeyValuePair(val date: LocalDate, val amount: BigDecimal)
class MyMapClass(@GraphQLIgnore val pMap: Map<LocalDate, BigDecimal>){
    val map by lazy {pMap.map {DateDecimalKeyValuePair(it.key, it.value)  }}
}
```

I understand that the client will see these as a list either way, but by automatically converting the data type we can:
- automatically add detail to the schema descriptions
- avoid writing the boiler plate code above
- create more meaningful names for KeyValuePairs without having to generate additional classes (using annotations to rename the graphql type).

dariuszkuc: re: sets - there is nothing wrong with using sets within your codebase and automatically convert it to a `GraphQLList` for output types. It becomes problematic if you plan to use sets as an input objects - i.e. in JSON you cannot enforce set properties so your deserialization logic would have to decide whether to drop duplicates or throw exception. We don't support sets out of the box because of those deserialization issues as we think the behavior there should be determined by the code owners. 

re: maps - problem with automatic conversion is that those key/value pairs can generally be better exposed as some very specific type instead. What is that `DateDecimalKeyValuePair ` supposed to represent? It looks like it is tracking a date and some money amount (based on the `BigDecimal` I assume some money) - maybe it would be better represented as `DailySaleTotals` or something like that. There are number of arguments for/against supporting maps in GraphQL spec (see https://github.com/graphql/graphql-spec/issues/101) but so far general consensus is against it.

smyrick:
@de55 With the following changes is your PR required?  Or can you try out using the new open method and overriding the set before calling the super method to generate the types

https://github.com/ExpediaGroup/graphql-kotlin/pull/593

de55:
@smyrick it is not. I'll close this. Thanks for opening those methods up and for the video/discourse.

PR #587, opened 2020-01-31
Allow additional types to be added with custom SchemaGenerator

smyrick:
### :pencil: Description
This is a patch change to allow additional types to more easily be extended in later PRs. As a side-effect of this change we actually simplify where types are all generated and the main generateSchema function becomes just a little more straight forward.

### :link: Related Issues
This will make https://github.com/ExpediaGroup/graphql-kotlin/pull/585 a little eaiser to implement

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -19,7 +19,6 @@ package com.expediagroup.graphql.generator
 import com.expediagroup.graphql.SchemaGeneratorConfig
 import com.expediagroup.graphql.TopLevelObject
 import com.expediagroup.graphql.directives.DeprecatedDirective
-import com.expediagroup.graphql.extensions.unwrapType
 import com.expediagroup.graphql.generator.state.ClassScanner
 import com.expediagroup.graphql.generator.state.TypesCache
 import com.expediagroup.graphql.generator.types.generateGraphQLType
@@ -31,7 +30,6 @@ import graphql.schema.GraphQLCodeRegistry
 import graphql.schema.GraphQLDirective
 import graphql.schema.GraphQLSchema
 import graphql.schema.GraphQLType
-import graphql.schema.GraphQLTypeReference
 import java.util.concurrent.ConcurrentHashMap
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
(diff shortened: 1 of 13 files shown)


smyrick:
I am not adding that functionality in this PR. That will be done in https://github.com/ExpediaGroup/graphql-kotlin/pull/585

As seen there, having a way to dynamically update the `additionalTypes` would be useful. Are you proposing we add a new method to the `SchemaGenerator` to just call `addAdditionalType` from the hooks? That would mean we have a circular dependency on the hooks and the schema generator. I prefer the current approach more.

dariuszkuc:
I was thinking more along the lines of exposing the functionality so one could extend the `SchemaGenerator` (similar to what `FederatedSchemaGenerator` does) and call the method to add the types. We can go with the route of more generic support by adding it through the config.

smyrick:
@dariuszkuc I updated the PR. Now the method `generateAdditionalTypes` is protected so someone can call it from a custom schema generator. This will allow them to keep track of any additional state and still have access to the Schema generator properties.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -94,10 +91,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
-            .forEach {
-                additionalTypes.add(it)
-            }
+        classScanner.getClassesWithAnnotation(annotation).forEach {
+            additionalTypes.add(it.createType())
+        }
     }
+
+    /**
+     * Gererate the GraphQL type for all the additional types. They are generated as non-inputs and not as IDs.
```
dariuszkuc:
```suggestion
     * Generate the GraphQL type for all the additional types. They are generated as non-inputs and not as IDs.
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -94,10 +91,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
+     * If you need to provide more custom additional types that were not picked up from reflection of the schema objects,
```
dariuszkuc:
```suggestion
     * If you need to provide custom additional custom types that were not picked up from reflection of the schema objects,
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -94,10 +91,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
+     * you modify the set before you call this method.
```
dariuszkuc:
```suggestion
     * you can modify the `additionalTypes` set before you call this method.
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
smyrick: Removed java doc as it is internal for now

smyrick:
I have undone making the new method open for calling in protected scope and instead just made it private. We are still discussing on the team if we do actually want developers to use this custom behaviour or are we going against GraphQL best practices.

If we wanted to open it to protected scope later that is a simple PR.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateGraphQLType.kt
```diff
@@ -58,9 +58,11 @@ private fun objectFromReflection(generator: SchemaGenerator, type: KType, inputT
     }
 
     val kClass = type.getKClass()
-    val graphQLType = generator.cache.buildIfNotUnderConstruction(kClass, inputType) { getGraphQLType(generator, kClass, inputType, type) }
 
-    return generator.config.hooks.willAddGraphQLTypeToSchema(type, graphQLType)
```
smyrick: We do not want to call the hook with type references

PR #588, opened 2020-01-31
Change scope of FunctionDataFetcher methods

smyrick:
### :pencil: Description
In #582 we opened some method of the FunctionDataFetcher so we could run some custom behaviour. Looking back at that change I realize that we only need to make the methods protected, not open, and that we can hide some others still as private. We just need to expose all the methods that we call in the basic implementation of get as protected so we can all those from custom implementations.

### :link: Related Issues
#582

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
@@ -72,7 +72,7 @@ open class FunctionDataFetcher(
     /**
      * Iterate over all the function parameters and map them to the proper input values from the environment
      */
-    protected open fun getParameterValues(fn: KFunction<*>, environment: DataFetchingEnvironment): Array<Any?> = fn.valueParameters
+    protected fun getParameterValues(fn: KFunction<*>, environment: DataFetchingEnvironment): Array<Any?> = fn.valueParameters
         .map { param -> mapParameterToValue(param, environment) }
         .toTypedArray()
 
@@ -86,7 +86,7 @@ open class FunctionDataFetcher(
      *
      *   - The entire environment is returned if the parameter is of type [DataFetchingEnvironment]
      */
-    protected open fun mapParameterToValue(param: KParameter, environment: DataFetchingEnvironment): Any? =
+    private fun mapParameterToValue(param: KParameter, environment: DataFetchingEnvironment): Any? =
         when {
             param.isGraphQLContext() -> environment.getContext()
             param.isDataFetchingEnvironment() -> environment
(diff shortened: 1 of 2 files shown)


smyrick:
This is a major change but there are no full releases out there that will make this backwards incompatible

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
```diff
@@ -98,7 +98,7 @@ open class FunctionDataFetcher(
      *
      * This is currently achieved by using a Jackson [ObjectMapper].
      */
-    protected open fun convertParameterValue(param: KParameter, environment: DataFetchingEnvironment): Any? {
+    private fun convertParameterValue(param: KParameter, environment: DataFetchingEnvironment): Any? {
```
dariuszkuc: by keeping it open we could allow others to implement some custom logic here, e.g. how to handle file uploads or skip converting the type if underlying object is already a correct type

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
```diff
@@ -86,7 +86,7 @@ open class FunctionDataFetcher(
      *
      *   - The entire environment is returned if the parameter is of type [DataFetchingEnvironment]
      */
-    protected open fun mapParameterToValue(param: KParameter, environment: DataFetchingEnvironment): Any? =
+    private fun mapParameterToValue(param: KParameter, environment: DataFetchingEnvironment): Any? =
```
dariuszkuc: by keeping it open one could technically implement some other autowired object besides `GraphQLContext` and `DataFetchingEnvironment`

smyrick: That is interesting. Do wee need to leave this method in particular open or do we just have them re-implement `get` and call their own custom method.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
smyrick: Did not think about that possibility. Let's discuss in more about both on the thread below.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
dariuszkuc: Since other methods are already open I'd keep those open as well. Once (or IF) we implement Kotlin native way of executing the queries (and drop the dependency on `CompletableFuture`) we could change the interface method signatures to something more appropriate.

smyrick:
I will close this in favor of updating docs

PR #591, opened 2020-02-01
[generator] deprecated directive cleanup

dariuszkuc:
### :pencil: Description
`graphql-java` now adds `@deprecated` directive definition automatically so we dont have to explicitly add it anymore. This change also fixes handling of multiline directive definition declaration in the federation library.

I verified locally that our federation example still works fine with deprecated query. Unsure if this integration test (i.e. starting up 2 SpringBoot apps and node gateway and then running HTTP request) is something we can automate. Posted a question about it on Github forum: https://github.community/t5/GitHub-Actions/integration-test-that-starts-multiple-apps/m-p/44918#M5969

### :link: Related Issues
Related issue: https://github.com/graphql-java/graphql-java/issues/1598


smyrick:
Nice!

PR #593, opened 2020-02-03
Allow updating custom additional types

smyrick:
### :pencil: Description
As discussed in other PRs #587 and #585 we can allow developers to have more customization of the schema by allowing them to add additional types of their own without having to add it to the schema directly. What they do with this feature will be up to them, but from a library perspective it is not bad feature to support.

### :link: Related Issues
See my comment here for more details: https://github.com/ExpediaGroup/graphql-kotlin/pull/585#discussion_r374316867


smyrick:
Actually this may not allow for someone to override the `get` method properly because they don't have access to all the `generate*` methods.

Do we instead need to make this method `open`?

dariuszkuc:
Yes. In order to allow easy extension `generateAdditionalTypes` should be `open`.

EXPErdelfavero:
Does this need an updated doc or sample?

smyrick:
@EXPErdelfavero Yes, it does. Thank you for keep on top of that for us. We need to get better at checking docs, especially if there is a new feature label.

PR #594, opened 2020-02-03
Add docs for new schema generator feature

smyrick:
## :pencil: Description
Update docs with new page describing the features

### :link: Related Issues
Added new features in https://github.com/ExpediaGroup/graphql-kotlin/pull/593 that we did not document


docs/customizing-schemas/deprecating-schema.md
```diff
@@ -1,6 +1,6 @@
 ---
-id: evolving-schema
-title: Evolving Schema
+id: deprecating-schema
```
dariuszkuc: lets keep it as more generic evolving schema and add notes about adding new fields etc

smyrick: I was mainly doing this to help the sidebar be a little more clear. Right now it is not super descriptive for what is there.

Do you think we can just include this all on one page?


Screen Shot 2020-02-03 at 2 48 46 PM

EXPErdelfavero: The only info in the topic right now is on deprecating schema elements, so I'd recomment the title: Deprecating Schema Elements.

If or when you talk about other ways to evolve the schema or its elements, I'd change to something like Evolving Schema.

Just my two cents.

Issue #596, opened 2020-02-08
Unable to use Reactor in an async model


johanflint:
Thank you for building this elegant library, it seems to suit our needs well as it written in Kotlin, runs on Spring Boot Webflux and offers a code-first approach to GraphQL.

Our architecture is based on reactive streams using Reactor, so I'm happy to [read](https://expediagroup.github.io/graphql-kotlin/docs/execution/async-models) that the library supports that. However, I cannot seem to get it to work properly.

As an example I wrote the following:
```kotlin
@Component
class MyAwesomeQuery : Query {
    fun widgetById(id: Int): Mono<Widget> = Mono.just(Widget(id, "value"))
}

data class Widget(val id: Int, val value: String)
```

If I execute the a query including the `id` and `value` I get the following response:
```json
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable type: 'Int' within parent 'Widget' (/widgetById/id)",
      "path": [
        "widgetById",
        "id"
      ],
      "errorType": "DataFetchingException",
      "locations": null,
      "extensions": null
    }
  ]
}
```

However, if I query only the `value`, I get this:
```json
{
  "data": {
    "widgetById": {
      "value": "Widget(id=7, value=value)"
    }
  }
}
```

The reason is that the `Mono` is not unwrapped properly but returned as is (`value` is the property in which the value of `Mono` is stored).

Here is the incomplete `SchemaGeneratorHooks` that I tried, but without this I'm still getting the same response.

```kotlin
@SpringBootApplication
class DomainModelApplication {
	@Bean
	fun hooks() = MonadHooks()
}

class MonadHooks : SchemaGeneratorHooks {
    override fun willResolveMonad(type: KType): KType = when (type.classifier) {
        Mono::class -> type.arguments.firstOrNull()?.type
        else -> type
    } ?: type
}
```

I know that I can return a `CompletableFuture<Widget>` instead, but that seems to be [blocking](https://stackoverflow.com/questions/58504527/is-mono-tofuture-blocking) and it is nice to use the same abstraction (Reactor) everywhere. 

What am I doing wrong here? Thank you for taking the time to read my question!

dariuszkuc:
Thanks for raising this. It appears that with recent refactorings we broke monad support, i.e.

```kotlin
fun delayedEchoUsingReactorMono(msg: String, delaySeconds: Int): Mono<String> = Mono.just(msg)
```

returns

```json
{
  "data": {
[...]

PR #597, opened 2020-02-08
Add new hooks for mapping the function returned value

smyrick:
### :pencil: Description
In the master branch it looks, like we actually don't have a way to resolve custom function wrapper types. The schema can be generated but the execution does not allow for custom behaviour. 

* Rename `willResolveMonad` to `willResolveFunctionType`
* Add new hook `willReturnFunctionResult` which allows developers to fetch the result of their custom function wrapper.

### :link: Related Issues
This is a first draft change related to #596

### DIFF
--- a/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/extension/CustomSchemaGeneratorHooks.kt
+++ b/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/extension/CustomSchemaGeneratorHooks.kt
@@ -22,6 +22,7 @@ import graphql.language.StringValue
 import graphql.schema.Coercing
 import graphql.schema.GraphQLScalarType
 import graphql.schema.GraphQLType
+import reactor.core.publisher.Mono
 import java.util.UUID
 import kotlin.reflect.KType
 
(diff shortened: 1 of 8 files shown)


smyrick:
I think I am like the idea more about having `SchemaExecutionHooks`. We could instead of making some of the classes open like the `FunctionDataFetcher` we could keep them internal but still provide the custom behaviour. And I like how the hooks make our API simpler too.

dariuszkuc:
> I think I am like the idea more about having `SchemaExecutionHooks`. We could instead of making some of the classes open like the `FunctionDataFetcher` we could keep them internal but still provide the custom behaviour. And I like how the hooks make our API simpler too.

I like the idea of execution hooks (similar to what we do for subscription).... but we are currently bound to a `DataFetcher` interface from `graphql-java` which have no idea about any of this. If we ever create our own kotlin native execution model then I'm all for it, as it is right now I am unsure if it is needed.

Separating schema hooks and execution hooks makes sense but also requires configuring two objects in order to enable support for monads. This is not ideal either.

dariuszkuc:
Also a side note -> execution hooks are conceptually the same as the instrumentation....

dariuszkuc:
I looked into existing `graphql-java` instrumentation and it doesn't look like it will work in this case. I believe the proper solution that explicitly defines the hooks/instrumentation is the implmenetation of Kotlin native execution of GraphQL queries. Until then I think the simplest workaround is to extend `FunctionDataFetcher` and explicitly add the conversion.

As I mentioned before, doing hooks in the current state will be easily broken if one creates their own custom data fetcher (e.g. one might want to do it so they can specify their custom coroutine context). Another alternative is to update our `KotlinDataFetcherFactoryProvider` to return our own custom factories that create Kotlin specific `DataFetcher`s but that seems like an overkill for now.

smyrick:
Closing in favor of overriding the data fetcher as documented here https://github.com/ExpediaGroup/graphql-kotlin/pull/599

PR #611, opened 2020-02-23
Refactor to remove build warnings

smyrick:
### :pencil: Description
Refactor library and test code to remove some build and IDE warnings. This is a non-functional change.

### :link: Related Issues
N\A

### DIFF
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
@@ -144,13 +144,21 @@ class ApolloSubscriptionProtocolHandler(
     }
 
     private fun onInit(operationMessage: SubscriptionOperationMessage, session: WebSocketSession, graphQLContext: Any?): Flux<SubscriptionOperationMessage> {
-        val connectionParams = operationMessage.payload as? Map<String, String> ?: emptyMap()
+        val connectionParams = convertPayloadToMap(operationMessage.payload)
         val onConnect = subscriptionHooks.onConnect(connectionParams, session, graphQLContext)
         sessionState.saveOnConnectHook(session, onConnect)
         val acknowledgeMessageFlux = Flux.just(acknowledgeMessage)
         val keepAliveFlux = getKeepAliveFlux(session)
         return acknowledgeMessageFlux.concatWith(keepAliveFlux)
     }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun convertPayloadToMap(payload: Any?): Map<String, String> =
+        if (payload is Map<*, *>) {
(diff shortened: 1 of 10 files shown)


dariuszkuc:
> The changes for the RouteConfiguration are there to provide more internal documentation and suppress the warning we had when calling jackson to read the post body from a suspend function. That is no longer relevant if we move it to its own private function.

Warning was valid - reading body is a blocking operation. Moving this code around doesn't change this. Personally I am also unsure whether the new refactored version of `RouteConfiguration` is that much easier to follow (i.e. you got to jump through multiple methods to see what is happening).

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
```diff
@@ -144,14 +144,22 @@ class ApolloSubscriptionProtocolHandler(
...
         val onConnect = subscriptionHooks.onConnect(connectionParams, session, graphQLContext)
         sessionState.saveOnConnectHook(session, onConnect)
         val acknowledgeMessageFlux = Flux.just(acknowledgeMessage)
         val keepAliveFlux = getKeepAliveFlux(session)
         return acknowledgeMessageFlux.concatWith(keepAliveFlux)
     }
 
+    @Suppress("UNCHECKED_CAST")
+    private fun convertPayloadToMap(payload: Any?): Map<String, String> =
+        if (payload is Map<*, *>) {
```
dariuszkuc: whats the point of extra `is` check? e.g. why not just keep it simply `payload as? Map<String, String> ?: emptyMap()` as it used to be (and in that case do we really need one liner method?)?

graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/spring/routes/SubscriptionRoutesConfigurationIT.kt
```diff
@@ -130,6 +130,8 @@ class SubscriptionRoutesConfigurationIT(
             .verify()
     }
 
+    private fun convertValueToString(message: SubscriptionOperationMessage) = objectMapper.writeValueAsString(message)
```
dariuszkuc: unsure how useful this method is -> it is just a test class anyway....

smyrick: Will delete

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionProtocolHandler.kt
smyrick: Yea, will delete. This was added so I could verify what the warning in intellij was complaining about.

PR #615, opened 2020-02-26
[generator] Fix issue generating nested additional types

smyrick:
### :pencil: Description
We would encounter a bug if generating the additional types produced more additional types. So we now call the function recursively and clear it out before generating more.

### :link: Related Issues
N\A

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -60,7 +60,7 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
         builder.query(generateQueries(this, queries))
         builder.mutation(generateMutations(this, mutations))
         builder.subscription(generateSubscriptions(this, subscriptions))
-        builder.additionalTypes(generateAdditionalTypes(additionalTypes))
+        builder.additionalTypes(generateAdditionalTypes())
         builder.additionalDirectives(directives.values.toSet())
         builder.codeRegistry(codeRegistry.build())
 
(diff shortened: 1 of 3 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -86,6 +86,18 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
      * Generate the GraphQL type for all the `additionalTypes`. They are generated as non-inputs and not as IDs.
      * If you need to provide more custom additional types that were not picked up from reflection of the schema objects,
      * you can modify the set of `additionalTypes` before you call this method.
+     *
+     * This function is recursive because generating the additionalTypes make create more additionalTypes.
```
dariuszkuc:
```suggestion
     * This function is recursive because while generating the additionalTypes it is possible to create additional types that need to be processed.
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -86,6 +86,18 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
      */
-    protected open fun generateAdditionalTypes(additionalTypes: Set<KType>): Set<GraphQLType> = additionalTypes.map { generateGraphQLType(this, it) }.toSet()
+    protected open fun generateAdditionalTypes(): Set<GraphQLType> {
+        val currentAdditionalTypes = this.additionalTypes.toSet()
```
dariuszkuc:
```suggestion
        val currentlyProcessedTypes = this.additionalTypes.toSet()
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -86,6 +86,18 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
+        this.additionalTypes.clear()
+        val graphqlTypes = currentAdditionalTypes.map { generateGraphQLType(this, it) }.toSet()
+
+        return if (this.additionalTypes.isNotEmpty()) {
+            generateAdditionalTypes().plus(graphqlTypes)
```
dariuszkuc: i think this should be reversed? i.e.

```
graphqlTypes.plus(generateAdditionalTypes())
```

smyrick: It shouldn't really matter as it would just at the types to an empty set but I can update

PR #616, opened 2020-02-26
Revert "Update publish docs to use a community action"

dariuszkuc:
Reverts ExpediaGroup/graphql-kotlin#614


smyrick:
@dariuszkuc The issue was that I setup the directory incorrect for where the build is. It actually still worked fine

I will fix forward

https://github.com/ExpediaGroup/graphql-kotlin/runs/469008143?check_suite_focus=true#step:6:123

PR #619, opened 2020-02-26
move generation of additional types to an iterative approach

tapaderster:
### :pencil: Description
Changed the current recursive approach of adding the additional types to iterative one.

This approach saves:
-  a bunch of `toSet()` calls which internally means creation of new LinkedHashSet
- `.plus()` calls which also creates a new LinkedHashSet and does an `addAll()` internally.

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -90,14 +90,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
      * This function is recursive because while generating the additionalTypes it is possible to create additional types that need to be processed.
      */
     protected open fun generateAdditionalTypes(): Set<GraphQLType> {
-        val currentlyProcessedTypes = this.additionalTypes.toSet()
-        this.additionalTypes.clear()
-        val graphqlTypes = currentlyProcessedTypes.map { generateGraphQLType(this, it) }.toSet()
+        val currentlyProcessedTypes = mutableSetOf<KType>()
+        val graphqlTypes = mutableSetOf<GraphQLType>()
+        do {
+            currentlyProcessedTypes.addAll(this.additionalTypes)
+            this.additionalTypes.clear()
+            graphqlTypes.addAll(currentlyProcessedTypes.map { generateGraphQLType(this, it) })
+            currentlyProcessedTypes.clear()
+        } while (this.additionalTypes.isNotEmpty())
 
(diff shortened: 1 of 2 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -90,14 +90,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
      * This function is recursive because while generating the additionalTypes it is possible to create additional types that need to be processed.
      */
     protected open fun generateAdditionalTypes(): Set<GraphQLType> {
-        val currentlyProcessedTypes = this.additionalTypes.toSet()
-        this.additionalTypes.clear()
-        val graphqlTypes = currentlyProcessedTypes.map { generateGraphQLType(this, it) }.toSet()
+        val currentlyProcessedTypes = mutableSetOf<KType>()
+        val graphqlTypes = mutableSetOf<GraphQLType>()
+        do {
```
dariuszkuc: i think you want to use `while` loop instead of `do` as additional types could be empty

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -90,14 +90,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
+            currentlyProcessedTypes.addAll(this.additionalTypes)
+            this.additionalTypes.clear()
+            graphqlTypes.addAll(currentlyProcessedTypes.map { generateGraphQLType(this, it) })
+            currentlyProcessedTypes.clear()
```
dariuszkuc: is constantly clearing the same object better than recreating the hash map?, e.g.

```
        val graphqlTypes = mutableSetOf<GraphQLType>()
        while (this.additionalTypes.isNotEmpty()) {
            val currentlyProcessedTypes = HashSet(this.additionalTypes)
            this.additionalTypes.clear()
            graphqlTypes.addAll(currentlyProcessedTypes.map { generateGraphQLType(this, it) })
        }
        return graphqlTypes.toSet()
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
smyrick: Whoops, merged to fast. Yes it would execute the first time, it would iterate over an empty set so not doing to much, but a while could be clearer

PR #622, opened 2020-02-27
Add additionalTypes to generateSchema

de55:
### :pencil: Description
Additional types cannot be edited, but comments suggest that is the appropriate way to add additional types before generateAdditionalTypes() is called:

```
    /**
     * Generate the GraphQL type for all the `additionalTypes`. They are generated as non-inputs and not as IDs.
     * If you need to provide more custom additional types that were not picked up from reflection of the schema objects,
     * you can modify the set of `additionalTypes` before you call this method.
     *
     */
```

### :link: Related Issues
#615

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -42,7 +42,6 @@ import kotlin.reflect.full.createType
 open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
 
     val additionalTypes = mutableSetOf<KType>()
-    
     internal val classScanner = ClassScanner(config.supportedPackages)
     internal val cache = TypesCache(config.supportedPackages)
     internal val codeRegistry = GraphQLCodeRegistry.newCodeRegistry()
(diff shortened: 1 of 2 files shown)


smyrick:
@de55 I was just looking at this last night and realized our mistake. Thank you for all being on top of changes as an active community member.

How about instead of changing the scope of `additionalTypes` we add a new optional argument to `generateSchema` that will allow you to add these.

```kotlin

open fun generateSchema(
        queries: List<TopLevelObject>,
        mutations: List<TopLevelObject> = emptyList(),
        subscriptions: List<TopLevelObject> = emptyList(),
        additionalTypes: Set<KType> = emptySet()
): GraphQLSchema {

    this.additionalTypes.addAll(additionalTypes)
    val builder = GraphQLSchema.newSchema()
    // ... rest of code
}
```

Then we can make `generateAdditionalTypes` just `protected` and not `protected open`.

smyrick:
If you update the scope of `generateAdditionalTypes` can you also change the javadocs to the following:

```kotlin
    /**
     * Generate the GraphQL type for all the `additionalTypes`. They are generated as non-inputs and not as IDs.
     * If you need to provide more custom additional types that were not picked up from reflection of the schema objects,
     * you can provide more types to be added through [generateSchema].
     *
     * This function loops because while generating the additionalTypes it is possible to create more additional types that need to be processed.
     */
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -53,17 +53,19 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
...
+        this.additionalTypes.addAll(additionalTypes)
         val builder = GraphQLSchema.newSchema()
         builder.query(generateQueries(this, queries))
         builder.mutation(generateMutations(this, mutations))
         builder.subscription(generateSubscriptions(this, subscriptions))
         builder.additionalTypes(generateAdditionalTypes())
         builder.additionalDirectives(directives.values.toSet())
         builder.codeRegistry(codeRegistry.build())
-
+
```
smyrick: You can delete this double empty blank line

```suggestion
```

smyrick:
This is a major change since we removed the open modified on `generateAdditionalTypes` but that was only added in on the 2.0 branch so we are back to being backwards compatible with 1.x for this method call at least

PR #623, opened 2020-02-27
BREAKING CHANGE: Make the KotlinDataFetcherFactoryProvider nullable

smyrick:
### :pencil: Description
We made the `FunctionDataFetcher` return nullable in this change https://github.com/ExpediaGroup/graphql-kotlin/pull/613. That is a good change but I only realized after integrating into our internal SDK that this also impacts the `KotlinDataFetcherFactoryProvider`. 

`KotlinDataFetcherFactoryProvider` returns a `DataFetcherFactory`, and in graphql-java those are supposed to implement the same types that the `DataFetcher` returns. So this change fixed that to match the same types in graphql-java.

This is a breaking change as the interface signature has now been changed.

### :link: Related Issues
Data Fetcher returns nullable here: https://github.com/ExpediaGroup/graphql-kotlin/pull/613


tapaderster:
lgtm

PR #624, opened 2020-02-27
Increase Gradle wrapper memory

smyrick:
### :pencil: Description
Increase the gradle jvm max memory as we are seeing intermittent OutOfMemory errors on GitHub Actions

See: https://github.com/gradle/gradle/issues/8139

### :link: Related Issues
N\A

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2.1-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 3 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2.1-bin.zip
```
dariuszkuc: do we need to update the wrapper binary as well?

smyrick: I ran ./gradlew wrapper and that did not change the jar or the executables

smyrick:
I don't think there are configurations for changing the runners

It looks like they all have the same resources

https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

smyrick:
And it still failed

```
DirectiveTests > SchemaGenerator marks deprecated queries and documents replacement() FAILED
    io.github.classgraph.ClassGraphException at DirectiveTests.kt:49
        Caused by: java.lang.OutOfMemoryError

332 tests completed, 24 failed, 2 skipped
```

I don't know if we want to try to increase again. It doesn't look like it is one test in particular running high usage as all the generator tests are just failing

https://github.com/ExpediaGroup/graphql-kotlin/runs/476345306?check_suite_focus=true#step:6:212

Issue #625, opened 2020-02-28
Some dependencies contain CVEs


sewilguler:
**Library Version**
2.0.0-RC9

**Describe the bug**
Hi,
I found io.netty 4.1.45 that is used in your project has some security problem. 
 
See below for more details:
<img width="1435" alt="Screenshot 2020-02-28 at 10 36 41" src="https://user-images.githubusercontent.com/7179520/75537526-9348f080-5a17-11ea-9f2c-47b7fff45ffd.png">

Overview
org.wso2.transport.http:org.wso2.transport.http.netty is a HTTP protocol handling implementations for C5 based products.

Affected versions of this package are vulnerable to HTTP Response Splitting due to HTTP Header validation being disabled.

Remediation
Upgrade org.wso2.transport.http:org.wso2.transport.http.netty to version 6.3.1 or higher.

References:
CONFIRM - https://snyk.io/vuln/SNYK-JAVA-ORGWSO2TRANSPORTHTTP-548944

sewilguler:
The dependency causing the problem added by org.springframework.boot:spring-boot-starter-reactor-netty:2.2.4.RELEASE which comes with org.springframework.boot:spring-boot-starter-webflux:2.2.4.RELEASE

smyrick:
It looks like Spring 2.2.5 was released with this fix just yesterday: https://spring.io/blog/2020/02/27/spring-boot-2-2-5-released

We can upgrade our dependency then

smyrick:
It looks like you were using this product to find security reports: https://snyk.io/product/

There are lots of security tools out there but we have not added one in graphql-kotlin yet other than the Github security alerts. This would be a good thing to add to our library

sewilguler:
Hi @smyrick,

Thank you so much for the help. I have a scheduled gradle task on my project running org.owasp.dependencycheck: 5.3.0 every 6 hours. So it is alerting whenever there is a new security issue. Thanks again for the help. Have a nice weekend.

sewilguler:
Hi @smyrick,

Problem actually still exists. And security fixes they made is not CVE-2019-10797. This link might help: netty/netty#10059.

smyrick:
Ah, ok. Thank you for checking. I was too quick to assume that a CVE fix in the release for Reactor Netty was the same issue.

[...]

PR #627, opened 2020-02-28
Upgrade springboot to fix security issue

smyrick:
### :pencil: Description
Spring Boot 2.2.5 was released which includes a fix for a security issue in netty. See: https://spring.io/blog/2020/02/27/spring-boot-2-2-5-released

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/625

### DIFF
--- a/graphql-kotlin-schema-generator/build.gradle.kts
+++ b/graphql-kotlin-schema-generator/build.gradle.kts
@@ -5,7 +5,7 @@ val graphQLJavaVersion: String by project
 val jacksonVersion: String by project
 val kotlinVersion: String by project
 val kotlinCoroutinesVersion: String by project
-val rxjavaVersion = "3.0.0"
+val rxjavaVersion: String by project
 
 dependencies {
     api("com.graphql-java:graphql-java:$graphQLJavaVersion")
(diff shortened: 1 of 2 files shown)


graphql-kotlin-schema-generator/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ val graphQLJavaVersion: String by project
 val jacksonVersion: String by project
 val kotlinVersion: String by project
 val kotlinCoroutinesVersion: String by project
-val rxjavaVersion = "3.0.0"
+val rxjavaVersion: String by project
```
dariuszkuc: maybe we should update our tests to create custom Reactor Monad and drop RxJava altogether?

dariuszkuc: *just a side note for future

smyrick: If we have it as an official example in our documentation I like having it included in our test dependencies. 

But as you mentioned we could just edit our docs and examples to just use reactor/spring wrappers and comment that they can work for any custom wrappers you want to include

PR #629, opened 2020-02-28
Add native subscription support for coroutine Flows

### LINKED ISSUE #358: native subscription support for Kotlin Flow
As per [graphql-java](https://www.graphql-java.com/documentation/v13/subscriptions/) currently we only support subscription functions that return some `Publisher` (e.g. Reactor `Flux`). Since we support coroutines for queries/mutations we should also support `Flow` for subscriptions.

Support Kotlin `Flow` type for subscriptions.

N/A

**Additional context**
N/A

josephlbarnett:
Implement a SubscriptionExecutionStrategy that allows
for `Flow`s and `Publisher`s to be returned from
graphql schema elements, and can be processed as a
`Flow` by subscription consumers.  Relax restrictions
that look for `Publisher`s to also allow `Flow`s.

Fixes #358

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FlowSubscriptionExecutionStrategy.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FlowSubscriptionExecutionStrategy.kt
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2020 Expedia, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.execution
+
+import graphql.ExecutionResult
+import graphql.ExecutionResultImpl
+import graphql.execution.DataFetcherExceptionHandler
+import graphql.execution.ExecutionContext
+import graphql.execution.ExecutionStrategy
+import graphql.execution.ExecutionStrategyParameters
+import graphql.execution.FetchedValue
+import graphql.execution.SimpleDataFetcherExceptionHandler
+import graphql.execution.SubscriptionExecutionStrategy
+import graphql.execution.reactive.CompletionStageMappingPublisher
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.future.await
+import kotlinx.coroutines.reactive.asFlow
+import org.reactivestreams.Publisher
+import java.util.Collections
+import java.util.concurrent.CompletableFuture
+
+/**
+ * [SubscriptionExecutionStrategy] replacement that returns an [ExecutionResult]
+ * that is a [Flow] instead of a [Publisher], and allows schema subscription functions
+ * to return either a [Flow] or a [Publisher].
+ *
+ * Note this implementation is mostly a java->kotlin copy of [SubscriptionExecutionStrategy],
+ * with [CompletionStageMappingPublisher] replaced by a [Flow] mapping, and [Flow] allowed
+ * as an additional return type.  Any [Publisher]s returned will be converted to [Flow]s,
+ * which may lose meaningful context information, so users are encouraged to create and
+ * consume [Flow]s directly (see https://github.com/Kotlin/kotlinx.coroutines/issues/1825
+ * https://github.com/Kotlin/kotlinx.coroutines/issues/1860 for some examples of lost context)
+ */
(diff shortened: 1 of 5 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FlowSubscriptionExecutionStrategy.kt
```diff
@@ -0,0 +1,153 @@
...
+        val sourceEventStream = createSourceEventStream(executionContext, parameters)
+
+        //
+        // when the upstream source event stream completes, subscribe to it and wire in our adapter
+        return sourceEventStream.thenApply { sourceFlow ->
+            if (sourceFlow == null) {
+                ExecutionResultImpl(null, executionContext.errors)
+            } else {
+                val returnFlow = sourceFlow.map {
+                    executeSubscriptionEvent(executionContext, parameters, it).await()
```
josephlbarnett: not entirely sure this is exactly equivalent to SubscriptionExecutionStrategy' use of `CompletionStageMappingPublisher`, but it appears to work -- possibly there's a better / more asynchronous way to do this though?

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FlowSubscriptionExecutionStrategy.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.reactivestreams.Publisher
+import java.util.Collections
+import java.util.concurrent.CompletableFuture
+
+/**
+ * [SubscriptionExecutionStrategy] replacement that returns an [ExecutionResult]
+ * that is a [Flow] instead of a [Publisher], and allows schema subscription functions
+ * to return either a [Flow] or a [Publisher].
+ */
+class FlowSubscriptionExecutionStrategy(dfe: DataFetcherExceptionHandler) : ExecutionStrategy(dfe) {
```
dariuszkuc: I think you can make it much simpler -> extend `CompletionStageMappingPublisher` and then you only need to create `createSourceEventStream` that returns `CompletableFuture<Publisher<Object>>` (i.e. convert from `Flow` to `Publisher`). Unfortunately `createSourceEventStream` is `private` so you would also need to override the `execute` method but it would basically be copy and paste.

josephlbarnett: Yeah, this class is mostly copy paste already.   I think what you're suggesting would work for allowing schemas to specify Flow and then consuming a Publisher from the ExecutionResult's data.  I think an execution strategy that did that could make sense for some cases.  In our case we want to consume a Flow out of the ExecutionResult's data and avoid the Publisher conversion due to https://github.com/Kotlin/kotlinx.coroutines/issues/1825

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FlowSubscriptionExecutionStrategy.kt
josephlbarnett: actually, `.await()` is a suspend function extension to the CompletableFuture, so this should be ok (had thought it was a blocking method when I wrote that comment)

josephlbarnett:
Also, if this ExecutionStrategy doesn't make sense for graphql-kotlin, would appreciate a way to implement it externally, either by the rest of the changes in this PR or some other/pluggable means.  

Right now I don't think the check in generateSubscription.kt is overridable, although I think the functionReturnTypes.kt diff can be implemented via schemaHooks.willResolveMonad()

smyrick:
This code changes in this PR looks good to me, I think if you wanted to provide a custom subscription execution strategy you can just provide your own `GraphQL` bean and that will be picked up https://github.com/ExpediaGroup/graphql-kotlin/blob/38a4f94cb92e9835873ed284377240a52211115b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/GraphQLAutoConfiguration.kt#L77-L78

As far as checking the types, we can add more hooks that would allow you to do this without merging a execution strategy for the entire library.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FlowSubscriptionExecutionStrategy.kt
smyrick: @dariuszkuc What do you think about this implementation then? Do we want to have this copy paste execution strategy in the library or just provide more hooks for other developers to customize the supported return types?

dariuszkuc: I'm leaning towards both - @josephlbarnett can you update `FlowSubscriptionExecutionStrategy` kdoc to indicate this is a copy of `SubscriptionExecutionStrategy` that adds support for `Flow`?

josephlbarnett: :+1: updated KDoc.

josephlbarnett:
Rebased off master to use the new subscription return type hooks, thanks those look good!

Issue #632, opened 2020-03-03
Ignored interface fields are still used for subtypes 


gscheibel:
**Library Version**
What version are you using?
Bug found in 1.x and 2.x

**Describe the bug**
When an interface has a field marked with `@GraphQLIgnore`, the field is still used when generating the subtypes.

**To Reproduce**
```
interface Service {
    val name: String

    @GraphQLIgnore
    val shouldNotBeInTheSchema: Boolean
}

data class WebService(override val name: String, override val shouldNotBeInTheSchema: Boolean = false) : Service
```

**Expected behavior**
The expectation is for the schema to contain an interface `Service` with a single `name` field and an object type `WebService` inheriting `name`.
`shouldNotBeInTheSchema` should neither appear in the interface definition nor in the object one.

smyrick:
Resolved in above issues

PR #634, opened 2020-03-03
Polymorphic ignore 2.x

### LINKED ISSUE #631: Polymorphic subtypes are not ignored with @GraphQLIgnore
**Library Version**
What version are you using?
Bug found in 1.x and 2.x branches

**Describe the bug**
When an interface subtype is marked with `@GraphQLIgnore` the generator still creates a graphql type from it and bypasses the annotation.

**To Reproduce**
```kotlin
interface Service {
    val name: String
}

data class WebService(override val name: String, override val shouldNotBeInTheSchema: Boolean = false) : Service

@GraphQLIgnore
data class MicroService(override val name: String, override val shouldNotBeInTheSchema: Boolean = true) : Service
```

**Expected behavior**
The expectation is for the schema to only contain an interface `Service` and an object type `WebService`.
`MicroService` should be ignored.

gscheibel:
### :pencil: Description

Fixes transitive properties of @GraphQLIgnore when applied on polymorphic types.
Apply the same changes as #633 for the 2.x / master branch

### :link: Related Issues
Fixes #631 and #632


smyrick:
@gscheibel Can you update this PR to solve the code coverage failures

Issue #635, opened 2020-03-08
List's of custom scalar types no longer work as arguments with Jackson 2.10


alexsparrow:
**Library Version**
2.0.0-RC10

**Describe the bug**
As noted in https://github.com/ExpediaGroup/graphql-kotlin/issues/518, the behaviour of `convertValue` appears to have changed in Jackson 2.10 so that it no longer short-circuits when the input value is already of a suitable type. 

For my use case, this causes an issue  when a function on my `TopLevelObject` takes a list of non-primitive types e.g. 
```
fun foo(ids: List<UUID>)
```
I have registered `UUID` as a scalar type using `SchemaGeneratorHooks` and defined a suitable `Coercing` for it.

With Jackson 2.9.X, this works as expected.

With Jackson 2.10.X, when I try to query this function, I get an exception like:
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.util.UUID (java.lang.String and java.util.UUID are in module java.base of loader 'bootstrap')
```

I believe this is due to the usage of `convertValue` in `FunctionDataFetcher.convertParameterValue`. In Jackson 2.10.X the list will always go through a serialization/deserialization step, losing the type information. 

**To Reproduce**
See above.

**Expected behavior**
Expect the `List<UUID>` to be forwarded as-is to my function.

Thanks for any help :-)

smyrick:
@alexsparrow Can you verify that you can run #636 locally? I checked out from the master branch I am not seeing any errors and we are running on jackson 2.10.2

alexsparrow:
Hey @smyrick thanks for the response and the tests! I left a couple of comments on your PR.

I don't think I'm running groovy (and definitely for the code that's causing the error).

smyrick:
@alexsparrow Ah, yes. I see now. It seems that we lose the information for converting custom scalars. We only added support for lists as input types in this PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/546

If you change the type to `Array<UUID>` then it should work. I will investigate why Lists are not working.

smyrick:
@alexsparrow I found the issue. It was infact in `FunctionDataFetcher`. I updated my PR to include the bug fix

alexsparrow:
Awesome thanks a lot for looking into this. The bug fix looks good and the workaround with `Array` is handy in the meantime.

smyrick:
Fixed in https://github.com/ExpediaGroup/graphql-kotlin/pull/636

PR #636, opened 2020-03-09
Custom scalar input

smyrick:
### :pencil: Description
Converting a list input type for custom scalars was losing the type information. We need to use a specific Jackson conversion method to properly convert the List interface

Screen Shot 2020-03-09 at 1 28 48 PM

### :link: Related Issues
#635

### DIFF
--- a/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/CustomScalarExecutionTest.kt
+++ b/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/CustomScalarExecutionTest.kt
@@ -27,6 +27,7 @@ import org.junit.jupiter.api.Test
 import java.util.UUID
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
+import kotlin.test.assertEquals
 import kotlin.test.assertFails
 import kotlin.test.assertNotNull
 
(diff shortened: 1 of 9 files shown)


graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/CustomScalarExecutionTest.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        assertNotNull(data?.get("uuidListInput"))
+    }
+
+    class QueryObject {
+        fun uuidOutput(): UUID = UUID.randomUUID()
+
+        fun uuidInput(uuid: UUID) = "You sent $uuid"
+
+        fun uuidListInput(uuids: List<UUID>) = "You sent ${uuids.size} items"
```
alexsparrow:
```suggestion
        fun uuidListInput(uuids: List<UUID>) = "You sent ${uuids.size} items with class ${uuids[0]::class.java}"
```

graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/CustomScalarExecutionTest.kt
```diff
@@ -0,0 +1,88 @@
...
+            graphQL.execute("{ uuidInput(uuid: \"foo\") }")
+        }
+    }
+
+    @Test
+    fun `a custom scalar can be used as input in a list`() {
+        val result = graphQL.execute("{ uuidListInput(uuids: [\"435f5808-936b-40ab-89b6-ec4e8cd1ba36\"]) }")
+        val data: Map<String, String>? = result.getData()
+
+        assertNotNull(data?.get("uuidListInput"))
```
alexsparrow:
```suggestion
        assertEquals("You sent 1 items with class UUID", data?.get("uuidListInput"))
```

alexsparrow: Hey @smyrick hopefully you can see what I'm getting at here :-) When I run these tests, this assertion fails because the inner type in the List is actually a `String`. Do you get the same?

PR #640, opened 2020-03-18
Update dependencies and gradle config

smyrick:
### :pencil: Description
Include certain dependencies as implementation instead of api so that they are not exposed by default. If you want to use coroutine in your GraphQL API you will need to also include the library


graphql-kotlin-spring-server/build.gradle.kts
```diff
@@ -14,8 +14,8 @@ dependencies {
     api(project(path = ":graphql-kotlin-federation"))
     api("org.springframework.boot:spring-boot-starter-webflux:$springBootVersion")
     kapt("org.springframework.boot:spring-boot-configuration-processor:$springBootVersion")
-    api("io.projectreactor.kotlin:reactor-kotlin-extensions:$reactorExtensionsVersion")
-    api("org.jetbrains.kotlinx:kotlinx-coroutines-reactor:$kotlinCoroutinesVersion")
+    implementation("io.projectreactor.kotlin:reactor-kotlin-extensions:$reactorExtensionsVersion")
```
dariuszkuc: I think that `spring-server` as SpringBoot autoconfiguration library should pull in all the required dependencies including both Reactor and Coroutines.

smyrick: We can exposes coroutines as an `api` dependency but I don't think we need to expose the reactor and kotlin extensions, right?

You don't need to use those to use `graphql-kotlin-spring-server`

dariuszkuc: i believe SpringBoot pulls in reactor so adding api dependency on coroutines should be fine

graphql-kotlin-federation/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 description = "Federated GraphQL schema generator"
 
 val junitVersion: String by project
+val kotlinCoroutinesVersion: String by project
```
dariuszkuc: this is not used

graphql-kotlin-spring-server/build.gradle.kts
smyrick: We do use these in the example apps, so we could make them api or do we want to update the example apps to pull in the extensions as well?

dariuszkuc: Let's keep them as `api` deps for now.

PR #641, opened 2020-03-18
Reuse schema generator for tests

smyrick:
### :pencil: Description
In an effort to save memory and build time, I made a common testGenerator that just uses the basic SchemaGeneratorConfig we already we using. Previously, every call to toSchema would create a new SchemaGenerator class which creates a new type cache and class scanner. Maybe this will help with the Github actions out of memory issues?

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -67,7 +67,7 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
         builder.codeRegistry(codeRegistry.build())
         val schema = config.hooks.willBuildSchema(builder).build()
 
-        classScanner.close()
+        cleanUpResources()
 
         return schema
     }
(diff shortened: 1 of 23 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -67,7 +67,7 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) {
         builder.codeRegistry(codeRegistry.build())
         val schema = config.hooks.willBuildSchema(builder).build()
 
-        classScanner.close()
```
smyrick: This will be closed automatically when the schema generator goes away

dariuszkuc: i'd still explicitly close it just in case

dariuszkuc: since `ScanResult` implements `Closeable` we could use Kotlin `use` (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html)

smyrick:
I have no idea why these checks are not failing for me locally. I keep running a build on java 11 but it is always passing for me

smyrick:
Closing in favor of https://github.com/ExpediaGroup/graphql-kotlin/pull/643

Issue #642, opened 2020-03-19
Sealed classes not treated as abstract


userkci:
**Library Version**
What version are you using?
1.4.2

**Describe the bug**
Sealed classes only generate a Type capturing their own class.

**To Reproduce**
Given the following class hierarchy:

```
sealed class Pet(val name: String) {
  data class Dog(name: String, val goodBoysReceived: Int) : Pet(name)
  data class Cat(name: String, val livesRemaining: Int) : Pet(name)
}
```

The resulting schema generated is:

```
type Pet {
  name: String!
}
```

**Expected behavior**
The expectation is that the sealed class defines an Interface/Union (as appropriate) and each subclass generates a type, such as below:

```
interface Pet { 
  name: String!
}

type Dog implements Pet {
  name: String!
  goodBoysReceived: Int!
}

type Cat implements Pet {
  name: String!
  livesRemaining: Int!
}
```

It looks like this is due to `KClass.isAbstract` returning false for a sealed class and therefore ` KClass<*>.isInterface()` does too. Digging into the Kotlin source this is simply a check if the modality of the class is `ABSTRACT`, but the modality is an enumeration of `OPEN`, `FINAL`, `ABSTRACT`, and `SEALED`. So I'd suspect that adding `KClass.isSealed` to the interface check would suffice.

smyrick:
@userkci Thank you for reporting this issue! It appears that we did not consider this when implementing the library, but it seems possible to support.

We will merge this into the master branch because I consider this a new feature we did not implement yet, so that means it will be available in the 2.x.x branch.

If you would like to provide a pull request to get this in sooner, please feel free to assign yourself to the issue, otherwise we won't wait to get this in the 2.0.0 release which means it may come out as a 2.1 release.

Issue #589, opened 2020-02-01
[generator] allow interfaces to implement other interfaces


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
GraphQL spec now supports nested interfaces. `graphql-kotlin-schema-generator` should support this as well.

**Describe the solution you'd like**
`graphql-kotlin-schema-generator` should be able to generate schemas with nested interfaces 

**Describe alternatives you've considered**
N/A

**Additional context**
GraphQL spec PR: https://github.com/graphql/graphql-spec/pull/373

smyrick:
We may just want to wait until this is added in graphql-java first: https://github.com/graphql-java/graphql-java/issues/1764

PR #645, opened 2020-03-20
Support sealed classes for interfaces

smyrick:
### :pencil: Description
Allows sealed classes to be used as GraphQL Interfaces. This means that you can restrict the possible implementations of the interfaces since subclasses can only be declared with the root class

https://kotlinlang.org/docs/reference/sealed-classes.html

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/642

### DIFF
--- a/docs/writing-schemas/interfaces.md
+++ b/docs/writing-schemas/interfaces.md
@@ -92,7 +92,7 @@ sealed class Pet(val name: String) {
 ```
 
 ### Known Issues
-> NOTE: Due to a feature added in 1.0.0, we no longer support multiple levels of interfaces in a schema because the GraphQL spec does not support this feature.  [See 419](https://github.com/ExpediaGroup/graphql-kotlin/issues/419). If you do have multiple interfaces you will have to either combine them into a single interface or ignore all the parent interfaces.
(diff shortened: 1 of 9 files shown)


docs/writing-schemas/interfaces.md
```diff
@@ -1,117 +1,131 @@
...
+class Square(sideLength: Double) : Shape(sideLength * sideLength)
+
+sealed class Pet(val name: String) {
+    class Dog(name: String, val goodBoysReceived: Int) : Pet(name)
+    class Cat(name: String, val livesRemaining: Int) : Pet(name)
+}
+```
+
+### Known Issues
+> NOTE: Due to a feature added in 1.0.0, we no longer support multiple levels of interfaces in a schema because the GraphQL spec does not support this feature.  [See 419](https://github.com/ExpediaGroup/graphql-kotlin/issues/419). If you do have multiple interfaces you will have to either combine them into a single interface or ignore all the parent interfaces.
```
dariuszkuc: new spec supports it -> https://github.com/ExpediaGroup/graphql-kotlin/issues/589, do we need `graphql-java` to update first?

smyrick: Yes, I can update the wording. The spec now supports it but we do not yet.

PR #646, opened 2020-03-22
Reduce number of class scanners created in tests

smyrick:
### :pencil: Description
In an attempt to fix the out of memory errors, I simplified some of our unit tests to create few ClassScanners and SchemaGenerators, and close all usages of the ones we still have left

### DIFF
--- a/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/exception/SimpleKotlinGraphQLError.kt
+++ b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/exception/SimpleKotlinGraphQLError.kt
@@ -32,13 +32,12 @@ open class SimpleKotlinGraphQLError(
 ) : GraphQLError {
     override fun getErrorType(): ErrorClassification = errorType
 
-    override fun getExtensions(): Map<String, Any> {
-        val newExtensions = mutableMapOf<String, Any>()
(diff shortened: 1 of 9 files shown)


smyrick:
Build passes on this one on first try so that is a good sign

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/exception/SimpleKotlinGraphQLError.kt
```diff
@@ -32,13 +32,12 @@ open class SimpleKotlinGraphQLError(
 ) : GraphQLError {
     override fun getErrorType(): ErrorClassification = errorType
 
-    override fun getExtensions(): Map<String, Any> {
-        val newExtensions = mutableMapOf<String, Any>()
+    override fun getExtensions(): Map<String, Any> =
         if (exception is GraphQLError && exception.extensions != null) {
-            newExtensions.putAll(exception.extensions)
+            exception.extensions
```
dariuszkuc: i believe the intent here was to ensure it is mutable map so we could add to it afterwards if needed

dariuszkuc: actually nvm this should be fine. if we need to add to the extensions then we can modify it to mutable map at that time

PR #648, opened 2020-03-28
Documentation: Add example using the Spark framework

rharriso:
* Documentation: Adding an example for the Spark framework demonstrates how to use
  `graphql-kotlin` outside of the context of a Spring Application. This example
  demonstrates the use of data loaders, nested data, and how exceptions are handled.

### :pencil: Description

I've added an example that uses the Spark framework (rather than spring). I had some trouble figuring out how to use `DataLoader`s without the context of a spring application, and though others might find this useful,

Additionally this example uses Maven for build and dependency management. This might be useful, or might be annoying to have two different build systems in the examples directory. If you'd like to change this to Gradel, that's fine, but I'll need some help.


examples/spark/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+> NOTE: in order to ensure you use the right version of Gradle we highly recommend that you use the provided wrapper scripts
+
+Start the server:
+
+* Open this directory in Intelij or another ide
+* Run `Application.kt` directly from your IDE
+
+Alternatively, you can compile the project with Maven and run the resulting jar
```
smyrick:
```suggestion
Alternatively, you can compile the project with Gradle and run the resulting jar
```

examples/spark/pom.xml
```diff
@@ -0,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
smyrick: Can we migrate this to Gradle so we still have just one build system for all the examples

smyrick:
@rharriso Thank you for your contribution! It is great to have more examples outside of spring as we also don't want to lock down other server implementations.

Can you also update the `/docs` website with a quick paragraph about the example. Maybe this page is a good place: https://expediagroup.github.io/graphql-kotlin/docs/examples

examples/spark/pom.xml
rharriso: Definitely understand. I'm not at all experienced with gradle, but I'll give it a shot later today.

rharriso:
@smyrick I having a bit of trouble converting to gradle. I'm getting the following error when I run `./gradlew build`

```
> Task :compileKotlin FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileKotlin'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find com.expediagroup:graphql-kotlin:.
     Required by:
         project :
   > Could not find com.sparkjava:spark-cor:.
     Required by:
         project :
   > Could not find org.slf4j:slf4j-log4j12:.
     Required by:
         project :

```

examples/spark/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+import java.util.Properties
+
+plugins {
+    kotlin("jvm") version "1.3.70"
```
smyrick: This plugin should be included in all sub-projects already. See the `examples/spring` build file

https://github.com/ExpediaGroup/graphql-kotlin/blob/2806cd80125dbba451150e42bbf700589649f923/examples/spring/build.gradle.kts#L3

examples/spark/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+application {
+    mainClassName = "com.expediagroup.graphql.examples.spark.Application"
+}
+
+dependencies {
+    implementation(kotlin("stdlib", "1.3.70"))
+    compile("com.expediagroup", "graphql-kotlin")
```
smyrick: This is not the library you need to include, the `examples` build script already includes all the depenencies for the sub-projects.

https://github.com/ExpediaGroup/graphql-kotlin/blob/2806cd80125dbba451150e42bbf700589649f923/examples/build.gradle.kts#L45

You need to add the sub-project here: https://github.com/ExpediaGroup/graphql-kotlin/blob/2806cd80125dbba451150e42bbf700589649f923/examples/settings.gradle.kts#L25

examples/spark/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+import java.util.Properties
+
+plugins {
+    kotlin("jvm") version "1.3.70"
+    application
+}
+
+allprojects {
```
dariuszkuc: no need for those repos as it should be defined in the parent example gradle build file

examples/spark/src/main/kotlin/com/expediagroup/graphql/examples/spark/Application.kt
```diff
@@ -0,0 +1,57 @@
+package com.expediagroup.graphql.examples.spark
+
+import org.apache.log4j.PropertyConfigurator
+import org.slf4j.LoggerFactory
+import spark.Spark.*
```
dariuszkuc: no wildcard imports please

examples/spark/build.gradle.kts
dariuszkuc: i think we'll have to change common dependencies to only include schema generator, for now you could exclude the spring server dependency and include schema gen explicitly

rharriso:
Hey guys, thanks for the help. `Application.kt` can now be run from Intellij. I don't have a gradle command to build and run it as an application

examples/spark/src/main/kotlin/com/expediagroup/graphql/examples/spark/Application.kt
rharriso: :+1: updated

examples/spark/build.gradle.kts
rharriso: Thanks, I wasn't catching the subprojects declaration.

examples/spark/build.gradle.kts
```diff
@@ -0,0 +1,7 @@
+description = "An Example GraphQL service servied by a Spark HTTP server"
+
```
dariuszkuc: if you add `application` plugin and configure main class you should be able to run the app using `run` task. See https://docs.gradle.org/current/userguide/application_plugin.html

rharriso: :+1: 
Added an application entry in the spark `build.gradle.kts` file.

EXPErdelfavero: There's a typo in line 1: "service servied by" should be "service *served* by"

rharriso: Updated

smyrick:
@rharriso Can you update all source code with the following license header: 

https://github.com/ExpediaGroup/graphql-kotlin/blob/master/CONTRIBUTING.md#add-license-information

rharriso:
> @rharriso Can you update all source code with the following license header:
> 
> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/CONTRIBUTING.md#add-license-information

Updated

examples/spark/src/main/kotlin/com/expediagroup/graphql/examples/spark/Application.kt
```diff
@@ -0,0 +1,76 @@
+/**
+ * Copyright $today.year Expedia, Inc
```
smyrick: All of these need to have `$today.year` replaced. If you are using an IDE like IntelliJ you can set this text as the template otherwise you will have to manually add this to all files

Screen Shot 2020-03-30 at 12 44 37 PM

rharriso: lol sorry

rharriso: Date set

rharriso:
:+1: Thanks for all your work on this project guys. It's been great to work with.

PR #650, opened 2020-04-03
Upload graphql-kotlin text logo

smyrick:
### :pencil: Description
This image is used as the social image for this repo, so I am uploading it here so we have a backup copy as well.

If we want to update the docs website to use this, that can be a separate PR


dariuszkuc:
Was wondering whether we should update our logo with IDEA typeface (https://www.jetbrains.com/lp/mono/)? This would make it reference Kotlin at least in the font.

Issue #651, opened 2020-04-07
Option for disabling introspection query


veckal:
Hi. I am using this library for production.
I want to disable introspection query for security. I tried to customize the Class `SchemaGenerator`, but it seems not working.
Is there any way to disable introspection quiery?

dariuszkuc:
Great question! Unfortunately currently we don't provide configuration option to easily disable it so you would have to do it manually by modifying the generated GraphQL schema.

See https://www.graphql-java.com/documentation/v14/fieldvisibility/ for details.

@smyrick I think that we probably should add configuration option to `spring-server` to support it

veckal:
@dariuszkuc Thank you for your answer.

smyrick:
I started a branch here which adds a new optional code registry bean: https://github.com/smyrick/graphql-kotlin/pull/new/code-registry


What do we think of this approach? This still needs doc updates and unit tests if we want to move forward with that.

dariuszkuc:
I think disabling introspection should be much simpler - in `spring-server` it should be controlled by a Boolean flag. As for the custom field visibility I think we should look into maybe some annotations as I don't think we should expose code registry directly.

smyrick:
@dariuszkuc The field visibility is handled through the code registry so if we want a config option as part of spring server, that means we have to open the registry to be modified in external packages.

PR #653, opened 2020-04-07
[generator] allow disabling introspection query

dariuszkuc:
### :pencil: Description

Schema generator should provide an option to easily disable introspection queries. Add new `introspectionEnabled` Boolean option to `SchemaGeneratorConfig` that is enabled by default.

spring-server was also updated to expose `graphql.introspection.enabled` property that will automatically configure the `SchemaGeneratorConfig` with the specified value.

### :link: Related Issues
Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/651


smyrick:
Needs unit test but overall it looks fine to me. I suppose we don't need to modify this in the generator at all as I was thinking

PR #654, opened 2020-04-07
[federation] Use apollo federation jvm

smyrick:
### :pencil: Description
Remove code that can be handled by the [apollo federation-jvm library](https://github.com/apollographql/federation-jvm)

This is just a patch changes since all code removed was internal and the schema produced is still the same.

Let's wait to merge until we can update the lib or verify why the SDL has some changes as compared to our implementation


smyrick:
It seems like there may be a few bugs with the schema but those can be fixed. I like having less code in our library and consolidating on a single apollo library anyway

dariuszkuc:
When creating federation module I explicitly decided not to use the `federation-jvm` as it was not providing much value to our use case. The intent was also to keep it separate so once we would migrate off `graphql-java` then changes to federation would be minimal.

dariuszkuc:
`We could remove even more code if we removed the FieldSet annotation in favor of just using a string for the fields value for all the other annotations but that would be a breaking change.`
That would break the spec. `FieldSet` is a custom scalar value that has special meaning for the federated gateway.

smyrick:
> > We could remove even more code if we removed the FieldSet annotation in favor of just using a string for the fields value for all the other annotations but that would be a breaking change.
> 
> That would break the spec. `FieldSet` is a custom scalar value that has special meaning for the federated gateway.

Yea, I see that now nvm: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#federation-schema-specification

dariuszkuc:
Will this actually work? We generate complete schema with all directive definitions and custom scalar information - AFAIK `federation-jvm` expects schema without those definitions and its transformation simply adds `_service` and `_entities` queries. Our logic explicitly strips out those definitions from the schema (i.e. those regexes) to transform the underlying SDL to something that federation can understand. Additionally we also added some logic there federated data fetchers -> I think it would break as well.

smyrick:
Closing for now as there are some unknown issues

Issue #655, opened 2020-04-09
@GraphQLIgnore not detected on BeanFactoryAware


SJean-Marc:
**Library Version**
1.4.2, 2.0.0

**Describe the bug**
If you use a subquery with scope prototype and inject a singleton into it, we have no guarantee, it is destroyed properly, I was expecting to extend BeanFactoryAware in order to guarantee the proper destruction of the bean, but even with @GraphQLIgnore, the graphql process try to map the beanFactory element in the subquery class.

**To Reproduce**

```
@Component
class SubQueries : Query, BeanFactoryAware {
    private lateinit var beanFactory: BeanFactory

    @GraphQLIgnore
    override fun setBeanFactory(beanFactory: BeanFactory) {
        this.beanFactory = beanFactory
    }

    fun main(): SubQuery = beanFactory.getBean()
}

@Component
@Scope("prototype")
class SubQuery , BeanFactoryAware {

    @GraphQLIgnore
    private lateinit var beanFactory: BeanFactory

    @GraphQLIgnore
    override fun setBeanFactory(beanFactory: BeanFactory) {
        this.beanFactory = beanFactory
    }
    @Autowired
    @GraphQLIgnore
    lateinit var batRepo: batRepository
    fun secondary() : Long{
     return batRepository.findById(1).id
   }
}
```

**Expected behavior**
Exception is thrown, it should accept the graphQLIgnore

smyrick:
@SJean-Marc Can you verify one more time that you are matching my example: https://github.com/ExpediaGroup/graphql-kotlin/pull/657


You don't need to include `@GraphQLIgnore` if you can make a function or property non-public.

Can you also try making `batRepository` private

smyrick:
Ahh, ok I see the issue. It is the fact that you are trying to do multiple levels of `BeanFactoryAware` and we are picking up the interface. So the issue is not with `@GraphQLIgnore` but you will have to update your hooks to exclude that interface with `isValidSuperclass`

```kotlin
override fun isValidSuperclass(kClass: KClass<*>): Boolean {
        return when {
            kClass.isSubclassOf(BeanFactoryAware::class) -> false
            else -> super.isValidSuperclass(kClass)
        }
    }
```

@dariuszkuc Do you think we should just add that the list of default excluded interfaces?

dariuszkuc:
[...]

PR #657, opened 2020-04-10
Add bean factory example and docs

smyrick:
### :pencil: Description
Add new docs explaining how the use Spring `BeanFactory` for nested queries

Screen Shot 2020-04-10 at 3 58 23 PM

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/655


smyrick:
New format

Screen Shot 2020-04-10 at 4 40 28 PM

docs/spring-server/spring-bean-factory.md
```diff
@@ -0,0 +1,49 @@
+---
+id: spring-bean-factory
+title: Spring BeanFactory
+---
+
+You may want to have schema inputs at different levels and if you are using Spring components that expect these inputs to be Autowired you will have to use [BeanFactory](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/BeanFactory.html).
+
+There is an example of how to set this up in [TopLevelBeanFactoryQuery.kt](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/TopLevelBeanFactoryQuery.kt) however you can also use the alternate approach of adding the [DataFetchingEnvironment](../writing-schemas/nested-queries.md) to access the arguments in parent nodes.
+
+Both methods have their pros and cons but you may find that the `DataFetchingEnvironment` approach means the Spring components are not directly tied together anymore and can be tested in isolation.
```
DelfCo, line 10: If you're going to introduce the idea that there are multiple pros and cons, you should really talk about them, and in terms that make it clear to a reader how they should weigh their own needs in light of either choice. (In other words, put it in terms of, "if you're doing X or Y, or Z is important to you, choose method A. If you're worried more about W, or you use U or T, then choose method B. If you like ice cream, then you must use method B.")

If you don't want to go into that, then just lose the "Both methods have their pros and cons, but" part, and leave the rest.

smyrick: I took your advice and just consolidated the docs into a single page for multiple ways you can use nested arguments

docs/writing-schemas/nested-arguments.md
```diff
@@ -0,0 +1,100 @@
+---
+id: nested-arguments
+title: Nested Arguments
```
dariuszkuc, line 3: i think this should be renamed to nested fields?

DelfCo: For ease of maintenance going forward, it's easiest if the id and file name both match the title. Makes it easier to find the right file when users report a bug based on the page title, which is the only part they can see. You'll have to update links to this page if you rename it. You'll have to update website/sidebars.json if you change the id.

In fact, you need to add this to website/sidebars.json in any case, it's not there yet.

Issue #668, opened 2020-04-23
Is it possible to return GraphQL ID from a query/mutation?


danbim:
I couldn't find a way to generate a Schema that has functions returning `ID` types. While this works for arguments it seems it does not for return types, here I could only chose `String`.

Taking a quick look into the generator code base [1] it seems in fact that return type is not in any way evaluated if it could be an `ID`.

Would be happy to help/contribute :)

[1] https://github.com/ExpediaGroup/graphql-kotlin/blob/81c00d55c2152ef6778ac4d1708143c2878598d1/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateFunction.kt#L53

dariuszkuc:
Great question! Currently we are relying on `@GraphQLID` annotation to determine whether given scalar field is of an `ID` type. While you can return complex objects with an `ID` field (e.g. https://expediagroup.github.io/graphql-kotlin/docs/writing-schemas/scalars#id) it appears that currently we do not support returning just raw `ID` by itself. With the current mechanism I am unsure whether we can easily support it. 

Feel free to open up a new bug report that describes this issue. Contributions are also always welcome!

dariuszkuc:
Actually it is a very simple fix. I'll create a PR in a bit.

danbim:
Thanks for the quick fix! That's exactly the solution I had in mind :)

PR #669, opened 2020-04-23
[generator] support returning GraphQL ID scalar from functions

dariuszkuc:
### :pencil: Description
Our current logic was not checking whether function return type was annotated with `@GraphQLID` annotation so it was not possible to return `ID` scalar type.

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/668

### DIFF
--- a/docs/writing-schemas/scalars.md
+++ b/docs/writing-schemas/scalars.md
@@ -36,7 +36,10 @@ data class Person(
   val name: String
 )
 
-fun createPerson(@GraphQLID id: String) = Person(id, "Jane Doe")
+fun locatePerson(@GraphQLID id: String) = Person(id, "Jane Doe")
+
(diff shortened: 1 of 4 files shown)


smyrick:
@dariuszkuc Does this work if we have a schema class like the following for output? 

```kotlin
class Widget(@GraphQLId val id: String, val value: String?)
```

dariuszkuc:
@smyrick yes, we already had example for it -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/ScalarQuery.kt#L38

smyrick:
Do we need to merge this back into the `1.x.x` branch? This seems like a major schema bug

PR #670, opened 2020-04-23
[generator] support returning GraphQL ID scalar from functions 1.x.x

dariuszkuc:
### :pencil: Description
Backport to 1.x.x

Our current logic was not checking whether function return type was annotated with @GraphQLID annotation so it was not possible to return ID scalar type.

### :link: Related Issues

* https://github.com/ExpediaGroup/graphql-kotlin/pull/669
* https://github.com/ExpediaGroup/graphql-kotlin/issues/668

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -3,7 +3,7 @@ name: Maven CI
 on:
   pull_request:
     branches:
-      - master
+      - 1.x.x
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


smyrick:
@dariuszkuc Can you also update the action to trigger on this branch https://github.com/ExpediaGroup/graphql-kotlin/blob/bcd047bb6670c40b78d46b1c8fd0ddbfdd5df7d7/.github/workflows/pr-check.yml#L6

PR #595, opened 2020-02-07
wip: gradle and maven plugins

dariuszkuc:
### :pencil: Description

WIP gradle plugin and client

PR #671, opened 2020-04-26
new gradle plugin with introspect schema task

dariuszkuc:
### :pencil: Description

This is an initial PR that introduces new `graphql-kotlin` Gradle plugin with schema introspection functionality. This new plugin task runs introspection query against target GraphQL endpoint and generates GraphQL schema based on the results.

PR size is very large due to the inclusion of raw introspection results that are used for testing of `plugin-core` and `gradle-plugin` projects. Those two files combined account for ~2800 lines of code.

NOTE: This is an attempt to split up https://github.com/ExpediaGroup/graphql-kotlin/pull/595 into more manageable pieces. Subsequent PRs will follow with additional functionality.

### :link: Related Issues
None

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,6 +9,7 @@ description = "Libraries for running a GraphQL server in Kotlin"
 extra["isReleaseVersion"] = !version.toString().endsWith("SNAPSHOT")
 
 plugins {
+    idea
     kotlin("jvm")
     id("org.jetbrains.dokka") apply false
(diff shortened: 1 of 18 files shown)


smyrick:
This doesn't break the existing library so it is not a major change

dariuszkuc:
> This doesn't break the existing library so it is not a major change

It is a brand new major functionality 🤷

smyrick:
It is a large major update yes, but in terms of [semver](https://semver.org/) it is not a `MAJOR` change

Issue #672, opened 2020-04-27
Custom GQL Context not created when endpoint called with trailing slash


imberda:
**Library Version**
1.4.2

**Describe the bug**

When the `graphql.endpoint` is configured as `/graphql` and a request is made to `/graphql/` (note the trailing slash), the request is handled by the route but the context is not generated nd added into the reactor subscriber context by `ContextWebFilter`.

This is because the route matches based on:
```
POST(config.endpoint)
```
Which is boils down to (something like):
```
PathPatternParser().parse("/graphql")
```
The `PathPatternParser` is automatically configured with `matchOptionalTrailingSeparator = true`, so a request such as `/graphql/` will be matched.

On the other hand the `ContextWebFilter` simply performs:
```
open fun isApplicable(path: String): Boolean = graphQLRoute.equals(path, ignoreCase = true)
```

**To Reproduce**
- Configure the graphql.endpoint to `/graphql`
- Make a request to `/graphql/`
- See that the context factory is not called to generate context

**Expected behavior**

The expected behaviour is that if the request is handled by the query handler, the context generator is called. 

There's likely different ways to solve this, but the most simple seems to be to enhance the `isApplicable` method in `ContextWebFilter`.

Perhaps some improved version of:
```
PathPatternParser().parse(enforceAbsolutePath(config.endpoint)).matches(PathContainer.parsePath(exchange.request.uri.path)
```

smyrick:
@imberda Thank you for raising the issue. This is a quick fix that I have provided in https://github.com/ExpediaGroup/graphql-kotlin/pull/673

If you can not wait for the library to be published the web filter is an open class and you can override the behaviour in `isApplicable` today.

imberda:
The fix looks good to me. Thanks for responding 👍

PR #673, opened 2020-04-27
Allow for trailing slashes on context filter

smyrick:
### :pencil: Description
The `ContextWebFilter` did a simple string compare to match the path. Instead we should use a proper `PathParser` to check the route and provide additional options like being case-insensitive and ignoring trailing slashes

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/672


graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ContextWebFilter.kt
```diff
@@ -37,6 +39,9 @@ const val GRAPHQL_CONTEXT_FILTER_ODER = 0
 open class ContextWebFilter<out T : GraphQLContext>(config: GraphQLConfigurationProperties, private val contextFactory: GraphQLContextFactory<T>) : WebFilter, Ordered {
     private val graphQLRoute = enforceAbsolutePath(config.endpoint)
     private val subscriptionsRoute = enforceAbsolutePath(config.subscriptions.endpoint)
+    private val parser = getPathPatternParser()
```
dariuszkuc, line 42: should we move all of this calculation to `init` block so we only expose the actual patterns?

smyrick: Done

PR #674, opened 2020-04-28
new Maven plugin with introspect schema mojo

dariuszkuc:
### :pencil: Description

This PR introduces new graphql-kotlin Maven plugin with schema introspection functionality (same functionality as provided by Gradle plugin). This new plugin MOJO runs introspection query against target GraphQL endpoint and generates GraphQL schema based on the results.

PR size is very large due to inclusion of raw introspection result used for integration testing as well as Maven wrapper scripts - together those files amount to over 2300 lines of code.

NOTE: This is a 2nd PR that attempts to split up #595 into more manageable pieces. Subsequent PRs will follow with additional functionality.

### :link: Related Issues
None


smyrick:
@dariuszkuc Do we need to somehow enable a local m2 repository directory for builds in actions?

dariuszkuc:
@smyrick we don't have to do anything but it I think we could potentially use .m2 caching there as well. Unfortunately caching action does not support multiple paths atm (scheduled for v2) so we need to run it twice.

smyrick:
@dariuszkuc I meant as a way to fix the build. It looks like it was a code issue though. For implementing the action cache I think we can just skip that for now

.github/workflows/pr-check.yml
```diff
@@ -30,6 +30,14 @@ jobs:
...
           restore-keys: |
             ${{ runner.os }}-${{ matrix.java }}-gradle-
+      # used by maven-plugin integration tests
+      - name: Set up Maven cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
smyrick, line 40: Should we stripe this by java version like we do with Gradle?

dariuszkuc: Yeah we should. 🤞 dokka 11 launches soon so we will be get rid of Java 8 build....

Issue #675, opened 2020-04-28
Exception thrown from ExecutionResult.toGraphQLResponse() when error has null locations


imberda:
**Library Version**
1.4.2 (but I think this also applies to 2.x)

**Describe the bug**

When a `GraphQLError` is return within the `ExecutionResult`,  extension function `ExecutionResult.toGraphQLResponse()` processes the  errors. Specifically this piece of code:
```
fun ExecutionResult.toGraphQLResponse(): GraphQLResponse {
    val filteredErrors = if (errors?.isNotEmpty() == true) {
        errors.map {
            when (it) {
                is SimpleKotlinGraphQLError -> it
                is Exception -> SimpleKotlinGraphQLError(exception = it, locations = it.locations, path = it.path, errorType = it.errorType)
                else -> it
            }
        }
    } else null
   ...
}
```
If one of the  `GraphQLError`s is an `Exception` the second branch of the `when` statement is evaluated. However, some errors return`null` from the `getLocations()` method. When the `null` is set into the  non-null `locations` property of  the`SimpleKotlinGraphQLError`, this results in an exception being thrown:
```
java.lang.IllegalStateException: it.locations must not be null
```

**To Reproduce**
- Define any simple schema
- Send a request which contains a field within an input type that does not exist within the schema
- This will cause a `InputMapDefinesTooManyFieldsException` (which returns `null` for `getLocations()`
- You will see the error contains this:
```
{
  "errors": [
    {
      "locations": [],
      "path": null,
      "errorType": "DataFetchingException",
      "extensions": {},
      "message": "Exception while fetching data () : it.locations must not be null"
    }
  ]
}
```

**Expected behavior**
The expected response would be something like this:
```
{
  "errors": [
    {
      "locations": [],
      "path": null,
      "errorType": "ValidationError",
      "message": "Exception while fetching data () : The variables input contains a field name 'foo' that is not defined for input object type 'ExampleInput' ",
      "extensions": {}
    }
  ]
}
```

As `GraphQLError` is a platform type we can't be sure if `getLocations()` will return a null. There seems to be two solutions, either replace `null` with an empty list (`it.locations.orEmpty()`) or change `SimpleKotlinGraphQLError.locations` to accept nulls.

dariuszkuc:
[...]

PR #677, opened 2020-04-29
[server] GraphQL Error response type should allow null locations

dariuszkuc:
### :pencil: Description
Based on the [Graphql spec](https://spec.graphql.org/June2018/#sec-Errors), error response should contain list of locations if error can be associated with particular point in the requested GraphQL document. We should not populate it if `locations` information is unavailable.

### :link: Related Issues
Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/675


smyrick:
This will require a major version release

PR #679, opened 2020-04-29
Incorrect Execution Command for base-app

acidbluebriggs:
The command to execute the base app was incorrect. It had two preceding colons, not one.


acidbluebriggs:
Well, actually, the gradlew command there is incorrect as well. I'm not a gradle person, so I may be doing something else wrong as the gradlew script is at the root of the repo's project. Thanks for the reply!

PR #680, opened 2020-04-29
Revert "[server] GraphQL Error response type should allow null locations"

smyrick:
Reverts ExpediaGroup/graphql-kotlin#677

I am going to revert to merge some other PRs first as there seems to be an issue with releasing different branches so I would like to get the bug fixes in the master branch first before we cut over to the 3.0.0 breaking changes

I will create a revert for the revert right after this is merged


dariuszkuc:
we could also just release `3.0` :P

smyrick:
Do we want to include this in the 2.x branch? If not then we can just move forward: https://github.com/ExpediaGroup/graphql-kotlin/pull/678

Issue #637, opened 2020-03-10
[federation] Referencing GraphQLList field using @RequiresDirective


aryadnasauchyts:
Hi there!
I have the following federated setup:
1) Basket app retrieving info about basket and its items (id, quantity, name).
2) Product app that extends basket app items with detailed product information. Of particular interest here is a product image.
I need to provide this image per each product which is pretty straight forward and might be done inside product app in some data fetcher. BUT I also need an image of last added item to basket to be available for basket itself. This part is a tricky one because:
-  First, I need to pick basket items
-  Then find the last added 
-  Then find a product related to this item
-  Get the image

Ideally, I would look into extending basket with image field in product service. Then using @RequiresDirective on this new image and using @ExternalDirective on items list in product service reference the items in basket service. This would allow having image retrieval logic (data fetcher + data loader) ecapsulated in one federated server.

In pseudocode:
`
@KeyDirective(fields = FieldSet("id"))
@ExtendsDirective
data class Basket(@ExternalDirective val id: Long) {

    @ExternalDirective
    var items: List<BasketItem> 

    @RequiresDirective(FieldSet("items"))
    fun image(): String {
        // logic for image retrieval
    }
}
`

Unfortunately, before generating the schema FederatedSchemaValidator rejects such scenario.  My guess is that we are allowed only to reference the primitive types to allow the Apollo Gateway build correctly the QueryPlan and fetch the necessary data. Is this correct ? Are there any plans to allow more complex references like my case?

dariuszkuc:
[FieldSet](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#scalar-_fieldset) has to represent valid set of fields. You cannot select just `items` as it is a list of objects as it has to be valid GraphQL field selection, e.g. you have to explicitly specify  the fields that should be available -> `@RequiresDirective(FieldSet("items { id type dateAdded }"))` (assuming you need `id`, `type` and `dateAdded` fields there).

NOTE: I am unsure whether `@requires` works with a list (but based on validations it appears that it should -> https://www.apollographql.com/docs/apollo-server/federation/errors/#requires) so you will have to test it out.

dariuszkuc:
I just tried it out and it is valid to select list items in `@requires` directive. Unfortunately we apply incorrect validation of `@requires` directive field set so it blows up at startup.

See https://github.com/dariuszkuc/graphql-kotlin/tree/federation_requires_list for a working example with disabled validations.

[...]

PR #685, opened 2020-04-30
Allow field set selection to be on lists

smyrick:
### :pencil: Description
The federation spec specifies that the `FieldSet` scalar can contain a nested selection

https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#scalar-_fieldset

And it doesn't say anything about lists so there for we should allow lists to be marked with `@requires`

https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#requires

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/637


graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/validation/validateFieldSet.kt
```diff
@@ -32,8 +31,7 @@ internal fun validateKeySetField(targetField: GraphQLFieldDefinition?, extendedT
             errors.add("$validatedDirective specifies invalid field set - type incorrectly references external field=${targetField.name}")
         }
 
-        when (GraphQLTypeUtil.unwrapNonNull(targetField.type)) {
-            is GraphQLList -> errors.add("$validatedDirective specifies invalid field set - field set references GraphQLList, field=${targetField.name}")
+        when (targetField.type.unwrapType()) {
```
smyrick, line 34: This unwrap call handles all combinations of lists and nullable types

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/validation/validateFieldSet.kt
```diff
@@ -32,8 +31,7 @@ internal fun validateKeySetField(targetField: GraphQLFieldDefinition?, extendedT
...
             is GraphQLInterfaceType -> errors.add("$validatedDirective specifies invalid field set - field set references GraphQLInterfaceType, field=${targetField.name}")
```
smyrick, line 35: Question: Do we need to check all these other types too? If it can just be on any `FIELD_DEFINITION` why do we need to exclude interfaces and unions?

dariuszkuc:
While this fixes the `@requires` directive it breaks `@key` directive validation which afaik cannot contain lists, unions or interfaces -> see https://www.apollographql.com/docs/apollo-server/federation/errors/#key

smyrick:
@dariuszkuc Update the changes

smyrick:
The requires integration tests is taken from the example in the federation spec: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#requires

smyrick:
Actions are not being triggered? Going to close and re-open

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/validation/validateFieldSet.kt
```diff
@@ -0,0 +1,50 @@
...
+    if (null != targetField) {
+        val externalField = targetField.getDirective(EXTERNAL_DIRECTIVE_NAME) != null
+        if (extendedType && !externalField) {
+            errors.add("$errorMessage specifies invalid field set - extended type incorrectly references local field=${targetField.name}")
+        } else if (!extendedType && externalField) {
+            errors.add("$errorMessage specifies invalid field set - type incorrectly references external field=${targetField.name}")
+        }
+
+        /**
+         * The @requires directive can have [list,interface,union] in the field selection.
```
dariuszkuc, line 37: I think those checks are only applicable on `@key` directive, e.g. don't see anything specific about those for `@provides` either (https://www.apollographql.com/docs/apollo-server/federation/errors/#provides)

dariuszkuc: NVM :)

PR #697, opened 2020-05-07
Add GraphQLTypeRestriction feature

smyrick:
### :pencil: Description
Add new annotation for the schema generator that allows users to restrict a class to a specific GraphQL input/output type

### :link: Related Issues
N\A

### DIFF
--- a/docs/customizing-schemas/restricting-types.md
+++ b/docs/customizing-schemas/restricting-types.md
@@ -1,0 +1,20 @@
+---
+id: restricting-types
+title: Restricting Classes to GraphQL Types
+---
+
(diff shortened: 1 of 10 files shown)


smyrick:
There has been some discussion internally about the actual need for this feature. We are asking for feedback from the community on if a feature like this is needed or if anyone has a better idea on how to implement the required end result

smyrick:
Going to closes this as it may be to specific of a feature and technically we still generate a valid schema. 

This should be more of a schema linter check or custom hook instead. 

We can re-open if this comes up later

dariuszkuc:
I think it would be awesome to provide GraphQL schema linter through some plugin mechanism - maybe create a separate issue to track it?

Issue #698, opened 2020-05-08
@GraphQLID support for List<ID>


lennyburdette:
I might be missing something—is it possible to use the `@GraphQLID` annotation on a `List<String>`? 

```kotlin
class Query {
  fun thingsByID(@GraphQLID ids: List<String>): List<Thing> = ...
```

When version 2.1.1, I get:
```
com.expediagroup.graphql.exceptions.InvalidIdTypeException: List is not a valid ID type, only Strings are accepted
```

I have a workaround, but I wanted to check to see if I was just doing something wrong:
```kotlin
typealias ID = String

class Query {
  fun thingsByID(ids: List<ID>): List<Thing> = ...
}

class CustomSchemaGeneratorHooks(
  federatedTypeRegistry: FederatedTypeRegistry
) : FederatedSchemaGeneratorHooks(federatedTypeRegistry) {
  override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier as? KClass<*>) {
    ID::class -> Scalars.GraphQLID
    else -> super.willGenerateGraphQLType(type)
  }
}
```

Thanks!

smyrick:
@lennyburdette Unfortunately there is not a better way right now to accomplish a GraphQL list of ids (`[ID]`) in a better way than what you have commented with setting custom hooks.

We had a discussion about this when changing the annotation support to only work on strings about including a wrapper `ID` type. https://github.com/ExpediaGroup/graphql-kotlin/issues/317

At the time though we had used a full class and not a typealias. Does the typealias show up in the reflection check? Because if it does we could include this in the library so you could use the typealias or the annotation.

smyrick:
I did a little code check locally and the check `type.classifier == ID::class` returns true because ID::class is just `String` so this doesn't actually tell us if you have used `List<ID>` or `List<String>`

smyrick:
I suppose we could go back to creating a wrapper class that only accepts a String value. This could mean we could clean up a lot of the logic we have around having to pass the `annotatedAsID` boolean around because we lose the annotation information

```kotlin
/**
 * Used to represent a GraphQL ID scalar type
 * which must serialize to a string value
 */
class ID(val value: String)
```

lennyburdette:
Awesome! Not expecting a blazing fast response!

[...]

PR #700, opened 2020-05-08
Remove GraphQLID annotation for ID class

smyrick:
### :pencil: Description
Add a simple wrapper class that will allow us to support `List<ID>`

Both serialization and deserialization work with the wrapper class

Screen Shot 2020-05-08 at 1 43 21 PM

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/698


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/types/ID.kt
```diff
@@ -0,0 +1,29 @@
...
+package com.expediagroup.graphql.types
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonValue
+
+/**
+ * Used to represent a GraphQL ID scalar type
+ * which must serialize/deserialize to a string value
+ */
+data class ID(@get:JsonIgnore val value: String) {
```
dariuszkuc, line 26: there is no need for `@JsonIgnore` as `@JsonValue` will be used for serialization anyway

dariuszkuc:
If we are adding a wrapper `ID` class do we want to support other types implicitly supported by `graphql-java`?

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/types/ID.kt
smyrick: I get the following error if I don't exclude the value getter with `JsonIgnore`

```
Cannot construct instance of `com.expediagroup.graphql.types.ID` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('1')
```

dariuszkuc: ah, without explicit `@JsonCreator` it complains, moving the `@JsonValue` on the parameter itself fixes it

```
fun main() {
    val mapper = jacksonObjectMapper()

    val writer = StringWriter()
    mapper.writeValue(writer, ID("ABCD"))
    println("serialized: ${writer.toString()}")
    val deserialized = mapper.readValue<ID>(writer.toString())
    println("deserialized: $deserialized")
}

data class ID(@get:JsonValue val value: String)
```

smyrick:
> If we are adding a wrapper `ID` class do we want to support other types implicitly supported by `graphql-java`?

Deserializing is going to be a little more difficult with the generics. Do you know how to set this up because I am getting the following error just changing `String` to `<T>`

```
Exception while fetching data (findPersonById) : class sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl cannot be cast to class java.lang.Class (sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl and java.lang.Class are in module java.base of loader 'bootstrap')
```

dariuszkuc:
Ah that error is actually due to the type erasure.... so at runtime it is just `Object` so it cannot be properly serialized. I'm using reified generics to fix this in graphql client but I don't think they will work here.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/types/ID.kt
smyrick: The following code doesn't serialize properly

```kotlin
data class ID(@get:JsonValue val value: String)
```

Screen Shot 2020-05-08 at 2 12 32 PM

dariuszkuc: If you run my example code you will see:
```
serialized: "ABCD"
deserialized: ID(value=ABCD)
```

`ID(value=1)` implies default `toString()` invocation and/or misconfiguration of Jackson.

smyrick:
Currently this PR is backwards compatible. I am going to make a commit which remove the `@GraphQLID` annotation though as a breaking change since we are already going to be releasing 3.0.0 due to a breaking change bug

docs/writing-schemas/scalars.md
```diff
@@ -25,39 +25,38 @@ extended scalar types provided by `graphql-java`.
 ## ID
 
 GraphQL supports a the scalar type ID, a unique identifier that is not intended to be human readable. ID's are
-serialized as a String. To mark given field as an ID field you have to annotate it with `@GraphQLID` annotation.
+serialized as a String. To mark given field as an ID field you can use the wrapper class `com.expediagroup.graphql.types.ID`
```
dariuszkuc, line 28:
```suggestion
serialized as a String. In order to expose a GraphQL `ID` field you have use `com.expediagroup.graphql.types.ID` class that wraps underlying String value.
```

docs/writing-schemas/scalars.md
```diff
@@ -25,39 +25,38 @@ extended scalar types provided by `graphql-java`.
...
 data class Person(
-  @GraphQLID
-  val id: String,
-  val name: String
+    val id: ID,
+    val name: String
 )
 
-fun locatePerson(@GraphQLID id: String) = Person(id, "Jane Doe")
+fun locatePersonWithId(id: ID) = Person(id, "John Smith")
```
dariuszkuc, line 39: lets rename this to `findPersonById` so it is consistent with example app

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/types/ID.kt
```diff
@@ -14,12 +14,16 @@
...
+import com.fasterxml.jackson.annotation.JsonValue
 
 /**
- * Throws when the KClass is not one of the supported types for a GraphQLID
+ * Used to represent a GraphQL ID scalar type
+ * which must serialize/deserialize to a string value
  */
-class InvalidIdTypeException(kClass: KClass<*>) :
-    GraphQLKotlinException("${kClass.simpleName} is not a valid ID type, only Strings are accepted")
+data class ID(@get:JsonIgnore val value: String) {
```
dariuszkuc, line 26: wondering whether we should rename this class to something explicit like `GraphQLID` (unsure about capitalization there...)

smyrick: I was debating this as well. The full package paths makes it pretty clear that it is a GraphQL ID but I was thinking since the name of the scalar in the schema is ID that it would make more sense to have it be the same here, because if we call it `GraphQLID` then this

```kotlin
data class Person(
    val id: GraphQLID,
    val name: String
)
```

becomes this 

```graphql
type Person {
    id: ID!
    name: String!
}
```

docs/writing-schemas/scalars.md
```diff
@@ -25,39 +25,38 @@ extended scalar types provided by `graphql-java`.
 ## ID
 
 GraphQL supports a the scalar type ID, a unique identifier that is not intended to be human readable. ID's are
```
DelfCo, line 27: "IDs". No apostrophe.
For some reason, GitHub won't let me edit the file.

docs/writing-schemas/scalars.md
```diff
@@ -25,39 +25,38 @@ extended scalar types provided by `graphql-java`.
...
-serialized as a String. To mark given field as an ID field you have to annotate it with `@GraphQLID` annotation.
+serialized as a String. In order to expose a GraphQL `ID` field you have use `com.expediagroup.graphql.types.ID` class that wraps underlying String value.
```
DelfCo, line 28: "To expose a GraphQL `ID` field, you must use the `com.expediagroup.graphql.types.ID` class, which wraps the underlying `String` value.

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/types/ID.kt
dariuszkuc: 👍 lets keep it as just `ID`

docs/writing-schemas/scalars.md
smyrick: @DelfCo Did you change your username?

I have added you as a `Triage` maintainer for the repo so you should be able to make suggestions

DelfCo: Oh. I see what happened. I was commenting on this one from my desktop at home, which is signed into my personal GitHub account rather than my Expedia Employee GitHub account. Thanks for the add. I'll try to maintain better sign-in discipline in future to avoid confusion about who I am and what I'm doing here.

PR #702, opened 2020-05-11
Add extends directive if not defined

smyrick:
### :pencil: Description
Since the extends directive is always used in federation on the Query object we should always have the directive definition in the schema. However it may already be added if the generated schema uses it, so only add the definition if it is not yet defined

The schema we generate today will still work on the server, it just fails if you happen to run through the generated SDL through the Apollo CLI or Gateway.

### :link: Related Issues
N\A

### DIFF
--- a/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
@@ -19,15 +19,16 @@ package com.expediagroup.graphql.federation
 import com.expediagroup.graphql.annotations.GraphQLName
 import com.expediagroup.graphql.directives.DEPRECATED_DIRECTIVE_NAME
 import com.expediagroup.graphql.extensions.print
-import com.expediagroup.graphql.federation.directives.EXTENDS_DIRECTIVE_NAME
-import com.expediagroup.graphql.federation.directives.EXTERNAL_DIRECTIVE_NAME
+import com.expediagroup.graphql.federation.directives.EXTENDS_DIRECTIVE_TYPE
+import com.expediagroup.graphql.federation.directives.EXTERNAL_DIRECTIVE_TYPE
 import com.expediagroup.graphql.federation.directives.FieldSet
 import com.expediagroup.graphql.federation.directives.KEY_DIRECTIVE_NAME
-import com.expediagroup.graphql.federation.directives.PROVIDES_DIRECTIVE_NAME
-import com.expediagroup.graphql.federation.directives.REQUIRES_DIRECTIVE_NAME
-import com.expediagroup.graphql.federation.directives.extendsDirectiveType
+import com.expediagroup.graphql.federation.directives.KEY_DIRECTIVE_TYPE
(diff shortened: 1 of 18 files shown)


graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorTest.kt
```diff
@@ -141,7 +141,28 @@ class FederatedSchemaGeneratorTest {
...
+              ) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+            "Marks the field or enum value as deprecated"
+            directive @deprecated(
+                "The reason for the deprecation"
+                reason: String! = "No longer supported"
+              ) on FIELD_DEFINITION | ENUM_VALUE
+
+            "Marks target object as extending part of the federated schema"
+            directive @extends on OBJECT | INTERFACE
```
smyrick, line 163: Before my fix when I set `includeDirectives = true` below, this definition was not there

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt
```diff
@@ -72,13 +72,17 @@ open class FederatedSchemaGeneratorHooks(private val federatedTypeRegistry: Fede
...
         val federatedCodeRegistry = GraphQLCodeRegistry.newCodeRegistry(originalSchema.codeRegistry)
-        val federatedSchema = GraphQLSchema.newSchema(originalSchema)
+        val federatedSchemaBuilder = GraphQLSchema.newSchema(originalSchema)
         val federatedQuery = GraphQLObjectType.newObject(originalQuery)
             .field(SERVICE_FIELD_DEFINITION)
             .withDirective(extendsDirectiveType)
 
+        // Add the @extends directive definition if it is not defined as we are using it on the Query object
+        if (originalSchema.getDirective(EXTENDS_DIRECTIVE_NAME) == null) {
+            federatedSchemaBuilder.additionalDirective(extendsDirectiveType)
```
dariuszkuc, line 83: should we add all federated directives here?

smyrick: The only other things we do with the `federatedSchemaBuilder` is to add the `_Any` scalar and update the `Query` object. So all the other directive definitions will be their if they are used, otherwise if they are not there it means they are not used by the schema do we don't need to add it

dariuszkuc: I was wondering whether all federated directives should always be present (i.e. similar behavior to default `include`, `skip` and `deprecated`).

smyrick: I suppose we could. It looks like the apollo-federation lib adds them to the schema

See this running example of federation and the schema tab: https://pw678w138q.sse.codesandbox.io/graphql

Issue #706, opened 2020-05-13
[plugin] generate client tasks should support test sources


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
Client generation logic currently generates code that is automatically added to main source sets. This is problematic if we want to use the plugins to generate test clients to be used in integration tests - since generated code is added to the main source set, those test clients will be part of the final generated jar.

**Describe the solution you'd like**
Generate client Gradle task/Maven MOJO should be configurable to allow generation of test source sets.

**Describe alternatives you've considered**
Using Gradle clients could create custom tasks that would modify the source sets but that is pretty hacky solution.

PR #718, opened 2020-05-17
[plugin] support generating graphql client test sources

dariuszkuc:
### :pencil: Description

Create new `graphqlGenerateTestClient` Gradle task and `generateTestClient` Maven Mojo to generate test clients. Functionality wise they are the same as corresponding `graphqlGenerateClient` task / `generateClient` Mojo, they only differ in default value of parameters (e.g. `generate-test-sources` build phase for Mojo, using `src/test/resources`, etc).

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/706


dariuszkuc:
I think it might be cleaner if we have explicit tasks for main and test source generation (at least for Maven due to implied default execution phase, query file directory and output file directory). I'm going  to refactor this to have separate tasks/mojos.

PR #733, opened 2020-05-30
[build] fix publishing of Gradle plugin marker

dariuszkuc:
### :pencil: Description

Current logic was incorrectly attaching project sources+javadoc to the Maven publication. Since Gradle plugin produces two publications -> plugin + plugin marker both of them were uploaded with attached plugin sources+javadoc jars.


dariuszkuc:
See: https://search.maven.org/artifact/com.expediagroup.graphql/com.expediagroup.graphql.gradle.plugin/3.0.0-RC7/pom

PR #749, opened 2020-06-06
remove package-lock from example federated gateway project

dariuszkuc:
### :pencil: Description

Example Federated Gateway is not published library and it is just provided as an example to run with federated GraphQL Kotlin services. There is no point in publishing the lock file.

Also resolves https://github.com/apollographql/apollo-server/security/advisories/GHSA-w42g-7vfc-xf37 which dependabot flagged as potential vulnerability based on the `package-lock.json`.


smyrick:
@dariuszkuc Even if this is not part of the published library we still want to include a lock file so that we know the example will always be running the same versions for others testing it out.

The `yarn.lock` is included in the gitignore because we should be using `npm` to install dependecies.

To resolve the security issue we can update the dependecy instead

smyrick:
See https://github.com/ExpediaGroup/graphql-kotlin/pull/750

PR #759, opened 2020-06-10
Clean up docs

smyrick:
### :pencil: Description
Take another pass at documentation cleanup

* Remove graphql descriptions in pages that are about other topics
* Rename top level page names
* Add proper header levels

### :link: Related Issues
N\A

### DIFF
--- a/docs/schema-generator/execution/exceptions.md
+++ b/docs/schema-generator/execution/exceptions.md
@@ -3,21 +3,30 @@ id: exceptions
 title: Exceptions and Partial Data
 ---
 
+## Returning GraphQL Errors
+
 Exceptions thrown during execution of a query will result in a GraphQLError that is added to a list of errors of the result. See
 [graphql-java documentation](https://www.graphql-java.com/documentation/v14/execution/) for more details on how to customize your exception handling.
 
-### Partial Data
 
-GraphQL allows you to return both data and errors in a single response. Depending on the criticality of the encountered error you may want to return
-partial data together with the corresponding errors. In Kotlin, functions return only a single value, which means that in order to return both data
+```kotlin
+fun getRandomNumberOrError(): Int {
+    val num = Random().nextInt(100)
+    return if (num <= 50) num else throw Exception("number is greater than 50")
+}
+```
+
+## Returning Data and Errors
+
+GraphQL allows you to return both data and errors in a single response, as long as the data returned still matches the schema. Depending on the criticality of the encountered error you may want to return
+null for some fields with the corresponding errors. In Kotlin, functions return only a single value, which means that in order to return both data
 and errors you have to explicitly return them wrapped in a `DataFetcherResult` object.
 
 ```kotlin
 class DataAndErrorsQuery {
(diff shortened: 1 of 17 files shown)


docs/schema-generator/execution/exceptions.md
```diff
@@ -3,21 +3,30 @@ id: exceptions
...
+fun getRandomNumberOrError(): Int {
+    val num = Random().nextInt(100)
+    return if (num <= 50) num else throw Exception("number is greater than 50")
+}
+```
+
+## Returning Data and Errors
+
+GraphQL allows you to return both data and errors in a single response, as long as the data returned still matches the schema. Depending on the criticality of the encountered error you may want to return
+null for some fields with the corresponding errors. In Kotlin, functions return only a single value, which means that in order to return both data
```
dariuszkuc, line 22: while previous sentence specifies that result has to match the schema i think this sentence might be somewhat confusing (i.e. nulls are only allowed for nullable fields). Maybe we should clarify that it should be some `default data or null for optional fields` or something like that

docs/schema-generator/writing-schemas/lists.md
```diff
@@ -34,20 +34,25 @@ type Query {
...
+| `kotlin.IntArray`     |
+| `kotlin.LongArray`    |
+| `kotlin.ShortArray`   |
+| `kotlin.FloatArray`   |
+| `kotlin.DoubleArray`  |
+| `kotlin.CharArray`    |
+| `kotlin.BooleanArray` |
 
 > NOTE: Underlying GraphQL types of primitive arrays will be corresponding to the built-in scalar types or extended
 > scalar types provided by `graphql-java`.
```
dariuszkuc, line 53: graphql-java no longer supports extended scalars -> think we should update the docs to indicate that as well

docs/schema-generator/writing-schemas/scalars.md
```diff
@@ -12,7 +13,7 @@ extended scalar types provided by `graphql-java`.
 | `kotlin.Float`      | `Float`      |
 | `kotlin.String`     | `String`     |
 | `kotlin.Boolean`    | `Boolean`    |
-| | **Extended GraphQL Types**       |
+| **Extended GraphQL Types** | **GraphQL Type** |
```
dariuszkuc, line 16: I believe graphql-java no longer supports extended types (by default) -> lets double check on it

smyrick: It says these are the current supported scalars

https://www.graphql-java.com/documentation/v15/scalars/

dariuszkuc: Maybe they haven't updated the docs -> https://github.com/graphql-java/graphql-java/releases/tag/v15.0

```
Any special handling for non standard Scalars got removed. This means you have to declare any Long, Char, etc Scalar explicitly in your SDL and provide an implementation.

The implementation for these non standard Scalars are still available in GraphQL Java itself (in the Scalars class) but they will be removed in the next release. The extended scalars project is the recommend way to migrate.
```

PR: https://github.com/graphql-java/graphql-java/pull/1780

Issue #762, opened 2020-06-11
Directive arguments support @GraphQLIgnore


oatmeal-raisin:
Creating a new directive with an ignored field will prevent the schema from being generated when the ignored field type is not in the supported packages. Is there a way to prevent this behavior? It seems like a type that is only referred to in an ignored field should not break schema generation even if it is not in the supported packages.

Example in a directive:
```
@GraphQLDirective()
annotation class TestDirective(
        val value: String,
        @GraphQLIgnore
        val ignore: Array<KClass<*>> = [] 
)
```
Results in
`Caused by: com.expediagroup.graphql.exceptions.TypeNotSupportedException: Cannot convert kotlin.reflect.KClass<*> since it is not a valid GraphQL type or outside the supported packages [...]`

Obviously something like kotlin.reflect should not be added to the supported packages and nothing representing KClass should be in the schema.

dariuszkuc:
Currently `@GraphQLIgnore` is applied on GraphQL fields only (i.e. functions and properties), we do not apply it against the directive definitions. 

What's the use case behind hiding some of the directive field values?

oatmeal-raisin:
Thank you for the quick response @dariuszkuc.

The current use case is driven by trying to simplify validation. The goal is to create an annotation that exposes constraint information on the SDL as a directive but can also be used as a javax validation constraint annotation. Constraint annotations have fields that do not make sense to include in directives like "groups" and "payload." 

Given that it is not currently supported, what is the best way to work around the problem of having the field ignored and preventing the type on that field from being pulled in? Is it possible with schema generation hooks?

smyrick:
@oatmeal-raisin One way you could solve the issue right now is to have the two different annotations on your kotlin code, one for the GraphQL information and one for the kotlin execution, or with the hooks it would be possible to modify the generated type before it is added to the schema.

However I think you raise a valid point and there is a place in the library code where we can add the requested feature.

dariuszkuc:
Yeah I think we should support it.

oatmeal-raisin:
Thanks @smyrick and @dariuszkuc. I really appreciate how responsive your team is and I am looking forward to the enhancement. Are there any hooks I can take a look at that do something similar? I will need to both exclude the field and to prevent the type on that ignored field from causing an exception because it is not in the configured packages.

smyrick:
[...]

PR #763, opened 2020-06-11
[generator] Include java getters for ignore checks

smyrick:
### :pencil: Description
Update the property generator code to check the getters for the `@GraphQLIgnore` annotation. This means that any Kotlin property can use both `@GraphQLIgnore` and `@get:GraphQLIgnore`, but it also solve the issue with directive arguments that the generated JVM code only has annotation on the argument getters.

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/762


docs/schema-generator/customizing-schemas/directives.md
```diff
@@ -153,3 +153,34 @@ Directives are applied in the order annotations are declared on the given object
 ```
 
 `Directive1` will be applied first followed by the `Directive2`.
+
+## Ignoring Directive Arguments
+
+Normally if you wanted to exclude a field or argument from the schema, you could use [@GraphQLIgnore](./excluding-fields.md).
+However, due to reflection limitations, the generator cannot cast annoations cannot back to the implementing classes to see what annotations are on the arguments.
```
dariuszkuc, line 160:
```suggestion
However, due to reflection limitations, the generator cannot cast annotations back to the implementing classes to see what annotations are on the arguments.
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
```diff
@@ -57,7 +57,9 @@ private fun getDirective(generator: SchemaGenerator, directiveInfo: DirectiveInf
         }
 
         val directiveClass = directiveInfo.directive.annotationClass
-        directiveClass.getValidProperties(generator.config.hooks).forEach { prop ->
+        val directiveArguments = directiveClass.getValidProperties(generator.config.hooks)
```
dariuszkuc, line 60: whats the point of creating separate value? I guess this is a leftover from the refactor?

smyrick: Yea, I was using this to inspect the value and I was trying before the merge separate values together.

dariuszkuc:
Hmm this might be some limitation of reflections, e.g. given instance of `Annotation` we do access member properties which are instances of `KPropertyImpl` (which in turn implements `KAnnotatedElement`) and we use those to generate directive information (https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt#L61).

I played around with different annotation use-site targets and I only got it to work when explicitly using `get` and reading annotation from generated getter (modified https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/types/GenerateDirectiveTest.kt#L34), e.g.

```
@GraphQLDirective
annotation class DirectiveWithString(@get:GraphQLIgnore val string: String)

// line 61 in generateDirective
prop.getter.findAnnotation<GraphQLIgnore>() // finds it
```

Could not get it to work by accessing constructor parameters or attempting to read annotations directly from the field.

smyrick:
@oatmeal-raisin So it appears we have two options:

1. 🚀  Don't add extra code to the library and require developers to filter out specific arguments with hooks, and update the documentation (this PR would accomplish this)

```kotlin
@GraphQLDirective
annotation class DirectiveWithString(val string: String)

val hooks = object : SchemaGeneratorHooks {
    override fun isValidProperty(kClass: KClass<*>, property: KProperty<*>): Boolean {
        if (kClass == DirectiveWithString::class && property.name == "string") {
            return false
        }
        return super.isValidProperty(kClass, property)
    }
}
```

--------------

2. ❤️  Add the checks above on annotation property getters which requires users to use `@get:GraphQLIgnore` specifically for annotation arguments.

```kotlin
@GraphQLDirective
annotation class DirectiveWithString(@get:GraphQLIgnore val string: String)
```
-----------

@ExpediaGroup/graphql-kotlin-committers and other contributors, let's vote with the following reactions 🚀  or ❤️ , on each option

dariuszkuc:
Hooks can be used regardless whether we provide additional support through the annotation or not.

It might be a JVM limitation that you annotating annotation parameters/fields does nothing - but it looks like annotating functions (i.e. getters) works so we should update lib to allow it. It is suboptimal that you have to explicitly specify `get` site target when applying `@GraphQLIgnore` on annotations but IMHO it is better than not supporting it at all.

dariuszkuc:
It is due to the JVM -> Kotlin `annotation class` is a syntactic sugar over Java `@interface`. Interfaces in Java don't store any values and only expose methods (or constants) so it makes sense annotation will only work on methods.

oatmeal-raisin:
I've tested the hook and it worked for me and I am not not sure that it will be a common need. If there is future need from other users, option 2 might be worth the investment. It would be pretty painful to fracture the pretty syntax based on the context though. I would suggest allowing the get target to optionally be specified even for types, if building out option 2.

As a side note, the usage I was originally after, an annotation for SDL directives and javax constraints may not be possible for another reason. The constraint annotation will not be picked up unless it targets get. If the annotation targets get, it will not work for directive generation. Even with support for ignoring fields in directives, my complicated case I will require hooks or other customizations.

smyrick:
I will update this PR to include the logic to check the getter. It isn't to much more logic and it is a unpopular edge case. So we will update the docs around it as well. But as mentioned, with the hooks you can pretty much do any custom behaviour you want around schema generation.

As for validation you can see an example [query here](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/query/ValidatedQuery.kt)

oatmeal-raisin:
Thanks @smyrick, I was trying to build on that example to build a combined annotation. Something along the lines of the sample below, to both expose the constraint on the SDL and enforce it in the service.

```
data class ValidatedDataClass {
        @field:PatternConstraintDirective(pattern = "[^#]*", message = "id must not contain #")
        val id: String? = null
}

@GraphQLDirective(
        locations = [
            Introspection.DirectiveLocation.FIELD_DEFINITION,
            Introspection.DirectiveLocation.FIELD,
            Introspection.DirectiveLocation.INPUT_FIELD_DEFINITION,
            Introspection.DirectiveLocation.INPUT_OBJECT
        ],
        name = "constraint"
)
@Constraint(validatedBy = [PatternConstraint::class])
annotation class PatternConstraintDirective(
        val pattern: String,
        val message: String,
        val groups: Array<KClass<*>> = [],
        val payload: Array<KClass<*>> = []
)

class PatternConstraint: ConstraintValidator<PatternConstraintDirective, String> {
  ....
}

```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/extensions/kPropertyExtensions.kt
```diff
@@ -27,6 +28,7 @@ import kotlin.reflect.KProperty
 internal fun KProperty<*>.isPropertyGraphQLIgnored(parentClass: KClass<*>): Boolean = when {
     this.isGraphQLIgnored() -> true
     getConstructorParameter(parentClass)?.isGraphQLIgnored().isTrue() -> true
+    this.javaGetter?.isGraphQLIgnored().isTrue() -> true
```
dariuszkuc, line 31: instead of `javaGetter` why not just `getter`?

smyrick: Yep, I can use that instead and update the extensions

PR #766, opened 2020-06-12
Upgrade dependencies and run ktlint format

smyrick:
### :pencil: Description
Non functional changes, just doing to dependency updates and since Ktlint added some more rules we had to reformat some code.

New indent rule in 0.37.0 is the major cause: https://github.com/pinterest/ktlint/releases/tag/0.37.0

If you want to view the changes that are not whitespace, check the option in the files tab of the PR

Screen Shot 2020-06-12 at 12 27 59 PM

### :link: Related Issues
N\A


examples/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.0.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
```
smyrick, line 3: Matching the wrapper version in the root

dariuszkuc:
Sounds like we need to update our root `build.gradle.kts` to correctly apply ktlint? Or am I missing something?

smyrick:
> Sounds like we need to update our root `build.gradle.kts` to correctly apply ktlint? Or am I missing something?

Yea, we don't have an auto-format currently

dariuszkuc:
Unsure if we need to apply auto-format but build should definitely fail if formatting doesn't match.

smyrick:
> Unsure if we need to apply auto-format but build should definitely fail if formatting doesn't match.

Oh, it did fail that is why I had to do all these updates

dariuszkuc:
So how did it worked in the past?

smyrick:
Not sure why the build is failing in GH. Passes locally with the wrapper for me, both library and examples

dariuszkuc:
Ah nvm those are new ktlint rules...

smyrick:
New indent rule in 0.37.0 is the major cause: https://github.com/pinterest/ktlint/releases/tag/0.37.0

PR #767, opened 2020-06-12
Publish doc version on library release

smyrick:
### :pencil: Description
Docusaurus can publish different versions.

To make it easy we can just have a version published for every library version. The latest docs will always be available too

https://docusaurus.io/docs/en/tutorial-version

I have included the version 3.1.1 on the first check in

### :link: Related Issues
N\A


dariuszkuc:
I don't think we need to cut version per each release.... I'd only do it per major

smyrick:
@dariuszkuc Would it be easier to just always release a version and then delete minor/patch versions?

dariuszkuc:
While cutting new doc version on each release would certainly be easier it doesn't mean that we should do it (IMHO it would lead to unnecessary churn on the git history).

I think for now we can follow manual process.

smyrick:
Ok, then for now we don't need a PR and we can cut a new version of the docs when we do the same for the library in 4.0.0

PR #768, opened 2020-06-16
enable official Kotlin style

dariuszkuc:
See: https://kotlinlang.org/docs/reference/code-style-migration-guide.html


dariuszkuc:
Actually I'm not sure whether we need the extra setting. It looks like it is the standard starting with Kotlin 1.3

Issue #764, opened 2020-06-12
Passing unique headers without recreating client


nutgaard:
I may obviously be missing something here. 

But is there a way of passing headers to `GraphQLClient.execute` when using the autogenerated query-classes?

Usecases may differ, but I need to pass a callId to allow tracking between services, and recreating the client for each request seems a bit redundant.

dariuszkuc:
Recreating client per each request is indeed an unnecessary overhead. I think we can probably expose the underlying Ktor client functionality to configure the headers (see: https://github.com/ExpediaGroup/graphql-kotlin/pull/769).

Currently you can do a workaround with OkHttp engine where you can add custom headers using interceptors (which probably would have to rely on some thread locals).

nutgaard:
Tried looking into it myself when creating the issue, and seems like we came to approx the same solution ( https://github.com/nutgaard/graphql-kotlin/pull/1/files ) 👍 

I exposed `HttpRequestBuilder` insteadof `HeadersBuilder`, but I'm rather new to this library and may be overlooking some trivial reason why that would not work.

In the end I ended up with a workaround similar to what you're describing, though using a ktor-client feature and some coroutine-context black-magic.

PR #769, opened 2020-06-16
[client] expose request builder to allow customization per query execution

dariuszkuc:
### :pencil: Description

Update client to allow specifying optional Ktor HttpRequestBuilder block that allows usage of DSL for specifying custom headers as well as provide some additional configuration options like timeouts per request, specify URL query parameters as well as custom attributes that could be referenced from the feature pipeline.

After client is generated we can configure it as follows:

```kotlin
val client = GraphQLClient(
  url = URL("http://localhost:8080/graphql"),
  engineFactory = OkHttp,
  mapper = jackson
) {
  // existing global client customization through Ktor HttpClientConfig
  engine {
    config {
      connectTimeout(10, TimeUnit.SECONDS)
      readTimeout(60, TimeUnit.SECONDS)
      writeTimeout(60, TimeUnit.SECONDS)
    }
  }
  install(Logging) {
    logger = Logger.DEFAULT
    level = LogLevel.HEADERS
  }
}
// create client for specific query
val helloWorldQuery = HelloWorldQuery(client)
val result = helloWorldQuery.execute(variables = HelloWorldQuery.Variables(name = null)) {
  // per query customization using Ktor HttpRequestBuilder
  header("X-B3-TraceId", "0123456789abcdef")
}
```

See https://api.ktor.io/1.3.2/io.ktor.client.request/-http-request-builder/ for available customization options.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/764

### DIFF
--- a/docs/client/client-customization.md
+++ b/docs/client/client-customization.md
@@ -13,13 +13,13 @@ See [Ktor HTTP Client documentation](https://ktor.io/clients/index.html) for add
 
 ### Global Client Customization
 
-Single instance of `GraphQLClient` can be used to power many GraphQL queries. You can specify target engine factory and
-configure it through corresponding [HttpClientConfig](https://api.ktor.io/1.3.2/io.ktor.client/-http-client-config/index.html).
-Ktor HTTP client provides a number of [standard features](https://ktor.io/clients/http-client/features.html) as well as
+A single instance of `GraphQLClient` can be used to power many GraphQL operations. You can specify a target engine factory and
+configure it through the corresponding [HttpClientConfig](https://api.ktor.io/1.3.2/io.ktor.client/-http-client-config/index.html).
+Ktor also provides a number of [standard HTTP features](https://ktor.io/clients/http-client/features.html) and
 allows you to easily create custom ones that can be configured globally.
 
-Example below configures new `GraphQLClient` to use `OkHttp` engine with custom timeouts, adds default `X-MY-API-KEY`
-header to all requests as well enables basic logging of the requests.
+The below example configures a new `GraphQLClient` to use the `OkHttp` engine with custom timeouts, adds a default `X-MY-API-KEY`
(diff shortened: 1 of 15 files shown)


smyrick:
As mentioned in the issue, would it be better to just expose the entire client builder so anyone can edit any part of the client? There may be more feature requests down the road and this way we are just exposing everything to the client developers and not hiding anything behind the library.

docs/client/client-customization.md
```diff
@@ -9,6 +9,18 @@ title: Client Customization
...
+
+### Global Client Customization
+
+Single instance of `GraphQLClient` can be used to power many GraphQL queries. You can specify target engine factory and
+configure it through corresponding [HttpClientConfig](https://api.ktor.io/1.3.2/io.ktor.client/-http-client-config/index.html).
+Ktor HTTP client provides a number of [standard features](https://ktor.io/clients/http-client/features.html) as well as
+allows you to easily create custom ones that can be configured globally.
+
+Example below configures new `GraphQLClient` to use `OkHttp` engine with custom timeouts, adds default `X-MY-API-KEY`
+header to all requests as well enables basic logging of the requests.
```
smyrick, line 22:
```suggestion
A single instance of `GraphQLClient` can be used to power many GraphQL operations. You can specify a target engine factory and
configure it through the corresponding [HttpClientConfig](https://api.ktor.io/1.3.2/io.ktor.client/-http-client-config/index.html).
Ktor also provides a number of [standard HTTP features](https://ktor.io/clients/http-client/features.html) and
allows you to easily create custom ones that can be configured globally.

The below example configures a new `GraphQLClient` to use the `OkHttp` engine with custom timeouts, adds a default `X-MY-API-KEY`
header to all requests, and enables basic logging of the requests.
```

smyrick: With the above changes committed, the rest of the PR looks good

Issue #765, opened 2020-06-12
Documentation improvement


olivermt:
**Is your feature request related to a problem? Please describe.**

First off: this project is awesome, so don't take my lengthy complaining as too hard of a criticism :)

There is a lack in the documentation on some common use-cases or if there is documnetation it is referring to the example repository and might show how to do solution X, but suddenly in the context of doing something completely different like using spark instead of spring webflux etc.

I've been coding graphql for years now in various languges (elixir, nodejs etc) and decided to take the leap back into the JVM after a six year hiatus because kotlin tickled my fancy. I know most of the traditional internals that spring use, but it's still a bit unclear what prescriptions you went for, in a library that is pretty prescriptive. 

**Describe the solution you'd like**
I'd like the documentation extended with the following basic use-cases
- Data loader as a first class citizen in the documentation. The mixing of dataloader into the spark example makes it very confusing how you are supposed to use and wire this in the 'normal' reflection based resolvers that your main documentation rely on.
- The **very common** scenario of interacting with the request/response objects. I have been trying all sorts of magic to get at it in the spring webflux default way of using graphql-kotlin and I still have no clue how to get at it. This is most likely a factor of me having not touched modern Spring Boot all that much, but graphql-kotlin as a pretty clean 'standalone' kotlin library for graphql that has pretty noob-friendly documentation all around, this is a pretty glaring hole if you need to do anything that revolves around headers, cookies etc. The current documentation  for making your own context sort of let me build my own internal context, and in that factory function I can access the request and response, but I am highly doubtful embedding the request and response into the custom context is the right way to go at it? (If that is the right way, it should be explicitly documented and not included as a trivial example I think :) ) Also, the text "For more details see the Contextual Data documentation." links to a dead url: https://expediagroup.github.io/graphql-kotlin/docs/execution/contextual-data
- Tangential to the one above, there is nothing about how to build security contexts or other request scoped beans to handle user authentication in your current documentation. If the user does not know how to use spring boot with webflux, there is no real path to the goal.

olivermt:
Also forgot to add, I'd be happy to contribute this if I got some guidance on best ways to go at things :)

smyrick:
@olivermt Thank you for pointing out these gaps in our documentation. We have been adding documentation pages for features as we add them, but as you pointed out, I think it is time we do a clean pass at all the pages and make sure the docs work well together.

[...]

Issue #682, opened 2020-04-30
[documentation] update all Gradle example documentation to use Kotlin script


dariuszkuc:
We should update our documentation Gradle examples to use Kotlin script syntax (instead of Groovy). This would make it consistent with the new Gradle plugin documentation.

dariuszkuc:
Looks like it is resolved by https://github.com/ExpediaGroup/graphql-kotlin/pull/773

PR #773, opened 2020-06-22
[docs] Add server http req/res page

smyrick:
### :pencil: Description
This is by no-means a complete update of the docs, but this at least creates some basic pages that we can continue to add content to.

### :link: Related Issues
See https://github.com/ExpediaGroup/graphql-kotlin/issues/765

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Visit our [documentation site](https://expediagroup.github.io/graphql-kotlin) fo
 * [graphql-kotlin-schema-generator](/graphql-kotlin-schema-generator) - Code only GraphQL schema generation for Kotlin
 * [graphql-kotlin-spring-server](/graphql-kotlin-spring-server) - Spring Boot auto-configuration library to create a GraphQL server
 * [graphql-kotlin-types](/graphql-kotlin-types) - Core types used by both client and server
-* [plugins](/plugins) - Gradle and Maven client plugins
+* [plugins](/plugins) - Gradle and Maven plugins
 
 ## ⌨️ Usage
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -5,58 +5,37 @@
...
-GraphQL Kotlin is a collection of libraries, built on top of [graphql-java](https://www.graphql-java.com/), that aim to simplify running GraphQL in Kotlin
+GraphQL Kotlin is a collection of libraries, built on top of [graphql-java](https://www.graphql-java.com/), that aim to simplify running GraphQL clients and servers in Kotlin.
+
+Visit our [documentation site](https://expediagroup.github.io/graphql-kotlin) for more details.
 
 ## 📦 Modules
 
-* [examples](/examples) - Example apps that use graphql-kotlin libraries to test and demonstrate usages
 * [graphql-kotlin-client](/graphql-kotlin-client) - Lightweight GraphQL Kotlin HTTP client
+* [plugins](/plugins) - Gradle and Maven client plugins
```
dariuszkuc, line 15: we kept those in alphabetical order -> why change it now?

smyrick: I thought I would put then in order of "importance" or group them together. I guess it just seems weird that the first one we list is the examples and not the server or client module

dariuszkuc:
Mind updating new example gradle scripts to kts format? (see: https://github.com/ExpediaGroup/graphql-kotlin/issues/682)

docs/spring-server/http-request-response.md
```diff
@@ -0,0 +1,18 @@
...
+
+That being said, there are common use cases like authentication and authorization that require inspecting HTTP headers.
+
+## GraphQL Context
+
+Using the [GraphQLContextFactory](./spring-graphql-context.md) you can access the request and response and add information into the GraphQL context on every operation which can then be used in your schema functions.
+
+## Query Handler
+
+The library uses the [QueryHandler](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/QueryHandler.kt) interface to execute the `GraphQLRequest`. While you do not have access to the HTTP information here, you can modify the [default query handler bean](./spring-beans.md), `SimpleQueryHandler` and create the request with custom logic.
```
dariuszkuc, line 18: query handler does not have access to the raw http request/response objects - you would still need to populate those values either through reactor context or graphql context.

README.md
dariuszkuc: if you prefer we can list all `graphql-kotlin*` modules first, followed by plugins and examples. We could also link to individual plugins (or have nested list to show them).

smyrick: I kept the order as alphabetical as that will match the directory structure too

README.md
```diff
@@ -15,48 +17,25 @@ GraphQL Kotlin is a collection of libraries, built on top of [graphql-java](http
 * [graphql-kotlin-schema-generator](/graphql-kotlin-schema-generator) - Code only GraphQL schema generation for Kotlin
 * [graphql-kotlin-spring-server](/graphql-kotlin-spring-server) - Spring Boot auto-configuration library to create a GraphQL server
 * [graphql-kotlin-types](/graphql-kotlin-types) - Core types used by both client and server
-* [plugins](/plugins) - GraphQL Kotlin Gradle and Maven plugins
+* [plugins](/plugins) - Gradle and Maven client plugins
```
dariuszkuc, line 20:
```suggestion
* [plugins](/plugins) - Gradle and Maven plugins
```

docs/federated/apollo-federation.md
```diff
@@ -9,9 +9,80 @@ feasible making it much harder to manage and leading to unnecessary bottlenecks.
...
+  <groupId>com.expediagroup</groupId>
+  <artifactId>graphql-kotlin-federation</artifactId>
+  <version>${latestVersion}</version>
+</dependency>
+```
+
+With Gradle:
+
+```kotlin
+implementation("com.expediagroup", "graphql-kotlin-federation", latestVersion)
```
dariuszkuc, line 35: within our build files we just use single string - we probably should follow same pattern? guess it doesn't matter

smyrick: I like the separated format to more clearly see the group, rather than looking for the `:` inside a string

docs/spring-server/http-request-response.md
```diff
@@ -0,0 +1,14 @@
+---
+id: http-request-response
+title: Access the HTTP Request-Response
+---
+
+Sometimes you may need access to the raw HTTP request and response object.
```
dariuszkuc, line 6: maybe we can add some additional context, something along the lines

```suggestion
GraphQL is strongly typed and any data that is not part of the schema is no longer automatically known by the clients. Relying on this information becomes an "undocumented" part of your API. As a result, by default, GraphQL query resolvers do not have access to the raw HTTP request and response objects.
```

docs/spring-server/http-request-response.md
```diff
@@ -0,0 +1,14 @@
...
+There are multiple ways and locations you can access this information, however note that GraphQL has different paradigms than traditional REST APIs.
+Any data that is not part of the schema is no longer known by clients for free and it makes communication more difficult as this becomes an "undocumented" part of your API.
+
+That being said, there are common use cases like authentication and authorization that require inspecting HTTP headers.
```
dariuszkuc, line 10:
```suggestion
That being said, there are some common use cases (like authorization) that require inspecting HTTP headers.
```

docs/spring-server/http-request-response.md
```diff
@@ -0,0 +1,14 @@
...
+
+## GraphQL Context
+
+Using the [GraphQLContextFactory](./spring-graphql-context.md) you can access the request and response and add information into the GraphQL context on every operation which can then be used in your schema functions.
```
dariuszkuc, line 14: maybe we should add a link to https://expediagroup.github.io/graphql-kotlin/docs/schema-generator/execution/contextual-data and then reference the spring implementation?

smyrick: If you are looking for the HTTP headers, you will be in the spring server library, so looking at the docs on how to generally use the `GraphQL Context` in the schema generator isn't that helpful. In the page I liked to we reference the schema generator docs.

docs/spring-server/http-request-response.md
```diff
@@ -0,0 +1,14 @@
+---
+id: http-request-response
+title: Access the HTTP Request-Response
+---
+
+GraphQL is strongly typed and any data that is not part of the schema is no longer automatically known by the clients. Relying on this information becomes an "undocumented" part of your API. As a result, by default, GraphQL query resolvers do not have access to the raw HTTP request and response objects.
+There are multiple ways and locations you can access this information, however note that GraphQL has different paradigms than traditional REST APIs.
+Any data that is not part of the schema is no longer known by clients for free and it makes communication more difficult as this becomes an "undocumented" part of your API.
```
dariuszkuc, line 8: this sentence is already captured above

docs/spring-server/http-request-response.md
dariuszkuc: Lets rephrase it then to something like

```
The most common way to access the raw HTTP request and response objects is to process them when creating the `GraphQLContext` through [GraphQLContextFactory](./spring-graphql-context.md). Using the factory you can then extract the information from the incoming request and store it in the context so they can be accessed from any resolver.
```

Issue #783, opened 2020-06-30
Support to see if user set field explicitly to `null` on input object


natashashams:
**Context**
I am trying to create a mutation to allow partial updates an object with many nested fields.
`fun updatePersonInfo(request: UpdatePersonInfo): Profile?`
nested fields are all optional, so the user only needs to supply fields they want to update in the request.

Fields absent from request will be not be changed.
Fields present in the request will be set to the newly values defined in request, including fields set to `null`.

Example --
a mutation to set familyName to "natasha-new-family-name" and middleName to `null` would look like:
```
mutation myMutation {
  updatePersonInfo(
    userId: 123,
    personInfo {
      “name”: {
        “familyName”: “natasha-new-family-name”, 
        “middleName”: null
      }
    }
  )
}
```

a mutation to only update family name, leaving middleName unchanged, would look like:
```
mutation myMutation {
  updatePersonInfo(
    userId: 123,
    personInfo {
      “name”: {
        “familyName”: “natasha-new-family-name”
      }
    }
  )
}
```

**Problem**
I am unable to cleanly use graphql-kotlin to see if a user has set a field to _null_  in the mutation input object (indicating desire to delete a field) because there is no way currently to differentiate if a field was set to null by the user, or if it was simply never defined (absent in request).

This is also nature of kotlin itself, undefined fields default to `null` as the value. However, in graphql there is a difference between undefined and set to `null` that I am hoping to leverage. https://graphql.github.io/graphql-spec/June2018/#sec-Input-Objects (see: Input Coercion)

**Proposed Solution**
A wrapper class built-in to graphql-kotlin such that a field can have three values to represent it (undefined, null, value). This wrapper would not show up in the schema definition of the object, but would be used to represent if the object it wraps was present in the request or not. If it was present, it's value is also defined.

Alternatively, leverage java optional wrapper to do this, where the optional wrapper would also not show up in the schema type.

**Alternatives**
Since I do not want the wrapper shown in the schema, my current solution is to parse through the available graphql-java DataFetchingEnvironment instead of using the input object itself (input object still defined for user input) to see if fields are present in the request or explicitly set to null (indicates desire to delete).

As a temporary solution, I wanted to create this wrapper myself, however it would show up in schema. Hoping a built-in optional wrapper supported by graphql-kotlin could help this...

**Additional context**
A similar question has already been asked in this thread:
https://github.com/ExpediaGroup/graphql-kotlin/issues/50#issuecomment-483432882
[...]

PR #785, opened 2020-06-30
[example] Add example of undefined input vs null

smyrick:
### :pencil: Description
Add example code to show how to check the difference between a client not sending a value or explicitly passing in null

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/783

### DIFF
--- a/docs/schema-generator/execution/optional-undefined-arguments.md
+++ b/docs/schema-generator/execution/optional-undefined-arguments.md
@@ -1,0 +1,36 @@
+---
+id: optional-undefined-arguments
+title: Optional Undefined Arguments
+---
+
+In GraphQL, input types can be optional when you mark them as nullable in Kotlin. This means that the client can either:
+
+* Not specify a value at all
+* Send null explictly
+* Send the non-null type
+
+```kotlin
+fun optionalInput(value: String?): String? = value
+```
+
+```graphql
+query OptionalInputQuery {
+  undefined: optionalInput
+  null: optionalInput(value: null)
+  foo: optionalInput(value: "foo")
+}
+```
+
+By default if an optional input value is not specified, then the execution engine will set the argument in Kotlin to `null`.
+This means that you can not tell, by just the value alone, whether the request did not contain any argument or the client explicitly passed in `null`.
+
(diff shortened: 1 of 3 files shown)


docs/schema-generator/execution/optional-undefined-arguments.md
```diff
@@ -0,0 +1,36 @@
+---
+id: optional-undefined-arguments
+title: Optional Undefined Arguments
+---
+
+In GraphQL, input types can be optional when you mark them as nullable in Kotlin. This means that the client can either:
```
dariuszkuc, line 6:
```suggestion
In GraphQL, input types can be optional which means that the client can either:
```

docs/schema-generator/execution/optional-undefined-arguments.md
```diff
@@ -0,0 +1,36 @@
...
+
+* Not specify a value at all
+* Send null explictly
+* Send the non-null type
+
```
dariuszkuc, line 11:
```suggestion

Optional input types are represented as nullable parameters in Kotlin
```

docs/schema-generator/execution/optional-undefined-arguments.md
```diff
@@ -0,0 +1,36 @@
...
+```graphql
+query OptionalInputQuery {
+  undefined: optionalInput
+  null: optionalInput(value: null)
+  foo: optionalInput(value: "foo")
+}
+```
+
+By default if an optional input value is not specified, then the execution engine will set the argument in Kotlin to `null`.
```
dariuszkuc, line 24:
```suggestion
By default, if an optional input value is not specified, then the execution engine will set the argument in Kotlin to `null`.
```

docs/schema-generator/execution/optional-undefined-arguments.md
```diff
@@ -0,0 +1,36 @@
...
+This means that you can not tell, by just the value alone, whether the request did not contain any argument or the client explicitly passed in `null`.
+
+Instead you should inspect the [DataFetchingEnvironment](./data-fetching-environment.md) where you can see if the request had the variable defined and even check parent arguments as well.
```
dariuszkuc, line 27:
```suggestion
Instead, you should inspect the [DataFetchingEnvironment](./data-fetching-environment.md) where you can see if the request had the variable defined and even check parent arguments as well.
```

PR #797, opened 2020-07-08
Include csv and html reports for jacoco

smyrick:
### :pencil: Description
Codecov is still having issues. It reads the xml files in the build logs and is able to make a push to codecov.io and returns a link for the report, but the report is empty. Maybe it can also read csv file reports? I am shooting in the dark here. If this does not work we could just remove codecov all together since all we are really using it for is the badge since we have coverage limits set in the jacoco plugin

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -92,7 +92,8 @@ subprojects {
         jacocoTestReport {
             reports {
                 xml.isEnabled = true
-                csv.isEnabled = false
+                csv.isEnabled = true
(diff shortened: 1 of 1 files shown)


dariuszkuc:
It looks like codecov only uses xml results so don't think this will make any difference. We definitely shouldn't generate CSV report -> we should generate XML for codecov and HTML for devs.

My guess is that codecov expects the report file under some specific directory (see https://github.com/codecov/example-gradle)

PR #808, opened 2020-07-24
[generator] built-in support for optional input

dariuszkuc:
### :pencil: Description

In the GraphQL world, input types can be optional which means that the client can either:

* Not specify a value at all
* Send null explicitly
* Send the non-null type

This is in contrast with the JVM world where objects can either have some specific value or don't have any value (i.e. are `null`). As a result, when using default serialization logic it is not possible to distinguish between missing/unspecified value and explicit `null` value.

This PR introduces new `OptionalInput` sealed class wrapper that when used for query arguments will be automatically deserialized to `Undefined` object if argument was not specified OR to `Defined` wrapper class when argument was specified (including explicit null).

### :link: Related Issues
Resolves: #783

### DIFF
--- a/docs/schema-generator/execution/optional-undefined-arguments.md
+++ b/docs/schema-generator/execution/optional-undefined-arguments.md
@@ -3,13 +3,44 @@ id: optional-undefined-arguments
 title: Optional Undefined Arguments
 ---
 
-In GraphQL, input types can be optional which means that the client can either:
+In the GraphQL world, input types can be optional which means that the client can either:
 
 * Not specify a value at all
-* Send null explictly
+* Send null explicitly
 * Send the non-null type
 
-Optional input types are represented as nullable parameters in Kotlin
+This is in contrast with the JVM world where objects can either have some specific value or don't have any value (i.e.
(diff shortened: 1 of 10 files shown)


docs/schema-generator/execution/optional-undefined-arguments.md
```diff
@@ -3,13 +3,44 @@ id: optional-undefined-arguments
...
+## Using OptionalInput wrapper
+
+`OptionalInput` sealed class is a convenient wrapper that provides easy distinction between unspecified, `null` and non-null
+value. If target argument is not specified in the request it will be represented as `Undefined` object, otherwise actual
+value will be wrapped in `Defined` class.
+
+```kotlin
+fun optionalInput(input: OptionalInput<String>): String = when (input) {
+    is OptionalInput.Undefined -> "input was not specified"
+    is OptionalInput.Defined<String> -> "input value: ${input.value}"
```
smyrick, line 25:
```suggestion
fun optionalInput(input: OptionalInput<String?>): String = when (input) {
    is OptionalInput.Undefined -> "input was not specified"
    is OptionalInput.Defined<String?> -> "input value: ${input.value}"
```

PR #813, opened 2020-07-26
[server] spring aware data fetcher

dariuszkuc:
### :pencil: Description

Create new `SpringDataFetcher` that automatically resolves Spring `@Autowired` beans that are specified as function arguments.

```kotlin
class SpringQuery : Query {
    fun getWidget(@GraphQLIgnore @Autowired repository: WidgetRepository, id: Int): Widget = repository.findWidget(id)
}
```

NOTE: `@Autowired` arguments should be explicitly excluded from the GraphQL schema by also specifying `@GraphQLIgnore`.


docs/spring-server/spring-schema.md
```diff
@@ -0,0 +1,115 @@
...
+
+type Mutation {
+  myAwesomeMutation(widget: Widget!): Widget!
+}
+
+type Subscription {
+  myAwesomeSubscription: Widget!
+}
+
+type Widget {
```
smyrick, line 50: There would also be an `input Widget`

docs/spring-server/spring-schema.md
```diff
@@ -0,0 +1,115 @@
...
+  id: Int!
+  value: String!
+}
```
smyrick, line 53:
```suggestion
}

input Widget {
  id: Int!
  value: String!
}
```

dariuszkuc: Nice catch 👍 there would be `Widget` and `WidgetInput` will fix it (copied it from intro page).

Issue #816, opened 2020-07-28
Add support for ScalarConverter from other types than String


stengvac:
Hello,

Using gradle plugin `id("com.expediagroup.graphql") version "3.4.2"` and  `implementation("com.expediagroup:graphql-kotlin-client:3.4.2")`, kotlin 1.3.72, jackson 2.11.0

Our graphql schema contains types UUID, Long, BigDecimal. UUID, Long can be handled as is shown in [example](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/client/gradle-client/src/main/kotlin/com/expediagroup/graphql/examples/client/UUIDScalarConverter.kt). But this approach does not work for me in case of BigDecimal (I don`t get why it works for Long).

Converters
```
import java.math.BigDecimal

class LongScalarConverter : ScalarConverter<Long> {
    override fun toScalar(rawValue: String): Long = rawValue.toLong()
    override fun toJson(value: Long): String = value.toString()
}

class BigDecimalScalarConverter : ScalarConverter<BigDecimal> {
    override fun toScalar(rawValue: String): BigDecimal = BigDecimal(rawValue)
    override fun toJson(value: BigDecimal): String = value.toString()
}
```
Generated classes for those registered converters

```
 /**
   * Long
   */
  data class Long(
    val value: kotlin.Long
  ) {
    @JsonValue
    fun rawValue() = converter.toJson(value)

    companion object {
      val converter: LongScalarConverter = LongScalarConverter()

      @JsonCreator
      @JvmStatic
      fun create(rawValue: String) = Long(converter.toScalar(rawValue))
    }
  }

  /**
   * BigDecimal
   */
  data class BigDecimal(
    val value: java.math.BigDecimal
  ) {
    @JsonValue
    fun rawValue() = converter.toJson(value)

    companion object {
      val converter: BigDecimalScalarConverter = BigDecimalScalarConverter()

      @JsonCreator
      @JvmStatic
      fun create(rawValue: String) = BigDecimal(converter.toScalar(rawValue))
    }
  }
```
fail with exception 

```
 val objectMapper = jacksonObjectMapper()
    val inJson = """
        0
    """.trimIndent()
//works
    val longRes = objectMapper.readValue(inJson, CreateCardMutation.Long::class.java)
//works
    val javaBigDecimal = objectMapper.readValue(inJson, BigDecimal::class.java)
    // this one throw ex
[...]

PR #819, opened 2020-07-30
[client] update ScalarConverter to accept any objects

dariuszkuc:
### :pencil: Description

Custom GraphQL scalars can be serialized as any of the other built-in scalars (i.e. int, float, etc). Changed method signatures to accept `Any` when converting to scalar, and produce `Any` when generating JSON value.

### :link: Related Issues

Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/816


dariuszkuc:
Verified locally with `examples/client` projects - will update those once we release a version.

PR #821, opened 2020-07-31
[generator] Do not add interfaces for input additional types

smyrick:
### :pencil: Description
When generating the additional input types it is possible someone provided an interface or union and mark `inputType` as true.  The generator handles this by just creating an interface as expected but if that interface was also added before as "output" type then we have a duplicate. In this case `graphql-java` will throw an error saying that the type is being redeclared. So before we generate the additional types we should just verify is is a valid type

### :link: Related Issues
Bug fix from the feature added here: https://github.com/ExpediaGroup/graphql-kotlin/pull/817

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -19,6 +19,7 @@ package com.expediagroup.graphql.generator
 import com.expediagroup.graphql.SchemaGeneratorConfig
 import com.expediagroup.graphql.TopLevelObject
 import com.expediagroup.graphql.exceptions.InvalidPackagesException
+import com.expediagroup.graphql.generator.extensions.distinctName
 import com.expediagroup.graphql.generator.state.AdditionalType
 import com.expediagroup.graphql.generator.state.ClassScanner
 import com.expediagroup.graphql.generator.state.TypesCache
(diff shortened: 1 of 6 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -117,10 +119,15 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) : Closeab
         while (this.additionalTypes.isNotEmpty()) {
             val currentlyProcessedTypes = LinkedHashSet(this.additionalTypes)
             this.additionalTypes.clear()
-            graphqlTypes.addAll(currentlyProcessedTypes.map { generateGraphQLType(this, it.kType, it.inputType) })
+            graphqlTypes.addAll(
+                currentlyProcessedTypes.map {
```
dariuszkuc, line 123: i thought our caching would catch duplicates? e.g. it is possible same type could be attempted to be added multiple times

smyrick: The cache does return the value but problem is that now the `additionalTypes` would have two of the type so if you try to add that to the schema builder and then call `.build()` it throws an error

ie

```kotlin
@MyAnnotation
interface Foo

generator. addAdditionalTypesWithAnnotation(MyAnnotation::class, inputType = false)
// 1 additional type 

generator. addAdditionalTypesWithAnnotation(MyAnnotation::class, inputType = true)
// 2 additional types
```

But since it is an interface is is not added as input type so there is a duplicate

smyrick: After further discussion, the issue is with addAdditionalTypesWithAnnotation so I needed to add checks there. This is much better

Issue #818, opened 2020-07-30
Simplify generateSchema signature 


smyrick:
**Describe the solution you'd like**
The `generateSchema` has to arguments: `additionalTypes` and `additionalInputTypes`. If we change the signature and use the new public class `AdditionalType` we can make this into one argument

**Describe alternatives you've considered**
We could keep the split arguments and make the `AdditionalType` class internal only

**Additional context**
This change was added here to make it backwards compatible but with a major version we can introduce a breaking change

https://github.com/ExpediaGroup/graphql-kotlin/pull/817

smyrick:
Another option is to move these initial arguments to the `SchemaGeneratorConfig` to keep the `generateSchema` method simple

miguelhrocha:
Hey @smyrick I created a PR for this: https://github.com/ExpediaGroup/graphql-kotlin/pull/825

smyrick:
See comments here: https://github.com/ExpediaGroup/graphql-kotlin/pull/825#issuecomment-670200690

PR #825, opened 2020-08-01
[generator] Simplify generateSchema signature

miguelhrocha:
### :pencil: Description
Simplify SchemaGenerator#generateSchema signature to just 1 set of additionalTypes

The generateSchema has to arguments: additionalTypes and additionalInputTypes. If we change the signature and use the new public class AdditionalType we can make this into one argument

### :link: Related Issues
[#818](https://github.com/ExpediaGroup/graphql-kotlin/issues/818)

### DIFF
--- a/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGenerator.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGenerator.kt
@@ -19,8 +19,8 @@ package com.expediagroup.graphql.federation
 import com.expediagroup.graphql.TopLevelObject
 import com.expediagroup.graphql.federation.directives.ExtendsDirective
 import com.expediagroup.graphql.generator.SchemaGenerator
+import com.expediagroup.graphql.generator.state.AdditionalType
 import graphql.schema.GraphQLSchema
-import kotlin.reflect.KType
 
 /**
  * Generates federated GraphQL schemas based on the specified configuration.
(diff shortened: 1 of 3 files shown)


smyrick:
@miguelhrocha Thanks for raising this PR! This will be a breaking change so we are going to wait until we are ready to make the 4.x cut until we will merge

dancetrain:
I'm sorry for asking for this, but what is the main reason of this change? How this solution will handle different types and inputs?

Small note:
GraphQL has types and inputs, and they are different and have different meaning.

Types are used for model description in queries, while inputs are used as parameters for migrations, you can't mess them.

Example:
```
schema {
    query: Query
    mutation: Mutation
}

type ShoppingListItem {
    id: Int!
    desc: String!
    priority: Int!
}

input ShoppingListItemInput {
    desc: String!
    priority: Int!
}

type Query {
    getShoppingListItems: [ShoppingListItem!]!
}

type Mutation {
    addShoppingListItem(item: ShoppingListItemInput!): ShoppingListItem!
    deleteShoppingListItem(id: Int!): Int
}
```

As you can see in example above `ShoppingListItem` has non-nullable field id, which we shouldn't pass from client to server

From documentation:
The fields on an input object type can themselves refer to input object types, but you can't mix input and output types in your schema. Input object types also can't have arguments on their fields.

Further reading:
https://graphql.org/learn/schema/#input-types

smyrick:
@dancetrain I think you bring up a good point actually. I had created the initial issue because having the single `additionalType: Set<AdditionalType` internally make the logic much easier when we have to loop over the types during generation as more additional types can be added.

However to schema developers `OBJECT` and `INPUT_OBJECT` are two very distinct things and maybe we should keep them separate.

@miguelhrocha Did you implement this PR because the issue was just created or do you think that they should be combined?

miguelhrocha:
Hey @smyrick I have to admit that I created PR because the issue was just created, I am still learning about GraphQL. But what @dancetrain says makes a lot of sense to me

smyrick:
I talked with some of the other maintainers and we are going to keep these as separate arguments. It makes it clear that this are different types and there is no need to expose another public class if we don't need to.

Thanks for the PR @miguelhrocha, if you would like to help contribute on other issues feel free to scan the other open ones

PR #826, opened 2020-08-01
Add new hook to override the isValidAdditionalType check

smyrick:
### :pencil: Description
Add new hooks setting that will allow you to override the `isValidAdditionalType` check in the `addAdditionalTypesWithAnnotation` method and when adding additional types from the `generateInterface` implementations.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/818

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateInterface.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateInterface.kt
@@ -16,7 +16,6 @@
 
 package com.expediagroup.graphql.generator.types
 
-import com.expediagroup.graphql.generator.state.AdditionalType
 import com.expediagroup.graphql.extensions.unwrapType
 import com.expediagroup.graphql.generator.SchemaGenerator
 import com.expediagroup.graphql.generator.extensions.getGraphQLDescription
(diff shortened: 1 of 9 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateInterface.kt
```diff
@@ -58,7 +57,7 @@ internal fun generateInterface(generator: SchemaGenerator, kClass: KClass<*>): G
         .forEach { builder.field(generateFunction(generator, it, kClass.getSimpleName(), null, abstract = true)) }
 
     generator.classScanner.getSubTypesOf(kClass)
-        .filter { it.isGraphQLIgnored().not() }
+        .filter { generator.config.hooks.isValidAdditionalType(it, inputType = false) }
```
dariuszkuc, line 60: this seems weird calling it from inteface generator

smyrick: It is another location in where we add to the `additionalTypes` so I don't see why we should exclude it. 

Also we need that to work for “input only” types as types could be added from interfaces and the implementations for input only but the interface is ignored

PR #831, opened 2020-08-07
Documentation: Fix typos and a broken link

Inego:

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ See more info in [CONTRIBUTING.md](CONTRIBUTING.md)
 
 ## 🛡️ Security
 
-For more info on how to contact the team for security issues or the supported versions that recieve security updates, see [SECURITY.md](./.github/SECURITY.md)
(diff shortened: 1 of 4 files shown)


Inego:
> can you remove the `.md` suffixes from the links?

Oh, the problem seems to be worse than I thought. These pages can be navigated both on expediagroup.github.io/graphql-kotlin/docs and here on GitHub. And GitHub doesn't support navigation to pages without the `.md` extension (since that's the exact file name), while the generated help site has the problem the other way.

I honestly don't even know what to do :(

Inego:
For example, if you open this page on GitHub:

https://github.com/ExpediaGroup/graphql-kotlin/blob/master/docs/schema-generator/execution/contextual-data.md

and try to click the link from there:
image

... you get a 404.

Probably it's just not supposed to be used like that since these files are only source files for an external docs site.

Inego:
Rolled back .md extensions and fixed a couple of typos in other places on the way.

smyrick:
Merged as the library checks are not needed. Yea we don't need to be able to navigate from the source code of markdown docs, this is all about the generated links in the docs site.

Thanks for the contributions!

PR #844, opened 2020-08-19
Update to kotlin 1.4

smyrick:
### :pencil: Description
This is a simple update of the version and does not yet truly take advantage of any of the newer features but it looks like with just a couple warning fixes we can update as a backwards compatible change. We are also not using any `1.4` specific features so this should be able to be used in `1.3.x` apps

### :link: Related Issues
https://kotlinlang.org/docs/reference/whatsnew14.html
https://github.com/Kotlin/kotlinx.coroutines/releases

### DIFF
--- a/examples/client/maven-client/pom.xml
+++ b/examples/client/maven-client/pom.xml
@@ -10,8 +10,8 @@
         <!-- those properties are set by gradle -->
         <graphql-kotlin.version>3.1.0</graphql-kotlin.version>
         <!-- lib versions -->
-        <kotlin.version>1.3.72</kotlin.version>
-        <kotlin-coroutines.version>1.3.6</kotlin-coroutines.version>
+        <kotlin.version>1.4.0</kotlin.version>
+        <kotlin-coroutines.version>1.3.8</kotlin-coroutines.version>
         <ktor.version>1.3.1</ktor.version>
     </properties>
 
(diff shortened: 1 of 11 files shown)


dariuszkuc:
Lets hold-off with this until we start merging 4.x.x changes

examples/client/maven-client/pom.xml
```diff
@@ -10,8 +10,8 @@
         <!-- those properties are set by gradle -->
         <graphql-kotlin.version>3.1.0</graphql-kotlin.version>
         <!-- lib versions -->
-        <kotlin.version>1.3.72</kotlin.version>
-        <kotlin-coroutines.version>1.3.6</kotlin-coroutines.version>
+        <kotlin.version>1.4.0</kotlin.version>
+        <kotlin-coroutines.version>1.3.8</kotlin-coroutines.version>
```
dariuszkuc, line 14: I think we probably use `1.3.9` for coroutines (or wait for official `1.4.0`)

plugins/graphql-kotlin-gradle-plugin/src/test/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePluginAbstractIT.kt
```diff
@@ -35,7 +35,7 @@ abstract class GraphQLGradlePluginAbstractIT {
     // unsure if there is a better way - correct values are set from Gradle build
     // when running directly from IDE you will need to manually update those to correct values
     private val gqlKotlinVersion = System.getProperty("graphQLKotlinVersion") ?: "3.0.0-SNAPSHOT"
```
dariuszkuc, line 37: can you bump it to 4.0.0-SNAPSHOT?

Issue #835, opened 2020-08-11
Ambiguous 500 server error when trying to run subscription


hawksdoves:
I am using your wonderful graphql-spring-server library and have it up and working for my queries and mutations (tested manually in my local playground) however, I am struggling implementing a simple subscription. When I try and run a subscription within the playground

eg

```

subscription {
  counter
}
```
as defined

```
@Component
class MySubscriptions: Subscription {
    val logger = LoggerFactory.getLogger(MySubscriptions::class.java)

    fun counter(): Flux<Int> = Flux.interval(Duration.ofSeconds(1)).map {
        val value = Random.nextInt()
        logger.info("Returning $value from counter")
        value
    }
}
```
In the playground I get error ``"Could not connect to websocket endpoint ws://localhost:15015/subscriptions. Please check if the endpoint url is correct."`` and on the server I get:

```
ERROR o.s.b.a.w.r.e.AbstractErrorWebExceptionHandler - [5fde5914]  500 Server Error for HTTP GET "/subscriptions"
java.lang.IllegalStateException: No WebSocketServerFactory available
               at org.springframework.util.Assert.state(Assert.java:73)
               Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException:
Error has been observed at the following site(s):
               |_ checkpoint ⇢ com.expediagroup.graphql.spring.execution.ContextWebFilter [DefaultWebFilterChain]
               |_ checkpoint ⇢ HTTP GET "/subscriptions" [ExceptionHandlingWebHandler]
Stack trace:
                              at org.springframework.util.Assert.state(Assert.java:73)
                              at org.springframework.web.reactive.socket.server.upgrade.JettyRequestUpgradeStrategy.upgrade(JettyRequestUpgradeStrategy.java:160)
```
I was hoping you might have experienced this before and could point me in the right direction on how to resolve this.

Thanks so much

smyrick:
@hawksdoves Can you see if version https://github.com/ExpediaGroup/graphql-kotlin/releases/tag/4.0.0-alpha.1

helps resolve your issue? It includes this fix: https://github.com/ExpediaGroup/graphql-kotlin/pull/845

PR #845, opened 2020-08-23
Fix subscriptions messaging and tests

smyrick:
### :pencil: Description
When updating to the latest spring boot version our subscription tests are failing. I can reproduce this with just the jump to Spring Boot `2.2.7`. 

The issue is that the `StepVerifier`s do not complete the publisher so the test stays running in the background and never completes. I have re-worked how to capture the published events in the tests, as well as fixed a bug in the subscription code that was not sending the `COMPLETE` event. However this still hangs when we are trying to process multiple events. 

The [line here](https://github.com/ExpediaGroup/graphql-kotlin/compare/master...smyrick:subscriptions-tests?expand=1#diff-d5e812810c739e39574a5826d45b46a8R86), using `toMono` stop the publisher after just one event which is why it fails when attempting to check the second message. However after using `toFlux` the test never completes. If I manually stop it the test has actually passed as all the data was present and completed, but the test is still waiting. 

If someone else can help investigate or knows the Reactor methods better than I do, it would be very helpful to figure what we have to do in these integration tests. The code outside integration tests looks good.

Screen Shot 2020-08-22 at 7 19 22 PM

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/835

### DIFF
--- a/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/dataloaders/CompanyService.kt
+++ b/examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/dataloaders/CompanyService.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2020 Expedia, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.examples.dataloaders
+
+import com.expediagroup.graphql.examples.model.Company
+import org.springframework.stereotype.Component
+
+@Component("companyServiceBean")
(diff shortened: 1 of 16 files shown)


gradle.properties
```diff
@@ -25,8 +25,8 @@ kotlinPoetVersion = 1.6.0
 ktorVersion = 1.3.1
 reactorVersion = 3.3.6.RELEASE
 reactorExtensionsVersion = 1.0.2.RELEASE
-springBootVersion = 2.2.6.RELEASE
-springVersion = 5.2.5.RELEASE
+springBootVersion = 2.2.7.RELEASE
```
dariuszkuc, line 28: if we are bumping deps maybe we should update to `2.2.9`

examples/spring/src/main/kotlin/com/expediagroup/graphql/examples/dataloaders/CompanyService.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.examples.dataloaders
+
+import com.expediagroup.graphql.examples.model.Company
+import org.springframework.stereotype.Component
+
+@Component("companyServiceBean")
```
dariuszkuc, line 22: whats the point of renaming from default name to `companyServiceBean` (i.e. adding bean suffix)?

smyrick: The application fails to start the examples if this is not included

PR #846, opened 2020-08-25
Add test to verify interface implementations can be ignored

smyrick:
### :pencil: Description
Just adding an additional unit test for documentation and verification purposes

### :link: Related Issues
N\A

### DIFF
--- a/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/InterfaceOfInterfaceTest.kt
+++ b/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/InterfaceOfInterfaceTest.kt
@@ -17,12 +17,15 @@
 package com.expediagroup.graphql.test.integration
 
 import com.expediagroup.graphql.TopLevelObject
+import com.expediagroup.graphql.annotations.GraphQLIgnore
+import com.expediagroup.graphql.extensions.deepName
 import com.expediagroup.graphql.testSchemaConfig
 import com.expediagroup.graphql.toSchema
 import graphql.schema.GraphQLInterfaceType
 import org.junit.jupiter.api.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
+import kotlin.test.assertNull
 
 class InterfaceOfInterfaceTest {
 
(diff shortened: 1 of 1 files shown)


graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/test/integration/InterfaceOfInterfaceTest.kt
```diff
@@ -61,5 +77,12 @@ class InterfaceOfInterfaceTest {
 
     class InterfaceOfInterfaceQuery {
         fun getClass() = MyClass(id = "1", name = "fooBar")
+        fun getIgnoredClass(): SecondLevel = IgnoredClass(id = "2", name = "baz")
```
dariuszkuc, line 80: this seems weird... so we return an interface without actual implementation?

smyrick: Yes, this can be helpful for cases where maybe you have some shared interfaces.

```kotlin
interface MySharedInnerType {
  val value: String
}

interface MySharedType {
  val id: String
  val value: MySharedInnerType
}
```

So then in the code you only want the services to ever use and expose the `MySharedType` but if the API developers wanted different implementations they could create that logic but clients would still only ever see interfaces.

dariuszkuc: still seems "wrong" to expose only interfaces without implementations, wonder if something like this causes issues for the clients (it definitely would for our own client code generation)

smyrick: I can close this PR if needed. This just verifies behaviour that is possible with the generator. Whether you do it or not is up to your schema design

dariuszkuc: I understand that this is a current behavior. My questions were more about asking whether this should be allowed or not. I don't see anything in the spec that explicitly says interface must have any implementations....

smyrick: No, and when you get into the schema stitching/federation world, we may have a published schema that is only interfaces and maybe only in the final merged schemas is there implementations but even then the fields just might only be interfaces.

For clients it doesn't matter when writing queries if the types (implementations) don't expose any additional fields

smyrick: If we want to change the behaviour in the future this test will verify we have done it correctly by failing

PR #847, opened 2020-08-25
[build] officially start targeting 4.0.0

dariuszkuc:
Just bumps the version in `gradle.properties` that is used for CI. Release builds use version from tags anyway.


smyrick:
`3.x.x` branch has been created from `master` right before this commit: https://github.com/ExpediaGroup/graphql-kotlin/tree/3.x.x

PR #851, opened 2020-08-29
Change federated resolver to include type name

smyrick:
### :pencil: Description
By changing the interface of `FederatedTypeResolver` to include the type name, we no longer to to manually create the `FederatedTypeRegistry`, which means that we can create it for library users behind the scenes as its own Spring bean. This also means that we need to reorganize the `FederatedSchemaAutoConfiguration` in `graphql-kotlin-spring-server` to be able to use all the configuration properties.

The end result is that the Spring configuration classes now have more hierarchy and library users don't need to manually create the `FederatedTypeRegistry` mapping strings names to resolvers. 

We may eventually be able to infer the type name in `FederatedTypeResolver` from the generic class info, but `T::class.simpleName` returns a nullable string currently.

### :link: Related Issues
N\A


graphql-kotlin-schema-generator/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ dependencies {
     api("com.graphql-java:graphql-java:$graphQLJavaVersion")
     api("org.jetbrains.kotlinx:kotlinx-coroutines-reactive:$kotlinCoroutinesVersion")
     api("com.fasterxml.jackson.module:jackson-module-kotlin:$jacksonVersion")
-    implementation(kotlin("reflect", kotlinVersion))
+    api(kotlin("reflect", kotlinVersion))
```
smyrick, line 15: I was getting the following errors locally without changing this

```
> Task :graphql-kotlin-federation:compileTestKotlin
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    /Users/smyrick/.m2/repository/org/jetbrains/kotlin/kotlin-reflect/1.3.61/kotlin-reflect-1.3.61.jar (version 1.3)
    /Users/smyrick/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib/1.4.0/kotlin-stdlib-1.4.0.jar (version 1.4)
    /Users/smyrick/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib-common/1.4.0/kotlin-stdlib-common-1.4.0.jar (version 1.4)
w: Consider providing an explicit dependency on kotlin-reflect 1.4 to prevent strange errors
w: Some runtime JAR files in the classpath have an incompatible version. Consider removing them from the classpath
```

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/state/ClassScanner.kt
```diff
@@ -32,7 +32,7 @@ internal class ClassScanner(supportedPackages: List<String>) : Closeable {
     @Suppress("Detekt.SpreadOperator")
     private val scanResult = ClassGraph()
         .enableAllInfo()
-        .whitelistPackages(*supportedPackages.toTypedArray())
+        .acceptPackages(*supportedPackages.toTypedArray())
```
smyrick, line 35: This method name was deprecated and replaced with no logic changes

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+        preparsedDocumentProvider.ifPresent {
+            graphQL.preparsedDocumentProvider(it)
+        }
+        return graphQL.build()
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    fun graphQLQueryHandler(
```
smyrick, line 99: I think for v4 we should rename this to `GraphQLRequestHandler`

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/FederatedSchemaAutoConfiguration.kt
```diff
@@ -33,24 +34,29 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
...
 import java.util.Optional
 
 /**
- * SpringBoot autoconfiguration for generating Federated GraphQL schema.
+ * SpringBoot autoconfiguration for generating federated GraphQL schema object.
+ * These will override the beans in [NonFederatedSchemaAutoConfiguration] if federation is enabled.
  */
 @ConditionalOnProperty(value = ["graphql.federation.enabled"], havingValue = "true")
 @Configuration
+@Import(GraphQLExecutionConfiguration::class)
```
smyrick, line 46: I needed to create this new root `GraphQLExecutionConfiguration` so I could share the same beans in the renamed `NonFederatedSchemaAutoConfiguration`

graphql-kotlin-federation/build.gradle.kts
```diff
@@ -1,9 +1,11 @@
 description = "Federated GraphQL schema generator"
 
+val kotlinVersion: String by project
 val junitVersion: String by project
 
 dependencies {
     api(project(path = ":graphql-kotlin-schema-generator"))
+    implementation(kotlin("reflect", kotlinVersion))
```
smyrick, line 8: I was getting the following errors locally without changing this

```
> Task :graphql-kotlin-federation:compileTestKotlin
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    /Users/smyrick/.m2/repository/org/jetbrains/kotlin/kotlin-reflect/1.3.61/kotlin-reflect-1.3.61.jar (version 1.3)
    /Users/smyrick/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib/1.4.0/kotlin-stdlib-1.4.0.jar (version 1.4)
    /Users/smyrick/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib-common/1.4.0/kotlin-stdlib-common-1.4.0.jar (version 1.4)
w: Consider providing an explicit dependency on kotlin-reflect 1.4 to prevent strange errors
w: Some runtime JAR files in the classpath have an incompatible version. Consider removing them from the classpath
```

examples/federation/extend-app/src/main/kotlin/com/expediagroup/graphql/examples/Application.kt
```diff
@@ -16,18 +16,11 @@
...
-import com.expediagroup.graphql.examples.extend.widgetResolver
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
-import org.springframework.context.annotation.Bean
 
 @SpringBootApplication
-class Application {
-
-    @Bean
-    fun federatedTypeRegistry() = FederatedTypeRegistry(mapOf("Widget" to widgetResolver))
```
smyrick, line 29: Example of library user code simplifications. This can just be deleted

graphql-kotlin-spring-server/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     api("io.projectreactor.kotlin:reactor-kotlin-extensions:$reactorExtensionsVersion")
     api("org.jetbrains.kotlinx:kotlinx-coroutines-reactor:$kotlinCoroutinesVersion")
     kapt("org.springframework.boot:spring-boot-configuration-processor:$springBootVersion")
+    annotationProcessor("org.springframework.boot:spring-boot-configuration-processor:$springBootVersion")
```
dariuszkuc, line 21: i believe this is already handled by `kapt`?

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/SpringKotlinDataFetcherFactoryProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+
+package com.expediagroup.graphql.spring.execution
+
+import com.expediagroup.graphql.execution.SimpleKotlinDataFetcherFactoryProvider
+import com.fasterxml.jackson.databind.ObjectMapper
+import graphql.schema.DataFetcherFactory
+import org.springframework.context.ApplicationContext
+import kotlin.reflect.KFunction
+
+class SpringKotlinDataFetcherFactoryProvider(
```
dariuszkuc, line 25: can we add javadoc?

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederatedTypeResolver.kt
```diff
@@ -23,6 +23,12 @@ import graphql.schema.DataFetchingEnvironment
  */
 interface FederatedTypeResolver<out T> {
 
+    /**
+     * This is the GraphQL name of the type [T]. It is used when running the resolvers and inspecting the
+     * GraphQL "__typename" property during the entities requests
+     */
+    val typeName: String
```
dariuszkuc, line 30: guess this is no different than having to specify explicit `string` -> resolver mapping. Ideally this should be coming from the underlying type `T` but when I last looked at it was problematic due to type erasure.

smyrick: Yea, this just removes the needed extra step that you need to create a map of `Strings` to resolvers. I think it is more clear here that the type name should be part of the resolver so you only have to change code in one location

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederatedTypeRegistry.kt
```diff
@@ -19,10 +19,15 @@ package com.expediagroup.graphql.federation.execution
 /**
  * Simple registry that holds mapping of all registered federated GraphQL types and their corresponding resolvers.
  */
-class FederatedTypeRegistry(private val federatedTypeResolvers: Map<String, FederatedTypeResolver<*>> = emptyMap()) {
+class FederatedTypeRegistry(federatedTypeResolvers: List<FederatedTypeResolver<*>> = emptyList()) {
+
+    /**
+     * Precompute the map of names to types so that we don't have to search the list every time.
+     */
+    private val namesToResolvers: Map<String, FederatedTypeResolver<*>> = federatedTypeResolvers.map { it.typeName to it }.toMap()
```
dariuszkuc, line 27: instead of `map { }.toMap()` I think you can just use `associateBy { it.typeName }`

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/NonFederatedSchemaAutoConfiguration.kt
```diff
@@ -33,16 +33,19 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
...
 /**
- * SpringBoot autoconfiguration for generating GraphQL schema.
+ * SpringBoot autoconfiguration for generating a non-federated GraphQL schema object.
+ * This will override the beans in [FederatedSchemaAutoConfiguration] if federation is disabled.
  */
 @ConditionalOnProperty(value = ["graphql.federation.enabled"], havingValue = "false", matchIfMissing = true)
 @Configuration
-class SchemaAutoConfiguration {
+@Import(GraphQLExecutionConfiguration::class)
+class NonFederatedSchemaAutoConfiguration {
```
dariuszkuc, line 46: it seems weird to call it with `NonFederated` prefix as that would imply `Federation` is the default

smyrick: Yea, I didn't know what to call it because we have this format.

What do I call the bean above this that imports both of these expecting one of the returns a `GrapQLSchema` bean and then creates a `GraphQL` bean? 

Screen Shot 2020-08-31 at 11 43 18 AM

smyrick: We have the follow config classes and they all are very close in name

* `GraphQLAutoConfiguration`
* `GraphQLSchemaConfiguration`
* `SchemaAutoConfiguraiton`

dariuszkuc: We only had `GraphQLAutoConfiguration` (root) and `SchemaAutoConfiguration` before ;). Technically new `GraphQLSchemaConfiguration` is actually only creating `GraphQL` bean and corresponding handler. Schemas are created in those other two configs that are imported.

Wondering whether we should be combining some of those together (as they only contain few beans) but I do like the separation by conditional properties better....

smyrick: We can always rename now that we have the `4.x.x` cut

PR #852, opened 2020-08-29
Update docs on library release

smyrick:
### :pencil: Description
We should update the docs versions on every code release. We can then come back manually and delete excess version numbers, but the docs should always reflect the latest

### :link: Related Issues
Original issue here: https://github.com/ExpediaGroup/graphql-kotlin/pull/767

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,3 +71,14 @@ jobs:
           publish_dir: ./website/build/graphql-kotlin
           user_name: eg-oss-ci
           user_email: oss@expediagroup.com
+
+      - name: Commit new files
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git commit -a -m "Add new docs version"
+
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,8 +5,9 @@ on:
     types: [published]
 
 jobs:
-  release:
+  release-code:
     runs-on: ubuntu-latest
+    if: github.repository == 'ExpediaGroup/graphql-kotlin'
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -34,3 +35,39 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: Build website
+        run: |
+          NEW_VERSION=$(echo "${GITHUB_REF}" | cut -d "/" -f3)
+          echo "New version: ${NEW_VERSION}"
+          cd website
+          npm install
+          npm run version $NEW_VERSION
```
dariuszkuc, line 64: this actually creates new folder/files under `website` dir, if you want to auto publish those then I guess we need to explicitly add them to the repo? I believe the next action only pushes to `gh-pages`  branch.

smyrick: This action here does publish to the `gh-pages` branch

https://github.com/ExpediaGroup/graphql-kotlin/pull/852/files#diff-56c30d9224897b08c28075052d4595c5R68

dariuszkuc: yes action pushes to `gh-pages` but we need to push to master - update sidebars (https://github.com/ExpediaGroup/graphql-kotlin/tree/master/website/versioned_sidebars) and create versioned markdown files (https://github.com/ExpediaGroup/graphql-kotlin/tree/master/website/versioned_docs/version-3.6.1)

smyrick: Oh, right. I will update to use the following action: https://github.com/marketplace/actions/github-push?version=v0.6.0

.github/workflows/release.yml
```diff
@@ -34,3 +35,50 @@ jobs:
...
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./website/build/graphql-kotlin
+          user_name: eg-oss-ci
+          user_email: oss@expediagroup.com
+
+      - name: Commit new files
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git commit -a -m "Add new docs version"
```
dariuszkuc, line 79: instead of `-a` should we limit to some website sub directories (just in case)?

PR #853, opened 2020-08-30
Update to gradle 6.6.1

smyrick:
### :pencil: Description
Update the gradle wrapper version

### :link: Related Issues
N\A


plugins/graphql-kotlin-gradle-plugin/build.gradle.kts
```diff
@@ -14,6 +15,7 @@ dependencies {
     api("org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion")
     testImplementation("com.github.tomakehurst:wiremock-jre8:$wireMockVersion")
     testImplementation("com.github.spullara.mustache.java:compiler:$mustacheVersion")
+    testImplementation(kotlin("reflect", kotlinVersion))
```
dariuszkuc, line 18: its already included in dependencies from the root `build.gradle.kts`

build.gradle.kts
```diff
@@ -181,7 +181,8 @@ subprojects {
     }
 
     dependencies {
-        implementation(kotlin("stdlib", kotlinVersion))
+        implementation(kotlin("stdlib-jdk8", kotlinVersion))
```
dariuszkuc, line 184: IMHO client libs should depend on `stdlib` only whereas all other ones should be pulling in `jdk8` flavor. Doing this change in the root file will apply it on ALL projects.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.6.1-bin.zip
```
dariuszkuc, line 3: lets separate gradle update to another PR

PR #858, opened 2020-09-01
[build] revert Kotlin version to 1.3.72

dariuszkuc:
### :pencil: Description

We are hitting some library and plugin issues with incompatibilities between Kotlin versions. Downgrading back to 1.3.72 for now, we will revisit update to 1.4 sometime in the future.

### :link: Related Issues

Update to 1.4 https://github.com/ExpediaGroup/graphql-kotlin/pull/844

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,8 +15,8 @@ org.gradle.jvmargs=-Xmx2g -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError
 
 # dependencies
 kotlinJvmVersion = 1.8
-kotlinVersion = 1.4.0
-kotlinCoroutinesVersion = 1.3.9
+kotlinVersion = 1.3.72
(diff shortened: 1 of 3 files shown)


dariuszkuc:
I'll update the examples in subsequent PR once we release another version.

pjwalstrom:
was this due to https://github.com/FasterXML/jackson-module-kotlin/issues/356?

dariuszkuc:
@pjwalstrom We generally don't use inline classes for schema generation so that one wasn't problematic. I did encounter another issue (https://github.com/FasterXML/jackson-databind/issues/2821) but that is happening regardless of Kotlin version.

I was getting issues with Gradle including https://issuetracker.google.com/issues/166582569 and https://youtrack.jetbrains.com/issue/KT-37435 that was blowing up when using our plugin in a separate project (unsure why integration tests worked).

PR #865, opened 2020-09-03
[client] Remove duplicate header

MarkusLund:
### :pencil: Description
It looks like the application/json accept header is set by `install(feature = JsonFeature)` on line 48 as well as `accept(ContentType.Application.Json)` on line 54. This causes a duplicate Accept header during the introspection query Gradle task. 

Spring handles the duplicate Accept header, but Jersey returns an 400 Bad Request when receiving the header.`Accept: application/json;application/json`

### :link: Related Issues
I have not created an issue.


smyrick (approved):

smyrick:
@MarkusLund Thanks for the PR! Can you also make the same update to the `3.x.x` branch so we can backport this to a current release

smyrick:
Actually we also need to make the same change in the download SDL task

PR #867, opened 2020-09-03
[3.x.x][client] Remove duplicate header

smyrick:
### :pencil: Description
Copy changes over to `3.x.x` branch

It looks like the application/json accept header is set by `install(feature = JsonFeature)` on line 48 as well as `accept(ContentType.Application.Json)` on line 54. This causes a duplicate Accept header during the introspection query Gradle task. 

Spring handles the duplicate Accept header, but Jersey returns an 400 Bad Request when receiving the header.`Accept: application/json;application/json`

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/pull/865


tapaderster (approved):

dariuszkuc:
Pretty sure we will have to remove it from the client as well -> https://github.com/ExpediaGroup/graphql-kotlin/blob/3.x.x/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/GraphQLClient.kt#L85

dariuszkuc (approved):

PR #871, opened 2020-09-04
[docs] Add a comparison to graphql-java

smyrick:
### :pencil: Description
Add a basic page the shows a comparison between creating a schema in graphql-java and graphql-kotlin.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/765

### DIFF
--- a/docs/graphql-java-comparison.md
+++ b/docs/graphql-java-comparison.md
@@ -114,13 +114,13 @@ class Query {
 class Book(
     val id: ID,
     val name: String,
-    private val pageCount: Int,
+    private val totalPages: Int,
     private val authorId: ID
 ) {
     private val authors: List<Author> = authorsFromDB()
 
     fun author(): Author? = authors.find { it.id == authorId }
-    fun totalPages(): Int = pageCount
+    fun pageCount(): Int = totalPages
 }
 
 class Author(
--- a/docs/graphql-java-comparison.md
+++ b/docs/graphql-java-comparison.md
@@ -37,24 +37,38 @@ type Author {
 Then write the runtime code that matches this schema to build the `GraphQLSchema` object.
 
 ```kotlin
+// Internal DB class, not schema class
+class Book(
+    val id: ID,
+    val name: String,
+    val totalPages: Int, // This needs to be renamed to pageCount
+    val authorId: ID // This is not in the schema
+)
+
+// Internal DB class, not schema class
+class Author(
+    val id: ID,
+    val firstName: String,
+    val lastName: String
+)
+
 class GraphQLDataFetchers {
-    private val books: List<Map<String, String>> = booksFromDB()
-    private val authors: List<Map<String, String>> = authorsFromDB()
+    private val books: List<Book> = booksFromDB()
+    private val authors: List<Author> = authorsFromDB()
 
     fun getBookByIdDataFetcher() = DataFetcher { dataFetchingEnvironment ->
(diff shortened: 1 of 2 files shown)


docs/graphql-java-comparison.md
```diff
@@ -0,0 +1,146 @@
+---
+id: graphql-java-comparison
+title: GraphQL Java Comparison
+---
+
+[graphql-java](https://graphql-java.com/) is one of the more popular JVM based GraphQL implemenations. GraphQL Kotlin is
+built on top of `grahpql-java` as there are many parts of the GraphQL specification that are trivial and this implemenatation
```
dariuszkuc, line 7: Unsure if I would  call the reason why we use it as `trivial pieces of implementation` so maybe we should rephrase that. I'm thinking maybe we should mention rewrite that sentence as

```suggestion
built on top of `grahpql-java` as it can be easily extended with additional functionality and this implementation
```

docs/graphql-java-comparison.md
```diff
@@ -0,0 +1,146 @@
...
+type Author {
+    id: ID!
+    firstName: String!
+    lastName: String!
+}
+```
+
+Then write the runtime code that matches this schema to build the `GraphQLSchema` object.
+
+```java
```
dariuszkuc, line 39: do we want to change this example to Kotlin (so it is easier to compare vs plain java)?

dariuszkuc (changes requested):

docs/graphql-java-comparison.md
smyrick: Showing it in java makes it look more verbose intentionally 😜

smyrick: This comes from this example: https://www.graphql-java.com/tutorials/getting-started-with-spring-boot/

docs/graphql-java-comparison.md
```diff
@@ -0,0 +1,146 @@
...
+    public DataFetcher getPageCountDataFetcher() {
+        return dataFetchingEnvironment -> {
+            Map<String,String> book = dataFetchingEnvironment.getSource();
+            return book.get("totalPages");
+        };
+    }
+}
+```
+
+This means that there are two sources-of-truth for your schema and changes in either have to be reflected in both locations.
```
EXPErdelfavero, line 106: "sources of truth", no hyphens

(Github's not letting me edit directly for some reason...)

docs/graphql-java-comparison.md
```diff
@@ -0,0 +1,146 @@
...
+As your schema scales to hundreds of types and many different resolvers, it can get more difficult to track what code needs to be changed if you want to add a new field,
+deprecated or delete an existing one, or fix a bug in the resolver code.
```
EXPErdelfavero, line 108: "deprecate", no 'd' at the end

docs/graphql-java-comparison.md
```diff
@@ -0,0 +1,122 @@
...
+    firstName: String!
+    lastName: String!
+}
+```
+
+Then write the runtime code that matches this schema to build the `GraphQLSchema` object.
+
+```kotlin
+class GraphQLDataFetchers {
+    private val books: List<Map<String, String>> = booksFromDB()
```
dariuszkuc, line 41: pretty sure we should be using data classes here instead of raw maps?

dariuszkuc (approved):

Issue #863, opened 2020-09-02
Kotlin data classes to TypeDefinitionRegistry


claymccoy:
From the high level usage I see in the docs, this project isn't actually helpful for me to generate my GraphQL schema from Kotlin data classes. I don't want to have to create a top level query object with methods and arguments. That is redundant. I just want to supply some classes and have this top level query with methods and args generated from those classes. Avoiding duplication is part the reason I wanted to use this library.

What I really need is a way to turn my Kotlin data classes into a graphql.schema.idl.TypeDefinitionRegistry that I can then hand to something else. I imagine you must be doing something simlar here with reflection. Are there some internals that I can use so that I don't have to write that all over again?

Thanks!

smyrick:
@claymccoy 

> I just want to supply some classes and have this top level query with methods and args generated from those classes

That is exactly what the [graphql-kotlin-schema-generator ](https://github.com/ExpediaGroup/graphql-kotlin/tree/master/graphql-kotlin-schema-generator) does.

You just have to call the `toSchema` with your Kotlin classes and it generates a `GraphQLSchema` object from `graphql-java`.

From there you can use that object to start a server, or use `graphql-kotlin-spring-server` which handles the server logic for you

claymccoy:
Right, I am familiar with the intended use and I am trying to communicate how id doesn't fit my needs. Using toSchema doesn't do enough of the generation for me and outputs a target that I don't want.
1) I shouldn't have to create the Query object if I just want all of my classes to be top level queries.
2) I also need to use another tool (neo4j-graphql-java) which needs either:
* the schema String
* graphql.schema.idl.TypeDefinitionRegistry.

I'd rather not rewrite all the reflection logic that you already have, but I want my model classes to output one of the two types used by neo4j-graphql-java above.

I was looking to "abuse" your library by doing one of the following.
* Directly use the functions that do the conversion using Kotlin reflection, but they are all internal.
* Putting my model in as additional types, letting your library generate the GraphQLSchema, and then trying to turn the GraphQLObjectTypes into ObjectTypeDefinitions.

claymccoy:
More directly I want to convert a Kotlin class like this:
```
@NodeEntity
data class Artifact(
        val groupId: String,
        val artifactId: String,
        val version: String,
        @Relationship(type = "IN_REPO", direction = Relationship.OUTGOING) val repo: ArtifactoryRepo?,
[...]

PR #873, opened 2020-09-06
[generator] Move type generator to separate class

smyrick:
### :pencil: Description
Expose a new public class, `TypeGenerator`, that allows anyone to convert any `KType` to a `GraphQLType`. This could be used outside the schema generator if needed

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/863


dariuszkuc:
Whats the use case behind this?

smyrick:
We have had a few people ask about generating code outside of a full schema or getting access to just the generator.

We could also use this internally for our shared types in Expedia since we just use a mock query object.

This could be seen as leaking internals which is why I wanted more opinions but the only new method is `generateType`. The config options are still the external ones!

dariuszkuc:
IMHO it is leaking internals of the library. One potential problem scenario could arise when custom hooks would use their own instance of type generator to build some custom types (which is good) but since it would use separate type cache (from the schema generator) there is a potential for some conflicts.

I think we should revisit our hooks to make it more useable.

smyrick:
We are going to close this as it does raise some complications

* What happens when you call it twice with the same type, you will get a type reference back that you can't really do anything with.
* We now have more coverage or areas that require a major version if we ever want to introduce a breaking change
* This can be solve by creating a simple query object, even if not used, and generating a full schema, either by using the types or using the additional types


If we have more requests for something similar to this in the future we can reopen the PR

PR #879, opened 2020-09-11
Allow context to be scope for suspend function fetchers

josephlbarnett:
### :pencil: Description

When executing a suspend function as a data fetcher, check
if the GraphQL context object implements CoroutineScope. If
it does, use that scope to launch the suspend function call
instead of using GlobalScope, allowing for structured
concurrency concepts to apply (coroutineContext inheritance,
cancellation, etc).

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
@@ -27,6 +27,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import graphql.schema.DataFetcher
 import graphql.schema.DataFetchingEnvironment
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.future.future
@@ -64,7 +65,8 @@ open class FunctionDataFetcher(
             val parameterValues = getParameterValues(fn, environment)
 
             if (fn.isSuspend) {
-                runSuspendingFunction(it, parameterValues)
+                val scope = (environment.getContext<Any?>() as? CoroutineScope) ?: GlobalScope
+                runSuspendingFunction(it, parameterValues, scope)
             } else {
                 runBlockingFunction(it, parameterValues)
             }
@@ -136,9 +138,10 @@ open class FunctionDataFetcher(
     protected open fun runSuspendingFunction(
         instance: Any,
         parameterValues: Array<Any?>,
+        scope: CoroutineScope,
         coroutineContext: CoroutineContext = EmptyCoroutineContext,
         coroutineStart: CoroutineStart = CoroutineStart.DEFAULT
-    ): CompletableFuture<Any?> = GlobalScope.future(context = coroutineContext, start = coroutineStart) {
+    ): CompletableFuture<Any?> = scope.future(context = coroutineContext, start = coroutineStart) {
         try {
             fn.callSuspend(instance, *parameterValues)
         } catch (exception: InvocationTargetException) {
--- a/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcherTest.kt
+++ b/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcherTest.kt
@@ -22,7 +22,14 @@ import graphql.GraphQLException
 import graphql.schema.DataFetchingEnvironment
 import io.mockk.every
 import io.mockk.mockk
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.future.await
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import org.junit.jupiter.api.Test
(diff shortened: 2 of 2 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
```diff
@@ -64,7 +65,12 @@ open class FunctionDataFetcher(
             val parameterValues = getParameterValues(fn, environment)
 
             if (fn.isSuspend) {
-                runSuspendingFunction(it, parameterValues)
+                val scope = try {
+                    environment.getContext<CoroutineScope>()!!
+                } catch (e: Exception) {
+                    GlobalScope
+                }
```
smyrick, line 72: Instead of throwing and catching exceptions on null can this just be the following? I haven't looked at the method signatures fully

```suggestion
                val scope = environment.getContext<CoroutineScope>() ?: GlobalScope
```

josephlbarnett: `environment.getContext<T>` looks like it just blindly does a cast to `T`, so would throw a ClassCastException if the context didn't implement scope.  maybe could do something like `(environment.getContext<Any>() as? CoroutineScope) ?: GlobalScope`, but not sure if that's better or worse?

smyrick: I think that is better than using `!!` as we can see what the code would be failing to do that returns null

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
```diff
@@ -136,9 +142,10 @@ open class FunctionDataFetcher(
     protected open fun runSuspendingFunction(
         instance: Any,
         parameterValues: Array<Any?>,
+        scope: CoroutineScope,
```
smyrick, line 145: Breaking changes since we modifying a public method signature

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt
```diff
@@ -64,7 +65,8 @@ open class FunctionDataFetcher(
             val parameterValues = getParameterValues(fn, environment)
 
             if (fn.isSuspend) {
-                runSuspendingFunction(it, parameterValues)
+                val scope = (environment.getContext<Any?>() as? CoroutineScope) ?: GlobalScope
```
dariuszkuc, line 68: I dont think this will work in a generic way, i.e. context can by anything and we are auto populating context parameter here -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/execution/FunctionDataFetcher.kt#L93

This change implies that context has to be of CoroutineScope type in order for this to work reliably.

dariuszkuc: There is also another problem with parent-child relationship - i.e. this will work for top level queries but it won't have proper relationship for the subqueries

josephlbarnett: Agreed this will only work for users that populate a context object that implements CoroutineScope, but couldn't figure out a way for the inner scope to be a child of an outer scope when there's non-suspend functions in between, so some way to pass the scope down seems to be needed.  Other than the context, can you think of a way to pass it down on a per-query basis?

re: parent-child relationships, where's the best example of one of those?  I'll take a look and see if thats solveable in a similar way.

dariuszkuc: Ideally this would all be automatically populated on the data fetching environment but without rewriting `graphql-java` execution logic its not going to be clean integration.

In order to preserve the parent-child relationship I guess you could store coroutine scopes in a map (or a tree) - basically you would look up parent scope (by path) and start resolver in a new child scope (that would also have to be stored in the same map so its children resolvers would be able to look it up). Since you would have to manually manage those relationships I think it would be somewhat problematic but I guess doable.

I think bigger problem would be on the cancellations -> per spec each field is resolved individually and unless there is some critical exception (i.e. attempting to return null for non-nullable field) they should resolve separately. You probably would need to use supervisor job to allow individual resolver failures but there also would have to be some mechanism to do the critical failure.

Due to the all of the above I don't think it is currently doable in a way that is generic and could be integrated to the lib*.

---
*without the execution logic rewrite to be coroutine aware

josephlbarnett: Yeah, been running with this + a supervisor scope added to the context and it seems to be working well in that:
 * the scope's CoroutineContext gets used (eg MDContext() gets applied properly)
 * cancelling the GraphQLContext scope allows for proper individual resolver failures to show up in `errors` as cancellations while resolvers that occur before cancellation still end up in `data`

While parent-child scope isn't being handled explicitly, it still seems to work "properly" for the cases I've tested -- ie having all resolvers share the same scope vs/ having hierarchical scope isn't manifesting itself as a problem.  Also not handling critical exceptions as failing the whole request, but at least for my cases would prefer all exceptions end up in `errors` vs/ failing the `graphql.execute*()` call and bubbling up?  That being said, I'm sure there are use cases that I'm not covering in my usage, and so understand your concerns.

dariuszkuc:
Going to close this for now as while it would work in some cases it does not solve the problem generically (i.e. does not honor proper parent-child relationship of coroutines and does not handle critical cancellation).

Issue #864, opened 2020-09-03
Two data classes are generated are from the same type


haagenhasle:
I'm using graphql-kotlin version 3.6.1. 

I'm generating code base on a schema and a query-file. In three different places in the query-file, the same type is referenced. (FolkeregisterMetadata). When graphql-kotlin generates code for this, it generate two data classes instead of one, FolkeregisterMetadata and FolkeregisterMetadata1. The first one is used two places, and the second one the third place. I can't find any differences in the query that should cause this, so I assume it could be a bug? I would have expected just one FolkeregisterMetadata-dataclass to be generated.

I have zipped up a small project that should show this, where the query is included. (The schema is in a public github repo, and is referenced in the build)

[TestKotlinGradle.zip](https://github.com/ExpediaGroup/graphql-kotlin/files/5167964/TestKotlinGradle.zip)

smyrick:
Hello, for graphql type-safe clients we need to have separate classes based of your field selection. That is because as a client you are not required to select all the fields. So the easiest way for us to handle that, and make sure that we don't populate used fields as null as that would imply the server returned null when it really didn't, is to have two separate classes

See: https://github.com/ExpediaGroup/graphql-kotlin/issues/815

smyrick:
If you do have a example of an unexpected, error please include a link to a Github repo with a branch we can clone instead of a zip file

haagenhasle:
You can take a look at https://github.com/haagenhasle/GraphQLKotlinTest

You can see there that Folkeregistermetadata2 is actually used twice, and Folkeregistermetadata is used once. So it seems you are able to reuse a class (or I would have had a Folkeregistermetadata3 as well?), so why don't you do it every time? 

I saw dariuszkuc's comment on the issue you linked to, and he mentioned "different selection sets for the same object". But in my case, I have the same selection set every time.

This is not a big deal, we can easily work with two identical classes. It is merely an inconvenience. I really like the library you have made, and thank you for your effort! :)

dariuszkuc:
Thanks for reporting. Sounds like a bug -> if you have same selection set it should reuse same data class. I'll take a look!

PR #883, opened 2020-09-14
[client] fix generation of duplicate classes

dariuszkuc:
### :pencil: Description

We were incorrectly using simple class name for the type lookup as all classes are wrapped in some root type (i.e. `Query.MyClass`) whereas GraphQL type name was just the actual type. We were also incorrectly comparing selection sets as we were not unwrapping selection sets of the children fields.

### :link: Related Issues
Fixes: https://github.com/ExpediaGroup/graphql-kotlin/issues/864

### DIFF
--- a/plugins/graphql-kotlin-plugin-core/src/main/kotlin/com/expediagroup/graphql/plugin/generator/types/generateTypeName.kt
+++ b/plugins/graphql-kotlin-plugin-core/src/main/kotlin/com/expediagroup/graphql/plugin/generator/types/generateTypeName.kt
@@ -97,7 +97,7 @@ internal fun generateCustomClassName(context: GraphQLClientGeneratorContext, gra
     } else {
         // verify we got same selection set for interface and/or objects (unions shouldn't have any fields)
         for (cachedType in cachedTypeNames) {
-            if (isCachedTypeApplicable(context, cachedType.simpleName, graphQLTypeDefinition, selectionSet)) {
+            if (isCachedTypeApplicable(context, cachedType.simpleNameWithoutWrapper(), graphQLTypeDefinition, selectionSet)) {
                 return cachedType
             }
         }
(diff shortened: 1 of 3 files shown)


plugins/graphql-kotlin-plugin-core/src/main/kotlin/com/expediagroup/graphql/plugin/generator/types/generateTypeName.kt
```diff
@@ -148,21 +151,28 @@ private fun isCachedTypeApplicable(context: GraphQLClientGeneratorContext, graph
...
+    path: String = ""
 ): Set<String> {
     val result = mutableSetOf<String>()
     selectionSet.selections.forEach { selection ->
         when (selection) {
-            is Field -> result.add(selection.name)
+            is Field -> {
+                result.add(path + selection.name)
+                if (selection.selectionSet != null) {
+                    result.addAll(calculateSelectedFields(context, targetType, selection.selectionSet, "$path${selection.name}."))
```
smyrick, line 173: Is this supposed to be `"$path.${selection.name}" `?

dariuszkuc: path is initialize as empty string and then we pass it as `path + selection + .` so current logic is correct (otherwise we end up with `.` prefix)

smyrick: What is the end result here because `pathfoo.` doesn't seem like a correct string

dariuszkuc: it is a prefix for recursive selection, `path` is a prefix
```
[0] prefix = "", field = "foo" -> "foo"
[1] prefix = "foo.", field = "bar" -> "foo.bar"
[2] prefix = "foo.bar.", field = "baz" -> "foo.bar.baz"
```

smyrick (approved):

PR #894, opened 2020-10-01
Update README with discussions links

smyrick:
### :pencil: Description
Drive more discussion with in Github by adding links and badges to [GitHub Discussions](https://github.com/ExpediaGroup/graphql-kotlin/discussions). This means all our open issues and questions will be searchable all within the repo


dariuszkuc:
+1 to point to discussions but I would not remove the reference to the public slack

dariuszkuc (approved):

Issue #899, opened 2020-10-07
Can't overwrite default buffer size


sobolewskikamil:
Hi,

We are using `graphql-kotlin-spring-client:4.0.0-alpha.3` to retrieve quite big response from server. We stumble across following error:
```org.springframework.core.io.buffer.DataBufferLimitException: Exceeded limit on max bytes to buffer : 262144```

We are aware that this is WebClient configuration concern, so we have configured WebClient in the following way in order to make the limit higher:
```
val webClientBuilder = WebClient.builder()
            .exchangeStrategies(
                ExchangeStrategies.builder()
                    .codecs { configurer ->
                        configurer
                            .defaultCodecs()
                            .maxInMemorySize(100 * 1024 * 1024)
                    }
                    .build())
```
Afterwards we create `GraphQLWebClient` in the following way: 
```
private val client = GraphQLWebClient(url, objectMapper, webClientBuilder)
```

Unfortunately I doesn't work and somehow our configuration is lost. We suspect that it is caused by `init` block inside `GraphQLWebClient` class.
```
init {
        mapper.enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)

        val strategies = ExchangeStrategies
            .builder()
            .codecs { configurer ->
                configurer.defaultCodecs().jackson2JsonEncoder(Jackson2JsonEncoder(mapper, MediaType.APPLICATION_JSON))
                configurer.defaultCodecs().jackson2JsonDecoder(Jackson2JsonDecoder(mapper, MediaType.APPLICATION_JSON))
            }.build()
        builder.exchangeStrategies(strategies)
    }
```

As you can see you are overwriting exchange strategies which causes configuration loss. Could you recommend some way to properly configure your client? Thanks :)

dariuszkuc:
Nice catch! Indeed `init` will overwrite your configuration - I'll look into properly merging those configs

sobolewskikamil:
@dariuszkuc Could you provide any ETA of the release with fix?

dariuszkuc:
@sobolewskikamil I just released the fix in `4.0.0-alpha.6`

PR #900, opened 2020-10-07
[client] update GraphQLWebClient to correctly apply codecs

dariuszkuc:
### :pencil: Description

Apply default codecs for JSON encoder/decoder without overwriting whole exchange strategies.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/899


dariuszkuc:
Ideally it would be great if we could do something like create builder with defaults and then apply the builder provided by the user but based on the [Spring docs](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-webclient-customization) there doesn't appear to be a simple way to do this.

Spring `WebClient.Builder` currently only expose `apply(Consumer<WebClient.Builder> builderConsumer)` which could be used to customize the builder but that would also complicate integrating client with Spring apps -> i.e. instead of autowiring beans managed by Spring users would have to explicitly provide lambda to customize the behavior. Unless Spring api changes I think this is probably the "best" we can do atm.

smyrick (approved):

Issue #893, opened 2020-10-01
[gradle plugin] GraphQL extension cannot be used from Groovy scripts


dariuszkuc:
**Library Version**
`4.0.0-alpha.4`

**Describe the bug**
[GraphQLPluginExtension](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt) is using Kotlin lambdas for configuration which makes it unusable from Groovy build files. We should use Gradle `Action` instead of lambdas - i.e. instead of `GraphQLPluginClientExtension.() -> Unit` use `Action<GraphQLPluginClientExtension>`

**To Reproduce**
Apply `graphql` extension in `build.gradle`

```groovy
graphql {
    client {
        endpoint = "http://localhost:8080/graphql"
        packageName = "com.example"
        clientType = GraphQLClientType.KTOR
    }
}
```

**Expected behavior**
GraphQL extension DSL can be used from both Groovy and Kotlin scripts.

**Workaround**
Instead of extension we can configure tasks explicitly

```groovy
tasks.getByName("graphqlIntrospectSchema") {
    endpoint.set("http://localhost:8080/graphql")
}
tasks.getByName("graphqlGenerateClient") {
    packageName.set("com.example")
    clientType.set(com.expediagroup.graphql.plugin.generator.GraphQLClientType.KTOR)
    schemaFile.set(graphqlIntrospectSchema.outputFile)
    dependsOn("graphqlIntrospectSchema")
}
```

dariuszkuc:
When working on this issue we should also update our documentation with samples for both Kotlin and Groovy based build files.

Issue #892, opened 2020-10-01
Gradle build make error


mmaryo:
Hello 

My project 
Kotlin 1.4.0
Quarkus 1.8.1.Final
graphql-kotlin 3.6.4 and 4.0.0-alpha.4

I follow tutorial for create a client in my project 

https://expediagroup.github.io/graphql-kotlin/docs/client/client-overview#build-configuration

I simply add in build.gradle

    implementation("com.expediagroup:graphql-kotlin-ktor-client:$graphqlKotlin")

and 

```
graphql {
    client {
        endpoint = "https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v2"
        packageName = "com.app.uniswap"
        clientType = GraphQLClientType.KTOR
    }
}

```

And I have a file .graphql inside resources folder 

gradle build throw this error 

> Could not set unknown property 'endpoint' for extension 'graphql' of type com.expediagroup.graphql.plugin.gradle.GraphQLPluginExtension.

What is wrong ?

mmaryo:
Hello @dariuszkuc 

This is the repo https://github.com/mmaryo/test-graphql-kotlin

and do ./gradlew build

dariuszkuc:
Thanks for raising this, it is indeed a bug on our end (https://github.com/ExpediaGroup/graphql-kotlin/issues/893) - only tested the plugin from Kotlin DSL. TLDR I've used Kotlin lambda which cannot be used from Groovy script so need update the plugin code to make extension useable from both Kotlin and Groovy.

In the meantime you will have to explicitly configure the tasks (or migrate to use KTS)
```
tasks.getByName("graphqlIntrospectSchema") {
    endpoint.set("https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v2")
}
tasks.getByName("graphqlGenerateClient") {
    packageName.set("com.mario.uniswap")
    clientType.set(com.expediagroup.graphql.plugin.generator.GraphQLClientType.KTOR)
    schemaFile.set(graphqlIntrospectSchema.outputFile)
    dependsOn("graphqlIntrospectSchema")
}
```

PR #901, opened 2020-10-08
[plugin] update Gradle plugin extension to use Actions

dariuszkuc:
### :pencil: Description
Originally extension was using Kotlin lambdas for some configuration which worked fine when used from Kotlin script (`build.gralde.kts`) but it could not be easily used from Groovy scripts (`build.gradle`). By changing from lambda with a receiver to a Groovy Action the resulting code will be translated using SAM conversion that makes it possible to configure the extension using Kotlin lambda or Groovy closure.

### :link: Related Issues
Resolves: #893
Discussion: #892

### DIFF
--- a/docs/plugins/gradle-plugin.md
+++ b/docs/plugins/gradle-plugin.md
@@ -10,12 +10,64 @@ GraphQL Kotlin Gradle Plugin provides functionality to introspect GraphQL schema
 `graphql-kotlin-gradle-plugin` is published on Gradle [Plugin Portal](https://plugins.gradle.org/plugin/com.expediagroup.graphql).
 In order to execute any of the provided tasks you need to first apply the plugin on your project.
 
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Kotlin-->
+
+Using plugins DSL syntax
+
 ```kotlin
 // build.gradle.kts
 plugins {
     id("com.expediagroup.graphql") version $graphQLKotlinVersion
 }
 ```
+
+Or by using legacy plugin application
+
+```kotlin
+// build.gradle.kts
+buildscript {
+  repositories {
+    maven {
+      url = uri("https://plugins.gradle.org/m2/")
+    }
+  }
+  dependencies {
+    classpath("com.expediagroup:graphql-kotlin-gradle-plugin:$graphQLKotlinVersion")
+  }
+}
+
+apply(plugin = "com.expediagroup.graphql")
+```
+<!--Groovy-->
+
+Using plugins DSL syntax
+
+```groovy
+// build.gradle
+plugins {
+    id 'com.expediagroup.graphql' version $graphQLKotlinVersion
+}
+```
+
+Or by using legacy plugin application
+
+```groovy
+// build.gradle
+buildscript {
+  repositories {
+    maven {
+      url "https://plugins.gradle.org/m2/"
+    }
+  }
+  dependencies {
+    classpath "com.expediagroup:graphql-kotlin-gradle-plugin:$graphQLKotlinVersion"
+  }
+}
+
+apply plugin: "com.expediagroup.graphql"
+```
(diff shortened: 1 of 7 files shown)


docs/plugins/gradle-plugin.md
```diff
@@ -10,13 +10,65 @@ GraphQL Kotlin Gradle Plugin provides functionality to introspect GraphQL schema
 `graphql-kotlin-gradle-plugin` is published on Gradle [Plugin Portal](https://plugins.gradle.org/plugin/com.expediagroup.graphql).
 In order to execute any of the provided tasks you need to first apply the plugin on your project.
 
+<!--DOCUSAURUS_CODE_TABS-->
```
dariuszkuc, line 13: with docusaurus v2 we could sync all of those tabs so switching anywhere on the page would switch all code samples

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt
```diff
@@ -62,7 +63,7 @@ open class GraphQLPluginClientExtension {
     /** Connect and read timeout configuration for executing introspection query/download schema */
     internal val timeoutConfig: TimeoutConfig = TimeoutConfig()
 
-    fun timeout(config: TimeoutConfig.() -> Unit = {}) {
-        timeoutConfig.apply(config)
+    fun timeout(action: Action<TimeoutConfig>) {
```
dariuszkuc, line 66: there appears to be an issue with SAM conversion or maybe it is Groovy issue but I'm hitting a problem when attempting to invoke `Action` from another `Action`, i.e.

```groovy
graphql {
    client {
        endpoint = "http://localhost:8080/graphql"
        packageName = "com.example.generated"
        // above works fine, but complains on the below, looks like "this" from the action refers to outer one
        timeout {
            connect = 10000
            read = 30000
        }
    }
}
```

results in 

`Could not set unknown property 'connect' for object of type com.expediagroup.graphql.plugin.gradle.GraphQLPluginClientExtension.`

smyrick (approved):

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt
krispharper: Is there a solution for this? I'm trying to set a connection timeout value per the documentation [here](https://expediagroup.github.io/graphql-kotlin/docs/plugins/gradle-plugin-tasks/#extension) and getting

```
> Could not set unknown property 'connect' for object of type com.expediagroup.graphql.plugin.gradle.GraphQLPluginClientExtension.
```

dariuszkuc: @krispharper - which version of the plugin are you using? this should be working fine now

dariuszkuc: if you are using Groovy you might need to explicitly name the parameter as otherwise it might be confused on which object to apply those params.

```groovy
timeout { t ->
  t.connect = 10000
  t.read = 30000
}
```

krispharper: Got it, making that change worked. Thanks a bunch.

Issue #886, opened 2020-09-18
[gradle plugin] queryFileDirectory cannot be set through ClientExtentions


huehnerlady:
**Is your feature request related to a problem? Please describe.**
So far you can set the `queryFileDirectory` via Task, but not via global configuration.
With the configuration you can just use the `queryFiles` property.
Would be good to add this in the plugin configuration

**Describe the solution you'd like**
```
graphql {
  client {
   [...]
    queryFileDirectory = "src/main/resources/graphql-queries"
  }
}
```

**Describe alternatives you've considered**
```
graphql {
  client {
   [...]
    queryFiles = fileTree("src/main/resources/graphql-queries").files
  }
}
```
This does not look very tidy and much more like a hack

PR #914, opened 2020-10-22
[plugin] support query file directory in Gradle extension

dariuszkuc:
### :pencil: Description

Adds ability to specify custom query file directory (previously we had to explicitly provide a list of query files).

```kotlin
graphql {
  client {
    endpoint = "http://localhost:8080/graphql"
    package = "com.example"
    queryFileDirectory = "${project.projectDir}/src/main/resources/queries"
  }
}
```

### :link: Related Issues

Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/886

### DIFF
--- a/docs/plugins/gradle-plugin.md
+++ b/docs/plugins/gradle-plugin.md
@@ -102,7 +102,9 @@ graphql {
     headers = mapOf("X-Custom-Header" to "Custom-Header-Value")
     // Target package name to be used for generated classes.
     packageName = "com.example.generated"
-    // Optional list of query files to be processed, if not specified will default to all query files under src/main/resources.
+    // Custom directory containing query files, defaults to src/main/resources
+    queryFileDirectory = "${project.projectDir}/src/main/resources/queries"
+    // Optional list of query files to be processed, takes precedence over queryFileDirectory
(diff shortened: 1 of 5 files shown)


dariuszkuc:
Some additional considerations:
* `queryFileDirectory` accept a String whereas `queryFiles` accept a list of files -> should we update those to be consistent?
  
  I kept `queryFileDirectory` as String to keep it consistent with the same property on the `graphqlGenerateClient` task which has to be a String in order to accept command line input. Making it a proper file would mean we would have to read its absolute path and set the String value on the task which then converts it back to a file...
* `queryFileDirectory` assumes passed in String is an absolute path -> should we add some logic to detect whether it is absolute and if not then automatically prepend the workspace directory? If we do automatically convert it, do we keep it simple by just checking `/` or should we also consider Windows paths?

smyrick:
I think at this point we can keep it simple and have the users specify the file location they want through Gradle configuration. If at some later point we want to add more features to resolve the path for them we can create a separate PR. But I like just having it be a string and assuming the path, so anyone using windows can just specify there custom path

smyrick (approved):

PR #924, opened 2020-10-28
[3.x.x][generator] Only add directives with valid locations

smyrick:
## :pencil: Description

When generating a schema you can have an annotation on any part of the schema, assuming you haven't set the Target, so if you are restricting the directive locations to only specific places they may still be added to the schema and then fail generation as they are added to invalid locations. While this could be viewed as a customer error, this is more of an issue with graphql-kotlin because we can have input and output objects with the same directives and things like the KeyDirective from federation are only valid on output. So in that case we shouldn't fail if someone uses the type as input elsewhere, it will still not be marked as a federation type for the input

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
@@ -21,7 +21,7 @@ import com.expediagroup.graphql.generator.extensions.getPropertyAnnotations
 import com.expediagroup.graphql.generator.extensions.getSimpleName
 import com.expediagroup.graphql.generator.extensions.getValidProperties
 import com.expediagroup.graphql.generator.extensions.safeCast
-import graphql.introspection.Introspection
+import graphql.introspection.Introspection.DirectiveLocation
 import graphql.schema.GraphQLArgument
 import graphql.schema.GraphQLDirective
 import java.lang.reflect.Field
@@ -33,7 +33,7 @@ import com.expediagroup.graphql.annotations.GraphQLDirective as GraphQLDirective
 internal fun generateDirectives(
     generator: SchemaGenerator,
     element: KAnnotatedElement,
-    location: Introspection.DirectiveLocation,
+    location: DirectiveLocation,
     parentClass: KClass<*>? = null
 ): List<GraphQLDirective> {
     val annotations = when {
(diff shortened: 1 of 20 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateSubscription.kt
```diff
@@ -38,6 +39,10 @@ internal fun generateSubscriptions(generator: SchemaGenerator, subscriptions: Li
             throw InvalidSubscriptionTypeException(kClass)
         }
 
+        generateDirectives(generator, subscription.kClass, DirectiveLocation.OBJECT).forEach {
```
smyrick, line 42: This is also a bug fix as we were not adding directives on subscriptions

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateQuery.kt
```diff
@@ -32,7 +33,7 @@ internal fun generateQueries(generator: SchemaGenerator, queries: List<TopLevelO
             throw InvalidQueryTypeException(query.kClass)
         }
 
-        generateDirectives(generator, query.kClass).forEach {
+        generateDirectives(generator, query.kClass, DirectiveLocation.OBJECT).forEach {
```
smyrick, line 36: The location for queries, mutations, and subscriptions is `OBJECT`. The other locations `QUERY`, `MUTATION`, `SUBSCRIPTION` are for client query directives

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
```diff
@@ -29,20 +30,27 @@ import kotlin.reflect.KClass
...
     }
 
     return annotations
         .mapNotNull { it.getDirectiveInfo() }
+        .filter { it.directiveAnnotation.locations.contains(location) }
         .map { getDirective(generator, it) }
 }
 
-internal fun generateFieldDirectives(generator: SchemaGenerator, field: Field): List<GraphQLDirective> =
+internal fun generateFieldDirectives(generator: SchemaGenerator, field: Field, location: Introspection.DirectiveLocation): List<GraphQLDirective> =
```
dariuszkuc, line 50: can you update the name to match the latest changes from master?

smyrick: Updated

dariuszkuc (approved):

PR #925, opened 2020-10-28
[generator] Only add directives with valid locations

smyrick:
### :pencil: Description
When generating a schema you can have an annotation on any part of the schema, assuming you haven't set the Target, so if you are restricting the directive locations to only specific places they may still be added to the schema and then fail generation as they are added to invalid locations. While this could be viewed as a customer error, this is more of an issue with graphql-kotlin because we can have input and output objects with the same directives and things like the KeyDirective from federation are only valid on output. So in that case we shouldn't fail if someone uses the type as input elsewhere, it will still not be marked as a federation type for the input

### :link: Related Issues
Copy changes from https://github.com/ExpediaGroup/graphql-kotlin/pull/924 to the master branch

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
@@ -21,7 +21,7 @@ import com.expediagroup.graphql.generator.extensions.getPropertyAnnotations
 import com.expediagroup.graphql.generator.extensions.getSimpleName
 import com.expediagroup.graphql.generator.extensions.getValidProperties
 import com.expediagroup.graphql.generator.extensions.safeCast
-import graphql.introspection.Introspection
+import graphql.introspection.Introspection.DirectiveLocation
 import graphql.schema.GraphQLArgument
 import graphql.schema.GraphQLDirective
 import java.lang.reflect.Field
@@ -33,7 +33,7 @@ import com.expediagroup.graphql.annotations.GraphQLDirective as GraphQLDirective
 internal fun generateDirectives(
     generator: SchemaGenerator,
     element: KAnnotatedElement,
-    location: Introspection.DirectiveLocation,
+    location: DirectiveLocation,
     parentClass: KClass<*>? = null
 ): List<GraphQLDirective> {
     val annotations = when {
@@ -47,10 +47,10 @@ internal fun generateDirectives(
         .map { getDirective(generator, it) }
 }
 
-internal fun generateFieldDirectives(generator: SchemaGenerator, field: Field): List<GraphQLDirective> =
+internal fun generateEnumValueDirectives(generator: SchemaGenerator, field: Field): List<GraphQLDirective> =
     field.annotations
         .mapNotNull { it.getDirectiveInfo() }
-        .filter { it.directiveAnnotation.locations.contains(Introspection.DirectiveLocation.ENUM_VALUE) }
+        .filter { it.directiveAnnotation.locations.contains(DirectiveLocation.ENUM_VALUE) }
         .map { getDirective(generator, it) }
 
 private fun getDirective(generator: SchemaGenerator, directiveInfo: DirectiveInfo): GraphQLDirective {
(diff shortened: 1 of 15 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateSubscription.kt
```diff
@@ -38,6 +39,10 @@ internal fun generateSubscriptions(generator: SchemaGenerator, subscriptions: Li
             throw InvalidSubscriptionTypeException(kClass)
         }
 
+        generateDirectives(generator, subscription.kClass, DirectiveLocation.OBJECT).forEach {
+            subscriptionBuilder.withDirective(it)
+        }
+
```
smyrick, line 45: This is also a bug fix as we were not adding directives on subscriptions

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateQuery.kt
```diff
@@ -32,7 +33,7 @@ internal fun generateQueries(generator: SchemaGenerator, queries: List<TopLevelO
             throw InvalidQueryTypeException(query.kClass)
         }
 
-        generateDirectives(generator, query.kClass).forEach {
+        generateDirectives(generator, query.kClass, DirectiveLocation.OBJECT).forEach {
```
smyrick, line 36: The location for queries, mutations, and subscriptions is `OBJECT`. The other locations `QUERY`, `MUTATION`, `SUBSCRIPTION` are for client query directives

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
```diff
@@ -29,20 +30,27 @@ import kotlin.reflect.KClass
...
+    parentClass: KClass<*>? = null
+): List<GraphQLDirective> {
     val annotations = when {
         element is KProperty<*> && parentClass != null -> element.getPropertyAnnotations(parentClass)
         else -> element.annotations
     }
 
     return annotations
         .mapNotNull { it.getDirectiveInfo() }
+        .filter { it.directiveAnnotation.locations.contains(location) }
```
dariuszkuc, line 46: should we fail if someone attempts to add directive on invalid location? guess it might be tricky in case of input/output objects that can be represented by the same class

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
```diff
@@ -29,20 +30,27 @@ import kotlin.reflect.KClass
...
         .map { getDirective(generator, it) }
 }
 
 internal fun generateFieldDirectives(generator: SchemaGenerator, field: Field): List<GraphQLDirective> =
     field.annotations
         .mapNotNull { it.getDirectiveInfo() }
+        .filter { it.directiveAnnotation.locations.contains(Introspection.DirectiveLocation.ENUM_VALUE) }
```
dariuszkuc, line 53: shouldn't this be `DirectiveLocation. FIELD_DEFINITION` or `DirectiveLocation. INPUT_FIELD_DEFINITION`?

dariuszkuc: looks like it is invoked from enum generator -> should we pass it explicitly instead?

smyrick: The name `generateFieldDirectives` is referring to the fact that enum values are of type `java.lang.reflect.Field`. If we did make it an argument it would just always be the value `ENUM_VALUE`

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
smyrick: That is what we are doing today. So with this change the directive will just not appear in the schema. If someone wants to catch that at Kotlin compile time they should update the `annotation class` `@Target`

graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateDirective.kt
dariuszkuc: this seems misleading then - lets either rename the method name OR make the location type an argument

smyrick: Updated

dariuszkuc (approved):

PR #926, opened 2020-10-30
[client 3.x.x] update ScalarConverter included in 3.x.x to accept any objects (#819)

benimario:
### :pencil: Description
ScalarConverter included in 3.x.x still accepts only String.

### :link: Related Issues
#819


dariuszkuc:
Thank you for raising this but unfortunately we won't be able to merge it into `3.x.x` branch. As per https://github.com/ExpediaGroup/graphql-kotlin/pull/819 this is a major breaking change which according to semantic versioning requires new major version (i.e. `4.x.x`). Change is already available on the current `4.x.x` branch.

Issue #921, opened 2020-10-27
Catch Mutation name collision on generation


dragonazz:
Hello Friends, 
First of all, I'm very thankful to your great library,
we are building our application peacefully with your library

**Is your feature request related to a problem? Please describe.**
When there are Mutation functions with same name, we found that the order in which SchemaGenerator reads depens on the build environment.

In our case,
we have two classes, 'AutoSaveMutation', 'SaveMutation'
They both have function named 'save'

When I build the application on windows10, schema-generator reads 'SaveMutation' first and 'AutoSaveMutation' is aborted
and When I build it on linux(centos 7, ubunto 15) ,schema-generator read 'AutoSaveMutation'

**Describe the solution you'd like**

I would be great if schema-generator finds out such name collision at the boot time and give us warning for it.

**Additional context**

Build Environment - Windows
- Window10
- adoptopenjdk11
- gradle 6.7
- kotlin 1.3.72

Build Environment - linux
- centos 7.7
- adoptopenjdk11
- gradle 6.7
- kotlin 1.3.72

Runtime Environment 
- centos 7.7
- adoptopenjdk11

Thank you

smyrick:
You can not have a mutation with the same name in two different classes that implement the `Mutation` marker interface. Regardless of what order the library reads your beans in it is still an invalid schema.

I don't see why you want want to know that you have a invalid schema at startup time instead of at compile time. You want to catch the errors as soon as possible. Can you clarify a bit more on why you would want to ignore the exceptions?

dariuszkuc:
@smyrick it sounds like there is an issue with current logic that will just overwrite the previous entry -> i.e. https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateMutation.kt#L48 we are processing number of different files and adding fields to the builder (we do the same thing for the query and subscription). We might need to validate whether field already exist before adding it to the root type.

dragonazz:
Thanks @dariuszkuc  ,
Your explanation is correct.

Two classes that implement the mutation and have the same function name do not cause any errors at compile time, 
So, If Schema-generator finds such schema error at building schema, It would be very helpful to write proper schema.

smyrick:
Ok, I see. Yep this is definitely a good enhancement we should add. PRs are always welcome if this is urgent to discuss possible solutions

dragonazz:
[...]

PR #928, opened 2020-11-03
feat: catching field name collision on schema generation

dragonazz:
### :pencil: Description

Since Operation types (mutation, mutation, subscription) can be implemented in plural classes, there are chances to make mistake by implementing plural operation classes with same function name. 
This PR catches such grammar errors on schema generation.

### :link: Related Issues

Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/921

### DIFF
--- a/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateMutation.kt
+++ b/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateMutation.kt
@@ -17,6 +17,7 @@
 package com.expediagroup.graphql.generator.types
 
 import com.expediagroup.graphql.TopLevelObject
+import com.expediagroup.graphql.exceptions.ConflictingFieldsException
 import com.expediagroup.graphql.exceptions.InvalidMutationTypeException
 import com.expediagroup.graphql.generator.SchemaGenerator
 import com.expediagroup.graphql.generator.extensions.getValidFunctions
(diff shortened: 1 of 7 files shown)


graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/types/generateMutation.kt
```diff
@@ -46,6 +47,9 @@ internal fun generateMutations(generator: SchemaGenerator, mutations: List<TopLe
             .forEach {
                 val function = generateFunction(generator, it, generator.config.topLevelNames.mutation, mutation.obj)
                 val functionFromHook = generator.config.hooks.didGenerateMutationField(mutation.kClass, it, function)
+                if (mutationBuilder.hasField(functionFromHook.name)) {
```
dariuszkuc, line 50: wondering whether we should keep mapping of fields to file names - it would provide much more meaningful error message but wondering whether this extra complexity is needed

@smyrick any thoughts?

dragonazz: At first, I also considered keeping mapping information, but it is too much unnecessary code only for error message,

or We could just build the type builder and get the definition, but I don't think this is appropriate neither.

smyrick: I think this is straight forward enough to just fail on first conflict

smyrick (approved):

smyrick:
Thank you for the contribution @dragonazz! 🥳

PR #937, opened 2020-11-12
fix(spring-server): adding credentials by default

libetl:
### :pencil: Description
Hello,

My brand uses a security context provided by cookies.
When opening the playground from a first party domain name (from the domain name having the graphql endpoints),
I want to have the auth cookie sent to the server without having to specify it in the "http headers" section of the playground

### :link: Related Issues
auth and graphql-kotlin jira tickets in my brand.


smyrick:
Looks fine to me, if you need further customization we could also override the existing bean: https://github.com/ExpediaGroup/graphql-kotlin/blob/ef9fd7d641e87d75e768c0b0c81e93397a7cde9b/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/PlaygroundRouteConfiguration.kt#L33

smyrick (approved):

PR #950, opened 2020-11-30
fixes typo and adds tests for subscription (context)

gerald24:
### :pencil: Description
Small typo in the subscription code constant and adds an example unit test

### :link: Related Issues
N\A


smyrick (approved):

smyrick:
This is a breaking change since the import is updated. Will have to keep in the 4.x.x release

Issue #860, opened 2020-09-01
Ktor GraphQL server documentation


Czocher:
**Describe the solution you'd like**
I'd like to use this project with [Ktor](https://ktor.io/) but the documentation seems to be lacking for any framework other than Spring. I believe it would be nice to improve the documentation with some other Kotlin frameworks taken into account, Ktor being one of them. At least some getting started docs would be perfect.

smyrick:
@Czocher If you have a working example PRs are always welcome. We do not directly use Ktor as a server library in Expedia Group so we would not have the knowledge to cover all the use cases appropriately. 

All our libraries (other than `graphql-kotlin-spring-server` and `graphql-kotlin-spring-client` of course) should be able to work with any webframework.

There is a minimal example of how to setup a Spark app here: https://github.com/ExpediaGroup/graphql-kotlin/tree/master/examples/spark

smyrick:
Anyone coming in, adding a new module `examples/kor-server` would be a great place to document the set up and maybe showing any issues or things that are more complicated that we could fix in the library.

sondr3:
I have a working Ktor/GraphQL Kotlin/Exposed-project for a student organization that you can take a look at if you want to. The GraphQL configuration is [here](https://github.com/echo-karriere/backend/blob/master/src/graphql/GraphQLConfiguration.kt), and an example query resolver is [here](https://github.com/echo-karriere/backend/blob/master/src/category/CategoryQueryResolver.kt). It was fairly straight forward to get it working, I might be able to contribute an example/documentation in the near future when I have some time to spare :smile:

dariuszkuc:
Example project added in #951

PR #951, opened 2020-11-30
add ktor example

adamsz-lume:
### :pencil: Description
As requested in https://github.com/ExpediaGroup/graphql-kotlin/issues/860#issuecomment-689067996 this PR is adding a working example of ktor implementation with graphql-kotlin

### :link: Related Issues
- subscriptions cannot be simply implemented due to spring connected implementation of com.expediagroup.graphql.spring.execution.ApolloSubscriptionProtocolHandler. What do you think about exporting that class implementation into a separate dependency? It could be then reused in Spark example as well.
- same with the com.expediagroup.graphql.spring.PlaygroundRouteConfiguration. Maybe it's worth take it out as a simple class/function which returns the html which than can be used in other frameworks.
- another idea would be to create a core library for those

I hope this PR will enable a nice ktor feature implementation of graphql.


smyrick:
@adamsz-lume Thank you for your contributions! To make it more clear this is for server code, let's rename the module to `examples/ktor-server`. I know the current `spring` module doesn't follow this pattern but I think we will rename this to make it more clear

examples/ktor/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+val ktor_version: String by project
+val logback_version: String by project
+
+description = "An Example GraphQL service served by a Spark HTTP server"
```
smyrick, line 4:
```suggestion
description = "An Example GraphQL service served by a Ktor HTTP server"
```

examples/ktor/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+
+val kotlinCoroutinesVersion: String by project
+dependencies {
+    implementation("io.ktor:ktor-server-core:$ktor_version")
+    implementation("io.ktor:ktor-server-netty:$ktor_version")
+    implementation("ch.qos.logback:logback-classic:$logback_version")
+
+    implementation("com.expediagroup", "graphql-kotlin-schema-generator")
+    implementation("org.jetbrains.kotlinx", "kotlinx-coroutines-jdk8", kotlinCoroutinesVersion)
+    testImplementation("org.slf4j", "slf4j-log4j12", "1.7.30")
```
smyrick, line 22: We can also add this version to the `gradle.properties` if we have the file

examples/ktor/src/main/kotlin/com/expediagroup/graphql/examples/ktor/GraphQLModule.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.ktor.routing.get
+import io.ktor.routing.post
+import io.ktor.routing.routing
+import java.lang.IllegalStateException
+
+fun Application.graphQLModule() {
+    install(Routing)
+
+    routing {
+        get("hello") {
```
smyrick, line 19: We can remove this route

examples/ktor/src/main/kotlin/com/expediagroup/graphql/examples/ktor/schema/HelloQueryService.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expediagroup.graphql.examples.ktor.schema
+
+class HelloQueryService {
```
smyrick, line 18: If the route is deleted we can delete the service too

smyrick (changes requested):

smyrick:
Let's also edit the `examples/README.md` file to list the new module

examples/ktor/build.gradle.kts
adamsz-lume: just spotted it was not needed as logback is used anyway

examples/ktor/src/main/kotlin/com/expediagroup/graphql/examples/ktor/GraphQLModule.kt
adamsz-lume: it's gone

examples/ktor/src/main/kotlin/com/expediagroup/graphql/examples/ktor/schema/HelloQueryService.kt
adamsz-lume: This service is part of the graphql schema per spark example. It is convenient to have it to make a quick query.

smyrick: Ah I see, I thought this was for the hello http route

smyrick (approved):

dariuszkuc:
Very nice!

PR #957, opened 2020-12-03
[build] deprioritize jcenter in favor of maven central

dariuszkuc:
See:
* https://github.com/gradle/gradle/issues/15406
* https://blog.autsoft.hu/a-confusing-dependency/


dariuszkuc:
We probably should enable https://docs.gradle.org/current/userguide/dependency_verification.html as well.

smyrick (approved):

PR #958, opened 2020-12-07
[docs] move server implementation examples under server directory

dariuszkuc:
### :pencil: Description

Updating examples directory structure to have consistent naming for example server integrations:

```
examples
|- client
|- federation
|- server
  |- ktor-server
  |- spark-server
  |- spring-server
```


docs/schema-generator/customizing-schemas/advanced-features.md
```diff
@@ -8,25 +8,24 @@ title: Advanced Features
 There are a couple ways you can add more types to the schema without having them be directly consumed by a type in your schema.
 This may be required for [Apollo Federation](federated/apollo-federation), or maybe adding other interface implementations that are not picked up.
 
+### `SchemaGenerator::generateSchema`
```
dariuszkuc, line 11: I accidentally pushed this doc fix in this PR as well -> @smyrick let me know if you want me to move this to separate commit

smyrick: If this PR is all doc ena example updates, it's fine to merge as one change.

docs/schema-generator/customizing-schemas/advanced-features.md
```diff
@@ -8,25 +8,24 @@ title: Advanced Features
...
+        additionalInputTypes = setOf(MyCustomInputObject::class.createType())
+)
+```
 
 ### `SchemaGenerator::addAdditionalTypesWithAnnotation`
 
 This method is protected so if you override the `SchemaGenerator` used you can call this method to add types that have a specific annotation.
 You can see how this is used in `graphql-kotlin-federation` as [an example](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGenerator.kt).
-
-### `SchemaGenerator::generateAdditionalTypes`
```
smyrick, line 17: We should still keep this section around though right? The code is still relevant

dariuszkuc: With the change to allow specifying additional types as arguments to `generateSchema` method we removed `open` modifier from both `addAdditionalTypesWithAnnotation` and `generateAdditionalTypes` so this is no longer valid.

smyrick (approved):

PR #960, opened 2020-12-08
[3.x.x] Add federation context and tracing support

smyrick:
### :pencil: Description

This adds a new `FederationGraphQLContext` and a `FederationGraphQLContextFactory` that are conditionally added if federation is enabled. Since they implement the base interfaces we already had we don't need to change the code in `ContextWebFilter` or the spring execution to use them differently. It will just be on federation users to implement the `FederationGraphQLContextFactory` now instead of the `GraphQLContextFactory` if they want tracing support.

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/418

### DIFF
--- a/graphql-kotlin-federation/build.gradle.kts
+++ b/graphql-kotlin-federation/build.gradle.kts
@@ -1,9 +1,11 @@
 description = "Federated GraphQL schema generator"
 
 val junitVersion: String by project
+val federationGraphQLJavaSupport: String by project
 
 dependencies {
     api(project(path = ":graphql-kotlin-schema-generator"))
+    implementation("com.apollographql.federation:federation-graphql-java-support:$federationGraphQLJavaSupport")
     testImplementation("org.junit.jupiter:junit-jupiter-params:$junitVersion")
 }
 
--- a/graphql-kotlin-federation/build.gradle.kts
+++ b/graphql-kotlin-federation/build.gradle.kts
@@ -5,7 +5,7 @@ val federationGraphQLJavaSupport: String by project
 
 dependencies {
     api(project(path = ":graphql-kotlin-schema-generator"))
-    implementation("com.apollographql.federation:federation-graphql-java-support:$federationGraphQLJavaSupport")
+    api("com.apollographql.federation:federation-graphql-java-support:$federationGraphQLJavaSupport")
     testImplementation("org.junit.jupiter:junit-jupiter-params:$junitVersion")
 }
 
--- a/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederationGraphQLContext.kt
+++ b/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederationGraphQLContext.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2020 Expedia, Inc
+ *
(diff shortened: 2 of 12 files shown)


graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederationGraphQLContext.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Apollo federation needs access to the HTTP headers to inspect if the
+ * request came from the Apollo Gateway. That means we need a special interface
+ * for the federation context.
+ */
+interface FederationGraphQLContext : GraphQLContext, HTTPRequestHeaders
+
+/**
+ * Can be used as a default [FederationGraphQLContext] if there is none provided.
+ */
+class EmptyFederationGraphQLContext : FederationGraphQLContext {
```
dariuszkuc, line 32: instead of empty should we favor the empty map instead? (so it will be consistent with https://github.com/ExpediaGroup/graphql-kotlin/pull/955)

graphql-kotlin-federation/build.gradle.kts
```diff
@@ -1,9 +1,11 @@
 description = "Federated GraphQL schema generator"
 
 val junitVersion: String by project
+val federationGraphQLJavaSupport: String by project
 
 dependencies {
     api(project(path = ":graphql-kotlin-schema-generator"))
+    implementation("com.apollographql.federation:federation-graphql-java-support:$federationGraphQLJavaSupport")
```
dariuszkuc, line 8: if it is required by anyone that pulls it in -> should this be `api` instead?

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/FederatedSchemaAutoConfiguration.kt
```diff
@@ -86,4 +89,21 @@ class FederatedSchemaAutoConfiguration {
 
         return schema
     }
+
+    /**
+     * Instrumentation is automatically added to the schema if it is registered as a spring component.
+     * This registers the federaiton tracing intrumentation for federated services.
```
dariuszkuc, line 95:
```suggestion
     * This registers the federation tracing instrumentation for federated services.
```

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/SchemaAutoConfiguration.kt
```diff
@@ -81,4 +83,8 @@ class SchemaAutoConfiguration {
 
         return schema
     }
+
+    @Bean
+    @ConditionalOnMissingBean
+    fun graphQLContextFactory(): GraphQLContextFactory<*> = EmptyContextFactory
```
smyrick, line 89: This will now only be used if `federation=false`

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederationGraphQLContext.kt
dariuszkuc: whoops my bad 👍

dariuszkuc (approved):

PR #962, opened 2020-12-08
Add federation context and tracing support

smyrick:
### :pencil: Description

This adds a new `FederationGraphQLContext` and a `FederationGraphQLContextFactory` that are conditionally added if federation is enabled. Since they implement the base interfaces we already had we don't need to change the code in `ContextWebFilter` or the spring execution to use them differently. It will just be on federation users to implement the `FederationGraphQLContextFactory` now instead of the `GraphQLContextFactory` if they want tracing support.

### :link: Related Issues
Support in 3.x.x added here: https://github.com/ExpediaGroup/graphql-kotlin/pull/960


graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/GraphQLContextFactory.kt
```diff
@@ -37,10 +39,23 @@ interface GraphQLContextFactory<out T : GraphQLContext> {
     suspend fun generateContext(request: ServerHttpRequest, response: ServerHttpResponse): T
 }
 
+/**
+ * Factory that generates Federation GraphQL context.
+ */
+interface FederationGraphQLContextFactory<out T : FederationGraphQLContext> : GraphQLContextFactory<T>
```
dariuszkuc, line 45:
```suggestion
interface FederatedGraphQLContextFactory<out T : FederationGraphQLContext> : GraphQLContextFactory<T>
```

graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/execution/FederationGraphQLContext.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.apollographql.federation.graphqljava.tracing.HTTPRequestHeaders
+import com.expediagroup.graphql.execution.DefaultGraphQLContext
+import com.expediagroup.graphql.execution.GraphQLContext
+
+/**
+ * Apollo federation needs access to the HTTP headers to inspect if the
+ * request came from the Apollo Gateway. That means we need a special interface
+ * for the federation context.
+ */
+interface FederationGraphQLContext : GraphQLContext, HTTPRequestHeaders
```
dariuszkuc, line 28:
```suggestion
interface FederatedGraphQLContext : GraphQLContext, HTTPRequestHeaders
```

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/FederatedSchemaAutoConfiguration.kt
```diff
@@ -86,4 +89,20 @@ class FederatedSchemaAutoConfiguration {
 
         return schema
     }
+
+    /**
+     * Instrumentation is automatically added to the schema if it is registered as a spring component.
+     * This registers the federation tracing instrumentation for federated services.
+     */
+    @Bean
+    @ConditionalOnMissingBean
```
dariuszkuc, line 98: should this be conditional on property instead?

smyrick: It is conditional, are you asking if we should not make it conditional? I thought that if someone wanted to provide a different tracing instrumentation with the debug options enabled they could do so by overriding the bean

dariuszkuc: I was thinking more what if I don't want this tracing to be on? i.e. bean is created if no `FederatedTracingInstrumentation` bean exist but what if I don't use Apollo and have my own custom instrumentation instead?

smyrick:
This doesn't work with the latest version, we need to update the library to graphql-java 16

smyrick:
This is now dependent on the library updating the graphql-java 16: https://github.com/apollographql/federation-jvm/pull/84

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/GraphQLContextFactory.kt
```diff
@@ -37,10 +38,29 @@ interface GraphQLContextFactory<out T : GraphQLContext> {
...
 /**
  * Default context factory that generates GraphQL context with empty concurrent map that can store any elements.
  */
 internal object DefaultContextFactory : GraphQLContextFactory<DefaultGraphQLContext> {
 
     override suspend fun generateContext(request: ServerHttpRequest, response: ServerHttpResponse) = DefaultGraphQLContext()
 }
+
+/**
+ * Default federation context factory that generates empty GraphQL context but still reads the http headers
```
dariuszkuc, line 55:
```suggestion
 * Default federation context factory that generates default GraphQL context but still reads the http headers
```

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLContextFactory.kt
```diff
@@ -30,10 +29,3 @@ interface GraphQLContextFactory<out Context : GraphQLContext, Request> {
      */
     fun generateContext(request: Request): Context?
 }
-
-/**
- * Default context factory that generates GraphQL context with empty concurrent map that can store any elements.
- */
-class DefaultGraphQLContextFactory<T> : GraphQLContextFactory<DefaultGraphQLContext, T> {
```
smyrick, line 37: This class was not used anywhere. Each server can provide their own default implmentation

servers/graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/server/spring/FederationConfigurationTest.kt
```diff
@@ -79,6 +87,38 @@ class FederationConfigurationTest {
...
+                "graphql.federation.tracing.enabled=true"
+            )
+            .run { ctx ->
+                assertThat(ctx).hasSingleBean(GraphQL::class.java)
+                assertThat(ctx).hasSingleBean(FederatedTracingInstrumentation::class.java)
+
+                val graphql = ctx.getBean(GraphQL::class.java)
+                val federatedContext = MockFederatedGraphQLContext()
+
+                // TODO: This is failing until federation-jvm updates to graphql-java 16
```
smyrick, line 105: This is going to fail until the federation-jvm libraries is updated: https://github.com/apollographql/federation-jvm/pull/84

smyrick:
Now ready for review

gradle.properties
```diff
@@ -28,6 +28,7 @@ reactorExtensionsVersion = 1.1.2
 slf4jVersion = 1.7.30
 springBootVersion = 2.4.2
 springVersion = 5.3.3
+federationGraphQLJavaSupport = 0.6.0
```
dariuszkuc, line 31: All other props for version have version suffix -> lets add it for consistency (we can drop the support suffix)

plugins/graphql-kotlin-gradle-plugin/build.gradle.kts
```diff
@@ -9,6 +9,10 @@ val kotlinCoroutinesVersion: String by project
 val wireMockVersion: String by project
 val mustacheVersion: String by project
 
+repositories {
+    jcenter()
```
dariuszkuc, line 13: its already defined in the root build file

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLConfigurationProperties.kt
```diff
@@ -18,12 +18,15 @@ package com.expediagroup.graphql.server.spring
...
 import org.springframework.boot.context.properties.ConstructorBinding
+import org.springframework.core.Ordered
+import org.springframework.core.annotation.Order
 
 /**
  * [ConfigurationProperties] bean that defines supported GraphQL configuration options.
  */
 @ConstructorBinding
 @ConfigurationProperties("graphql")
+@Order(Ordered.HIGHEST_PRECEDENCE)
```
dariuszkuc, line 29: does this have any effect?

dariuszkuc (commented):
Since we only use the tracing part of the `federation-jvm` maybe we should shade it instead? this would drop the need for explicit dependency on jcenter for maven builds (which IMHO is pretty bad)

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLConfigurationProperties.kt
smyrick: I think this got pulled in from some local debugging I was doing, I will remove

plugins/graphql-kotlin-gradle-plugin/build.gradle.kts
smyrick: Removing..

smyrick:
@dariuszkuc Are you talking about using this to relocate the jar? https://imperceptiblethoughts.com/shadow/

dariuszkuc:
> @dariuszkuc Are you talking about using this to relocate the jar? https://imperceptiblethoughts.com/shadow/

Yeah I think thats the most popular shade plugin (based on the ⭐ ). Instead of relocating the JAR I was thinking that during the build we would extract the needed bits and repackage it in GraphQL Kotlin federation module directly so end users wouldn't need to point to JCenter. Since [federation-jvm](https://github.com/apollographql/federation-jvm/blob/master/LICENSE) is using `MIT License` it should be fine as long as we include all the necessary information.

smyrick:
Now that jcenter is deprecated, the apollo lib will have to move soon: https://github.com/ExpediaGroup/graphql-kotlin/issues/1033

We can either check this in with the requirement of jcenter or wait but I think we will not take up the work to shadow the dependency anymore

smyrick:
I am not sure why all these changes are not failing locally for me but hopefully soon we won't have to deal with the jcenter requirement

dariuszkuc:
@smyrick we should wait for Apollo to push to MavenCentral and drop all JCenter configs

PR #963, opened 2020-12-08
[3.x.x] Fix some federated tracing issues

smyrick:
### :pencil: Description
Rename the classes before we published from `Federation*` to `Federated*` to match patterns of other classes. We have not published yet so it is not a breaking change.

### :link: Related Issues
Initial PR https://github.com/ExpediaGroup/graphql-kotlin/pull/960


dariuszkuc (changes requested):
Need to update https://github.com/ExpediaGroup/graphql-kotlin/blob/3.x.x/docs/spring-server/spring-graphql-context.md

dariuszkuc (approved):

graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/FederatedSchemaAutoConfiguration.kt
```diff
@@ -95,8 +95,9 @@ class FederatedSchemaAutoConfiguration {
      * This registers the federation tracing instrumentation for federated services.
      */
     @Bean
-    @ConditionalOnMissingBean
-    fun federatedTracing(): FederatedTracingInstrumentation = FederatedTracingInstrumentation()
+    @ConditionalOnProperty(value = ["graphql.federation.tracing.enabled"], havingValue = "true")
```
dariuszkuc, line 98: should we add `match if missing` as well?

smyrick:
You can now see tracing locally if you pass in the federation headers:

smyrick:
This one is ready to go now, the 4.x.x changes here are blocked by the graphql-java 16 update: https://github.com/ExpediaGroup/graphql-kotlin/pull/962

dariuszkuc (changes requested):
can you update https://github.com/ExpediaGroup/graphql-kotlin/blob/3.x.x/docs/spring-server/spring-beans.md and https://github.com/ExpediaGroup/graphql-kotlin/blob/3.x.x/docs/spring-server/spring-properties.md as well?

dariuszkuc:
I think we should also add a blurb about the federated context to https://github.com/ExpediaGroup/graphql-kotlin/blob/3.x.x/docs/spring-server/spring-graphql-context.md (as currently that info is under federated tracing info which may not be obvious)

dariuszkuc (approved):

PR #981, opened 2020-12-27
Support default Kotlin values

smyrick:
### :pencil: Description
Add support for Kotlin default values. By calling functions with the `callBy` methods instead we can pass a map of `KParameters`. Then if one of the params is optional and it is not sent a value from GraphQL input we can exclude that parameter from the map, which mean the default value will be used.

Kotlin
```kotlin
fun print(message: String? = "hello"): String? = message
```

GraphQL
```graphql
query PrintMessages {
    first: print(message = "foo") # foo
    second: print(message = null) # null
    third: print # hello
}
```

This info will not be [conveyed in the schema](https://graphql.org/learn/schema/#arguments), but it simplifies the logic needed for optional arguments. If you still need to have logic for the default values I recommend using the `OptionalInput` wrapper, since you will still not be able to determine the difference when the client didn't pass a value or it passed in the same value as the default.

This is a breaking change because we are changing the function signature that is open for `FunctionDataFetcher`

### :link: Related Issues
Fixes https://github.com/ExpediaGroup/graphql-kotlin/issues/53

### DIFF
--- a/docs/schema-generator/execution/optional-undefined-arguments.md
+++ b/docs/schema-generator/execution/optional-undefined-arguments.md
@@ -7,7 +7,7 @@ In the GraphQL world, input types can be optional which means that the client ca
 
 * Not specify a value at all
 * Send null explicitly
-* Send the non-null type
+* Send a non-null value
 
 This is in contrast with the JVM world where objects can either have some specific value or don't have any value (i.e.
 are `null`). As a result, when using default serialization logic it is not possible to distinguish between missing/unspecified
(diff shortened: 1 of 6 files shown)


smyrick:
TODO - update docs

dariuszkuc (approved):

rowi1de:
Is this change intended: https://kotlinlang.slack.com/archives/CQLNT7B29/p1616235196005400

smyrick:
I could make this change: https://github.com/ExpediaGroup/graphql-kotlin/compare/master...smyrick:null-on-missing-args?expand=1

But then this unit test no longer works and neither will default values.
https://github.com/smyrick/graphql-kotlin/blob/c5b8fdc3739930f712a114b9abc15dee8fc[…]diagroup/graphql/generator/execution/FunctionDataFetcherTest.kt

Since we can’t inspect if a `KParameter` has a default value, we can only support one use case or the other.
Support Kotlin default values but require them for nullable input
Do not support Kotlin default values and always pass null for input if argument is missing in request

Issue #705, opened 2020-05-13
[plugin] generate schema artifact task/mojo


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
`graphql-kotlin-schema-generator` follows code first approach for building GraphQL schema where schema is built at runtime based on the source code. There are many use cases where it is beneficial to have schema artifact (e.g. it could be used to detect breaking changes, schema stitching etc).

**Describe the solution you'd like**
New Gradle task/Maven MOJO that can generate GraphQL schema artifact during build time.

**Describe alternatives you've considered**
Using Gradle it is possible to create custom tasks to generate the schema but it implies every service would need to re-implement it themselves. This is not easily achievable in Maven.

dariuszkuc:
This turned out to be slightly more complex than expected so won't be included in `3.0.0` release. While overall logic is pretty straightforward there is an underlying problem that schema generation requires an instance of `SchemaGeneratorHooks`. 

* Maven follows static configuration so in order to instantiate custom hooks we would generally have to fallback to use reflections (e.g. [GenerateSchemaAbstractMojo](https://github.com/dariuszkuc/graphql-kotlin/blob/plugin_generate_schema/plugins/graphql-kotlin-maven-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/maven/GenerateSchemaAbstractMojo.kt#L85)). As a result we would only be able to support hooks that have default constructor OR rely on some static factory method that would mock it out for us. 
* Gradle technically could allow creating new hooks instance directly from script but only if custom hooks class is in some dependent jar that could be added to the buildscript path. If custom hook would be part of the same project then we would still have to use reflections same as in Maven case.

The above shows the problem with current hooks configuration - they contain both build configuration for generating the schema as well as additional runtime configuration (e.g. runtime wiring, federated type resolvers, custom data fetchers, etc). I think we should revisit our configuration and see whether we could refactor it to extract static build configuration that could be referenced by the plugins.

WIP: https://github.com/dariuszkuc/graphql-kotlin/tree/plugin_generate_schema

dariuszkuc:
Resolved by: 

* SDL generator PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/990
* Gradle task PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/993
* Maven mojo PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/994

PR #982, opened 2020-12-29
[POC] new generateSDL Gradle task

dariuszkuc:
Creating a new generic `graphqlGenerateSDL` Gradle task that can generate GraphQL schema in SDL format at built time. New task utilizes Gradle [Worker API](https://docs.gradle.org/nightly/userguide/worker_api.html) to isolate the plugin classpath from the build classpath.

In order to generate the SDL we rely on the [ServiceLoader](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html) mechanism to locate an instance of a `SchemaGeneratorHooksProvider` that is responsible for creating an instance of the hooks (default no-op hooks and federated hooks are provided by the `graphql-kotlin` lib).

```kotlin
interface SchemaGeneratorHooksProvider {
    fun create(): SchemaGeneratorHooks
}
```

With the change to `Worker API` we can update our plugin directory structure to the following

```
plugins
|- graphql-kotlin-gradle-plugin
|- graphql-kotlin-maven-plugin
|- client
  |- graphql-kotlin-client-generator // current plugin-core
|- schema
  |- graphql-kotlin-default-hooks-provider
  |- graphql-kotlin-sdl-printer
  |- graphql-kotlin-hooks-provider
  |- graphql-kotlin-federated-hooks-provider
```

---
Within corresponding Maven MOJO I'm thinking we would invoke the SDL generation process (that still uses the service loader mechanism and explicit plugin dependencies for the hooks providers) but we would need to explicitly provide target classpath.

---
resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/705


smyrick:
Do we need separate packages for all the different hook providers? Can we just have one package for those and one for the  schema SDL printer plugin?

graphql-kotlin-types/src/main/kotlin/com/expediagroup/graphql/operations/Query.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.expediagroup.graphql.spring.operations
+package com.expediagroup.graphql.operations
```
smyrick, line 17: Should we create a separate `com.expediagroup.graphql.server` package as this marker interfaces are only used for the server?

dariuszkuc: 👍 sure, I'll move this change into separate PR.

dariuszkuc:
> Do we need separate packages for all the different hook providers? Can we just have one package for those and one for the schema SDL printer plugin?

We can but I think it complicates things. I kept it simple - i.e. single jar = single provider so we 

1) Don't need to add some ordering logic to determine which provider should be picked. (assuming single provider jar that has interface + default implementations) 
2) Don't need custom logic to determine whether to generate federated or non federated schema - its determined by the provided hook instance. We could add additional parameter for switching between federation/default but that flag would be ignored for custom providers.

dariuszkuc:
Split up into:
* SDL generator PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/990
* Gradle task PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/993
* Maven mojo PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/994

PR #988, opened 2021-01-05
Common graphql-kotlin-server

smyrick:
### :pencil: Description
Create a new `graphql-kotlin-server` module that can be used by Ktor and Spark. In general the different server implementations just need to do the following:

* Create a `GraphQL` object that can execute operations
* Convert the HTTP request into a `GraphQLRequest`
* Send the `GraphQLResponse` via the HTTP response calls

Everything else in terms of execution is generic and can be moved to a common library or overridden if specific logic is needed. 

This is a breaking change because of the following:

* The class `SimpleKotlinGraphQLError` moved to new module and renamed `KotlinGraphQLError`
* The interface `DataLoaderRegistryFactory` moved to new module
* The interface `QueryHandler` moved and renamed to `GraphQLRequestHandler`
* The extension functions for `GraphQLRequest` and `GraphQLResponse` moved to new module
* `SimpleSubscriptionHandler` is now `SpringGraphQLSubscriptionHandler`
* New interface `GraphQLContextFactory` in new module replaces the old one that was only in spring
* `DataLoaderRegistryFactory` moved to new module

### :link: Related Issues
N\A


smyrick:
An easy way to also see the changes is to look at the new files in the module

Screen Shot 2021-01-05 at 12 35 23 AM

dariuszkuc:
We got `clients`, `docs`, `plugins` so maybe change `server` to `servers` as well?

examples/server/ktor-server/build.gradle.kts
```diff
@@ -14,6 +14,7 @@ application {
 
 val kotlinCoroutinesVersion: String by project
 dependencies {
+    implementation("com.expediagroup", "graphql-kotlin-server")
```
dariuszkuc, line 17: lets follow consistent pattern -> either single string or change to explicit parts but lets not mix it

server/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/DataLoaderRegistryFactory.kt
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Factory used to generate [DataLoaderRegistry] per GraphQL execution.
+ */
+interface DataLoaderRegistryFactory {
+
+    /**
+     * Generate [DataLoaderRegistry] to be used for GraphQL request execution.
+     */
+    fun generate(): DataLoaderRegistry?
```
dariuszkuc, line 29: should this generate not-nullable registry?

smyrick:
These are now the update files in the new `graphql-kotlin-server` package

Screen Shot 2021-01-08 at 6 55 33 PM

smyrick:
Subscription example still running

Screen Shot 2021-01-08 at 7 26 41 PM

smyrick:
At this point I am ready to merge this in, however I think there is still some more work we can do as a part 2 but this PR is big enough already

docs/getting-started.md
```diff
@@ -65,9 +64,9 @@ Using `graphql-kotlin-federation`, you can generate an [Apollo Federation](https
 See the docs in [Apollo Federation](./federated/apollo-federation.md).
 
 ## Running a Server
-`graphql-kotlin-spring-server` is a combination of the schema generator, fedeation, and server libraries. If you are looking to run a GraphQL server, this is the place to start.
+`graphql-kotlin-server` is a combination of the schema generator, fedeation, and server libraries. If you are looking to run a GraphQL server, this is the place to start.
```
dariuszkuc, line 67:
```suggestion
`graphql-kotlin-server` is a combination of the schema generator, federation, and server libraries. If you are looking to run a GraphQL server, this is the place to start.
```

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLServer.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Default execution logic for handling a [Request] and returning a [GraphQLResponse].
+     *
+     * If null is returned, that indicates a problem parsing the request or context.
+     * If the request is valid, a [GraphQLResponse] should always be retuned.
+     * In the case of errors or exceptions, return a response with [GraphQLResponse.errors] populated.
+     * If you need custom logic inside this method you can override this class or choose not to use it.
+     */
+    open suspend fun getResponse(request: Request): GraphQLResponse<*>? {
```
dariuszkuc, line 39: wondering whether we should rename this method to `execute` to match naming in request handler `executeRequest` and  `graphql-java` `GraphQL#execute`/`GraphQL#executeAsync`

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionHooks.kt
```diff
@@ -25,30 +25,37 @@ import reactor.core.publisher.Mono
...
+     * Called when the client executes a GraphQL operation.
+     * The context can not be updated here, it is read only.
      */
     fun onOperation(
         operationMessage: SubscriptionOperationMessage,
         session: WebSocketSession,
-        graphQLContext: Any?
-    ): Mono<Unit> = mono { }
+        graphQLContext: GraphQLContext
+    ): Unit = Unit
```
dariuszkuc, line 48: why the change from `Mono` to `Unit`?

dariuszkuc (approved):
LGTM

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/spring/execution/ApolloSubscriptionHooks.kt
smyrick: Because it drastically simplifies the subscription logic. Since we couldn't return anything from these hooks anyway we can just make them `Unit` functions that just get called in the life cycle.

smyrick: By not requiring mono we dont need to have all the combiners in the subscription handler to combine the hooks monos with the real data fluxs

dariuszkuc (approved):

PR #990, opened 2021-01-08
[plugins] create sdl-generator module that can generate graphql schema

dariuszkuc:
### :pencil: Description

`graphql-kotlin-sdl-generator` module can be used to generate GraphQL schema build artifact directly from your source code. Module scans the classpath for classes implementing `graphql-kotlin-types` marker interfaces (`Query`, `Mutation` and `Subscription`) and then uses that information to generate a schema using `graphql-kotlin-schema-generator`. By default, this module will generate the schema using `NoopSchemaGeneratorHooks`. If your project uses custom hooks or needs to generate the federated GraphQL schema, you will need to provide an instance of `SchemaGeneratorHooksProvider` that will be used to create an instance of your custom hooks.

`SchemaGeneratorHooksProvider` is a service provider interface that exposes a single `hooks` method to generate an instance of `SchemaGeneratorHooks` that will be used to generate your schema. By utilizing Java [ServiceLoader](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html) we can dynamically load custom providers directly from the classpath. Service provider can be provided as part of target project sources, included inside of one of the project dependencies or through explicitly provided artifact. Since we need to be able to deterministically choose a single hooks provider, generation of schema will fail if there are multiple providers on the classpath.

### :link: Related Issues

Splitting up https://github.com/ExpediaGroup/graphql-kotlin/pull/982/files into smaller PRs

### DIFF
--- a/plugins/schema/graphql-kotlin-federated-hooks-provider/README.md
+++ b/plugins/schema/graphql-kotlin-federated-hooks-provider/README.md
@@ -1,4 +1,4 @@
 # GraphQL Kotlin Federated Hooks Provider
 
 This module provides default implementation of `SchemaGeneratorHooksProvider` that can be used to generate federated GraphQL
-schema in SDL format.
+schema in SDL format. This provider creates a default instance of `FederatedSchemaGeneratorHooks`.
--- a/plugins/schema/graphql-kotlin-federated-hooks-provider/README.md
+++ b/plugins/schema/graphql-kotlin-federated-hooks-provider/README.md
@@ -1,0 +1,4 @@
+# GraphQL Kotlin Federated Hooks Provider
+
+This module provides default implementation of `SchemaGeneratorHooksProvider` that can be used to generate federated GraphQL
+schema in SDL format.
--- a/plugins/schema/graphql-kotlin-sdl-generator/src/main/kotlin/com/expediagroup/graphql/plugin/schema/GenerateSDL.kt
+++ b/plugins/schema/graphql-kotlin-sdl-generator/src/main/kotlin/com/expediagroup/graphql/plugin/schema/GenerateSDL.kt
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2021 Expedia, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.plugin.schema
+
+import com.expediagroup.graphql.SchemaGeneratorConfig
+import com.expediagroup.graphql.TopLevelObject
+import com.expediagroup.graphql.extensions.print
+import com.expediagroup.graphql.federation.FederatedSchemaGeneratorConfig
+import com.expediagroup.graphql.federation.FederatedSchemaGeneratorHooks
+import com.expediagroup.graphql.federation.toFederatedSchema
+import com.expediagroup.graphql.hooks.NoopSchemaGeneratorHooks
+import com.expediagroup.graphql.server.operations.Mutation
+import com.expediagroup.graphql.server.operations.Query
+import com.expediagroup.graphql.server.operations.Subscription
+import com.expediagroup.graphql.plugin.schema.hooks.SchemaGeneratorHooksProvider
+import com.expediagroup.graphql.toSchema
+import io.github.classgraph.ClassGraph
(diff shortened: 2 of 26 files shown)


plugins/schema/graphql-kotlin-federated-hooks-provider/README.md
```diff
@@ -0,0 +1,4 @@
+# GraphQL Kotlin Federated Hooks Provider
+
+This module provides default implementation of `SchemaGeneratorHooksProvider` that can be used to generate federated GraphQL
```
smyrick, line 3: Do we need to call it `federated-hooks-provider`? How about just `hooks-provider`. Choosing regular or federated hooks is just the internal logic and you might not get federated hooks

dariuszkuc:
In current implementation `sdl-generator` module contains both the service provider interface as well as the logic to generate the schema using service loader. When implementing custom hooks provider we need to add a dependency on a module that contains the provider interface -> I'm wondering whether it would make sense to separate it out to another module?

I think it might be a little bit cleaner if we separate it out as 

* `hooks-provider` - contains just the service provider interface
* `sdl-generator` - contains just the logic to generate the schema using service loader

@smyrick thoughts?

smyrick:
> I think it might be a little bit cleaner if we separate it out as
> 
> * `hooks-provider` - contains just the service provider interface
> * `sdl-generator` - contains just the logic to generate the schema using service loader

Separation may not be needed here unless there is something we can do to make the plugin code smaller in the final package, similar to what was reported here: https://github.com/ExpediaGroup/graphql-kotlin/pull/974

If we still need to pull in the dependency any way maybe we should just keep it all as one. I don't see any future code where we would need to get the hooks.

Would separating make it easier to create a Maven plugin as well?

dariuszkuc:
From the plugin perspective there is no difference -> it needs both the interface (so it can load service providers from classpath) and actual generation logic. As per https://github.com/ExpediaGroup/graphql-kotlin/pull/982 we will be using Gradle WorkerAPI to isolate plugin classpath from buildscript classpath.

From the custom provider perspective I think it might be beneficial to split it up - its a clearer intent behind adding dependency and dependency has fewer transitive dependencies to pull in.

e.g. from our default federated hooks provider (https://github.com/dariuszkuc/graphql-kotlin/blob/sdl_generator/plugins/schema/graphql-kotlin-federated-hooks-provider/build.gradle.kts#L5)

```kotlin
// current
implementation(project(":graphql-kotlin-sdl-generator"))

// vs
implementation(project(":graphql-kotlin-hooks-provider"))
```

By looking at the project dependencies I think having artifact named `hooks-provider` as a dependency is more meaningful than `sdl-generator`. From the classpath perspective it would also be a little bit simpler on transitive dependencies*

```kotlin
// current sdl-generator
implementation(project(path = ":graphql-kotlin-types"))
implementation(project(path = ":graphql-kotlin-schema-generator"))
implementation(project(path = ":graphql-kotlin-federation"))
implementation("io.github.classgraph:classgraph:$classGraphVersion")
implementation("org.slf4j:slf4j-api:$slf4jVersion")

// vs hooks-provider
implementation(project(path = ":graphql-kotlin-schema-generator"))
```

*end result of a project transitive dependencies would be pretty much the same, as in order to generate the SDL, underlying projects would still need to depend on pretty much the same dependencies as `sdl-generator`...

smyrick:
You have had more change to work with all this code so if you think it is more beneficial to split it up I say lets do it.

As for my opinion based on everything you described it sounds like we should split it up to.

plugins/schema/graphql-kotlin-sdl-generator/src/main/kotlin/com/expediagroup/graphql/plugin/schema/GenerateSDL.kt
```diff
@@ -0,0 +1,110 @@
...
+    val scanResult = ClassGraph()
+        .enableAllInfo()
+        .acceptPackages(*supportedPackages.toTypedArray())
+        .scan()
+
+    val queries = findTopLevelObjects(scanResult, Query::class.java)
+    val mutations = findTopLevelObjects(scanResult, Mutation::class.java)
+    val subscriptions = findTopLevelObjects(scanResult, Subscription::class.java)
+
+    // TODO support top level name overrides?
```
smyrick, line 70: We support the name overrides in the generator but I am wondering if we should just drop support for those all together. It should be common practice now to have the query object just called `Query`

smyrick: The alternative is that we move the top level name override to the hooks but that doesn't seem like the best approach to me

smyrick (approved):

plugins/schema/graphql-kotlin-sdl-generator/src/main/kotlin/com/expediagroup/graphql/plugin/schema/GenerateSDL.kt
dariuszkuc: Thats why i did not implement this yet :) - we can pass those values from plugin task/mojo directly. Technically renaming those top fields is valid per spec but yeah unsure how often this is used.

PR #999, opened 2021-01-09
Move spring-server to com.expediagroup.graphql.server.spring

smyrick:
### :pencil: Description
Move the spring server package to match the nested structure of the project and to also clarify spring client from spring server

### :link: Related Issues
Part 2 of the big refactor: https://github.com/ExpediaGroup/graphql-kotlin/pull/988


dariuszkuc (commented):
I think you might want to revert on the `examples/client` changes as those projects won't compile (they have a dependency on older released version of libs). They are not part of composite build as client examples require running server before they can be built and this is not something that we can easily do in Gradle. Wondering whether we could run those builds in Docker (using compose).

dariuszkuc (approved):
Changes look fine. Just wondering whether we should pro-actively rename the `graphql-kotlin-spring-server` to `graphql-kotlin-spring-webflux-server` (which I assume would also imply updating packages).

smyrick:
I think we can keep the name around so we don't have to change to much on the users.

Because if we wanted to ever support Web MVC there could be code reuse for the config and GraphQL beans, it is just the specific routers and request handler code

PR #1014, opened 2021-01-14
Scope package of examples

smyrick:
### :pencil: Description
Move each example app into its own separate package path. The examples shouldn't clash since they are their own individual modules, but to keep consistent with the library and help for easier refactoring, we can move each own to it's own package

### :link: Related Issues
N\A


examples/federation/base-app/src/main/kotlin/com/expediagroup/graphql/examples/federation/base/schema/Widget.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.expediagroup.graphql.examples.model
+package com.expediagroup.graphql.examples.federation.base.schema
```
dariuszkuc, line 17: `schema` package implies (at least to me) something related to either loading or modifying the schema, this package holds data model that is exposed in the schema but i guess as long as we use consistent packages then it should be fine

dariuszkuc (approved):

PR #1016, opened 2021-01-15
[examples] use proper junit5 method order

dariuszkuc:


smyrick (approved):

smyrick:
And if course it still failed...

I don't know at that m this point. It is still executing subscription code.

We may need to add more debug logs to the test and server code

PR #1019, opened 2021-01-17
[server] support batch requests

dariuszkuc:
### :pencil: Description

Add support for handling both individual GraphQL requests (current) as well as batch requests (new) that specify a list of `GraphQLRequest`s in their HTTP body. Single context is created for batch requests and each one of the requests has to provide valid response.

### :link: Related Issues

N/A

### DIFF
--- a/servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLServerRequest.kt
+++ b/servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLServerRequest.kt
@@ -16,11 +16,19 @@
 
 package com.expediagroup.graphql.server.execution
 
-/**
- * A generic server interface that handles parsing the specific server implementation request to a [GraphQLServerRequest].
- * If the request is not valid return null.
- */
-interface GraphQLRequestParser<Request> {
+import com.expediagroup.graphql.types.GraphQLRequest
 
-    suspend fun parseRequest(request: Request): GraphQLServerRequest<*>?
-}
+/**
+ * GraphQL server request abstraction that provides a convenient way to handle both single and batch requests.
+ */
+sealed class GraphQLServerRequest<T : Any>(val request: T)
+
+/**
+ * Wrapper that holds single GraphQLRequest to be processed within an HTTP request.
+ */
+class GraphQLSingleRequest(graphQLRequest: GraphQLRequest) : GraphQLServerRequest<GraphQLRequest>(graphQLRequest)
+
+/**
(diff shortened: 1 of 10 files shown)


dariuszkuc:
Running against example Spring app

// request
```json
[
    {
        "query": "query AnimalQuery {\n  animal(type: CAT) {\n    sound\n    ... on Cat {\n      ignoreEveryone\n    }\n    ... on Dog {\n      doSomethingUseful\n    }\n  }\n}",
        "operationName": "AnimalQuery"
    },
    {
        "query": "query RandomQuery { generateRandomId generateRandomUUID generateNullableNumber }",
        "operationName": "RandomQuery"
    }
]
```

// response
```json
[
    {
        "data": {
            "animal": {
                "sound": "meow",
                "ignoreEveryone": "ignore everyone"
            }
        }
    },
    {
        "data": {
            "generateRandomId": "093927e9-385f-441d-8d74-2e731ccd6085",
            "generateRandomUUID": "434e9bf7-a3fc-4aaf-a88b-df020d158dd6",
            "generateNullableNumber": 0
        }
    }
]
```

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLServerRequest.kt
```diff
@@ -0,0 +1,9 @@
+package com.expediagroup.graphql.server.execution
+
+import com.expediagroup.graphql.types.GraphQLRequest
+
+sealed class GraphQLServerRequest<T: Any>(val request: T)
+
+class GraphQLSingleRequest(graphQLRequest: GraphQLRequest) : GraphQLServerRequest<GraphQLRequest>(graphQLRequest)
```
smyrick, line 7: Should we move these to `graphql-kotlin-types` to be used by clients as well?

dariuszkuc: I'm still playing around with the client batching but I don't think it will make much sense to do - I used sealed class so I can have single return type from the request parser method. While we could potentially follow the same on the client, I'm wondering if there is any benefit of using the extra wrapper over just

```kotlin
client.execute(request) // single
client.execute(listOf(request1, request2)) // batch

// or a vararg
client.execute(request1, request2)
```

Currently we also do not use the `GraphQLRequest` in the clients due to the difference in `variables` field definition - in server we want to access those as a generic map, on server we want to send `Any` (so it will serialize to a generic map).

smyrick (approved):

PR #1022, opened 2021-01-21
Update spring server example

valoricDe:
### :pencil: Description
Hi, I had to update the spring example to make it work with 4.0.0
Are these changes correct?


dariuszkuc:
Hello 👋 
Examples* are part of a composite build that always targets the latest `-SNAPSHOT` changes from the main (`master`) branch. Recently we've done some cleanup on the namespacing of the modules which might not have been released yet.

*currently only client examples target older released version of lib and should be updated

Issue #1023, opened 2021-01-21
"%" in schema comments is not escaped, causing KotlinPoet failure.


bsorbo:
**Library Version**
3.6.8

**Describe the bug**
Now that we understand the (presumable) bug, this is easy to work around, but filing the issue since the problem wasn't immediately obvious in large auto-generated schemas.

Presently when a "%" is included in any schema comment, that will be passed as-is to KotlinPoet. KotlinPoet uses "%" for various formatters within the provided text string, causing parsing to fail. An example stack is:

```
Exception in thread "main" java.lang.IllegalArgumentException: index 1 for '% ' not in range (received 0 arguments)
        at com.squareup.kotlinpoet.CodeBlock$Builder.add(CodeBlock.kt:310)
        at com.squareup.kotlinpoet.TypeSpec$Builder.addKdoc(TypeSpec.kt:473)
        at com.expediagroup.graphql.plugin.generator.types.GenerateGraphQLEnumTypeSpecKt.generateGraphQLEnumTypeSpec(generateGraphQLEnumTypeSpec.kt:38)
        at com.expediagroup.graphql.plugin.generator.types.GenerateTypeNameKt.generateCustomClassName(generateTypeName.kt:89)
        at com.expediagroup.graphql.plugin.generator.types.GenerateTypeNameKt.generateTypeName(generateTypeName.kt:59)
        at com.expediagroup.graphql.plugin.generator.types.GeneratePropertySpecsKt.generatePropertySpecs(generatePropertySpecs.kt:45)
        at com.expediagroup.graphql.plugin.generator.types.GeneratePropertySpecsKt.generatePropertySpecs$default(generatePropertySpecs.kt:37)
        at com.expediagroup.graphql.plugin.generator.types.GenerateGraphQLObjectTypeSpecKt.generateGraphQLObjectTypeSpec(generateGraphQLObjectTypeSpec.kt:49)
        at com.expediagroup.graphql.plugin.generator.types.GenerateGraphQLObjectTypeSpecKt.generateGraphQLObjectTypeSpec$default(generateGraphQLObjectTypeSpec.kt:35)
        at com.expediagroup.graphql.plugin.generator.types.GenerateTypeNameKt.generateCustomClassName(generateTypeName.kt:87)
        at com.expediagroup.graphql.plugin.generator.types.GenerateTypeNameKt.generateTypeName(generateTypeName.kt:59)
[...]

PR #1024, opened 2021-01-22
fix passing doc string as format parameter

brennantaylor:
### :pencil: Description

The kotlin poet api takes as first argument a format parameter for `addKdoc()`.
So comments with percent signs would cause the format argument parser to error.

### :link: Related Issues

https://github.com/ExpediaGroup/graphql-kotlin/issues/1023


plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generateGraphQLEnumTypeSpec.kt
```diff
@@ -51,7 +51,7 @@ internal fun generateGraphQLEnumTypeSpec(context: GraphQLClientGeneratorContext,
     }
 
     val unkownTypeSpec = TypeSpec.anonymousClassBuilder()
-        .addKdoc("This is a default enum value that will be used when attempting to deserialize unknown value.")
+        .addKdoc("%L", "This is a default enum value that will be used when attempting to deserialize unknown value.")
```
dariuszkuc, line 54: since the passed in text is a constant without any special chars we technically don't need the extra formatting here

dariuszkuc (approved):

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generateGraphQLEnumTypeSpec.kt
brennantaylor: Fair enough. I figured the consistency was more important.

PR #1030, opened 2021-01-30
[client] Batch query support

dariuszkuc:
### :pencil: Description

Refactor of GraphQL client to support batch operations.

Instead of injecting an instance of GraphQL client to a generated operation class, generated classes are now simple POJOs that represent underlying GraphQL request. This change breaks the tight coupling between generated classes and client implementation and allows clients to execute arbitrary `GraphQLClientRequest`s.

Old generated model

```kotlin
class HelloWorldQuery(
  private val graphQLClient: GraphQLKtorClient<*>
) {
  suspend fun execute(requestBuilder: HttpRequestBuilder.() -> Unit = {}): GraphQLResponse<HelloWorldQuery.Result> =
      graphQLClient.execute(HELLO_WORLD_QUERY, "HelloWorldQuery", null, requestBuilder)

  data class Result(
    val helloWorld: String
  )
}
```

New generated model

```kotlin
class HelloWorldQuery: GraphQLClientRequest(HELLO_WORLD_QUERY, "HelloWorldQuery") {
  override fun returnType(): Class<HelloWorldQuery.Result> = HelloWorldQuery.Result::class.java

  data class Result(
    val helloWorld: String
  )
}
```

`GraphQLClient` interface was udated to accept arbitrary `GraphQLClientRequest`s which allows us to support both individual as well as batch requests.

```kotlin
interface GraphQLClient<R> {

    suspend fun <T> execute(request: GraphQLClientRequest, requestBuilder: R.() -> Unit = {}): GraphQLResponse<T>

    suspend fun execute(requests: List<GraphQLClientRequest>, requestBuilder: R.() -> Unit = {}): List<GraphQLResponse<*>>
}
```

In order to be able to generically deserialize batch requests, `GraphQLClientRequest`s also need to provide return type `Class` information. For convenience, GraphQL Kotlin plugin also generate corresponding `execute<OperationName>` extension methods on the target client type that return operation specific `GraphQLResponse`.

### DIFF
--- a/clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/GraphQLClientRequest.kt
+++ b/clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/GraphQLClientRequest.kt
@@ -16,20 +16,19 @@
 
 package com.expediagroup.graphql.client
 
-import com.expediagroup.graphql.types.GraphQLResponse
-
 /**
- * A lightweight typesafe GraphQL HTTP client.
+ * Abstract class representing GraphQL request that follows the common GraphQL HTTP request format.
+ *
+ * @see [GraphQL Over HTTP](https://graphql.org/learn/serving-over-http/#post-request) for additional details
  */
-interface GraphQLClient<RequestCustomizer> {
+abstract class GraphQLClientRequest(
+    val query: String,
+    val operationName: String? = null,
+    val variables: Any? = null
+) {
 
     /**
-     * Executes [GraphQLClientRequest] and returns corresponding [GraphQLResponse].
+     * Parameterized type of a corresponding GraphQLResponse.
      */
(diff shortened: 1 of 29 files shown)


clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/GraphQLClientRequest.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.expediagroup.graphql.client
+
+/**
+ * Abstract class representing GraphQL request that follows the common GraphQL HTTP request format.
+ *
+ * @see [GraphQL Over HTTP](https://graphql.org/learn/serving-over-http/#post-request) for additional details
+ */
+abstract class GraphQLClientRequest(
```
dariuszkuc, line 24: created separate abstract class over using existing `GraphQLRequest` data class as:
* data classes cannot be extended
* need to provide explicit response type to support deserialization of batch responses
* variables are set as `Any?` instead of a `Map` so we can pass arbitrary data classes and they will be automatically serialized as JSON map

smyrick (approved):

PR #1031, opened 2021-01-30
Simplify setup of dataloaders in spring server

jarlehansen:
### :pencil: Description

**Why**
The setup of dataloaders is not always straight forward, also the examples are pretty limited. The reason behind this PR is to simplify the setup of dataloaders in the spring server.

**How**
This code works by adding an interface called `KeysDataLoader`. This is meant to be implemented on your own dataloaders. It is automatically registering in the `DataLoaderRegistry` by autowiring the Spring beans by type. All beans of type `KeysDataLoader` is registered.

From the query class the dataloaders is used by calling the load extension function on the `DataFetchingEnvironment`. This will lookup the correct dataloader for you, based on type and/or name given, and return the correct value. The value is wrapped in a `CompletableFuture` to support batching of the dispatching in the dataloader. The load function does not use coroutines because it is the java-dataloader implementation to trigger the dispatching. However, the load function and the actual dataloader does use coroutines and suspended functions. See `SpringDataLoaderConfiguration.batchDataLoader` for more details on how this integration between coroutines and java futures is done.

This feature, of automatically registering the dataloader spring beans, can be enabled/disable with the `graphql.spring.dataloader.enabled` property.

Several good examples on how this works is available in the `SpringDataLoaderIT` class.

**Future work**
- Should the number of backing threads for the coroutines scope be configurable?
- Should it be possible to enable caching in the data loaders?
- Improve unit tests coverage. I have added an integration test `SpringDataLoaderIT` that covers most of the changes, and `SpringDataLoaderTest` that is a normal unit test. However, it seems that jacoco is not including the extension function as part of the coverage. How can I fix this?


jarlehansen:
We are currently using this in our graphql gateway at the Norwegian Tax Administration: https://github.com/Skatteetaten/gobo
Here is one example where it is used: https://github.com/Skatteetaten/gobo/blob/feature/AOS-5279-dataloaders/src/main/kotlin/no/skatteetaten/aurora/gobo/graphql/affiliation/Affiliation.kt#L31

This setup has been working quite well for us, and we hope that it can be helpful for others as well.

RIP21:
Looks interesting :) Hope maintainers will take a look at this one :) 
WIll be glad to get rid of the half-baked solution I'm using and have "native" support from the library.
I guess also a good thing/idea would be is to add examples to the `examples/server/spring-server`. Replacing the example that they're having with dataloader to be using this one instead.

smyrick:
I like the idea of simplifying the loading logic. To make it easier for us as the library maintainers I think we should still let users create and configure the data loader. So what we can help with is the registration and loading of values. 

It simplifies everything and is pretty straight forward for us. Here is a PR into your branch with some changes: https://github.com/jarlehansen/graphql-kotlin/pull/1

dariuszkuc:
Can we also update [documentation](https://expediagroup.github.io/graphql-kotlin/docs/next/server/data-loader-registry-factory) to reflect the changes?

jarlehansen:
I have merged the PR on my branch. I can certainly help with the documentation as well.
Thanks for the refactoring :)

smyrick:
After doing this refactoring I also noticed that we really don't need the `DataLoaderRegistryFactory` anymore either. As long as there are some `KotlinDataLoaders` that we can use we can have common code to add them to the registry.

This would be a major change though and would directly impact the docs we would have to write here. So I am going to follow up with another PR of my own of this breaking change and update the docs there instead.

smyrick (approved):

RIP21:
@smyrick is there anything I can subscribe to get notified of release with this feature?
@jarlehansen have you wrote documentation? :) Want to see how it will look like after its release to refactor respectively.

smyrick:
@RIP21 You can "watch" the repo in Github, and if you don't want to be notified of all changes, you can just subscribe to releases.

I updated the docs here: https://github.com/ExpediaGroup/graphql-kotlin/pull/1040

Issue #929, opened 2020-11-03
Migrate graphql-client from jackson to kotlinx-serialization


smyrick:
**Is your feature request related to a problem? Please describe.**
Currently we can not use kotlin multi platform since Jackson only works on the JVM: https://github.com/FasterXML/jackson

**Describe the solution you'd like**
We should drop the dependency on jackson and instead migrate to https://github.com/Kotlin/kotlinx.serialization

**Describe alternatives you've considered**
We could continue to use jackson as it is a well established and supported serialization library

**Additional context**
This would help us support multiplatform and kotlin native approach: https://github.com/ExpediaGroup/graphql-kotlin/issues/876

Issue #876, opened 2020-09-09
Kotlin multiplatform ktor client


jillesvangurp:
**Is your feature request related to a problem? Please describe.**

The current ktor client (4.0.0-alpha.3) is close to what we need to target IOS and Android with a multiplatform client for our graphql server but not quite. However, a few simple fixes would make it possible to turn this into a multiplatform project allowing the generated code to be used from web, native, wasm, and jvm platforms.

**Describe the solution you'd like**
- Change the gradle build file for the ktor client to use the multiplatform plugin. Define e.g. jvm, ios-x86 and ios-arm, and js variants.
- Move configurations for ktor engine and json serialization into the plaftorm specific part of the source tree and keep the rest of the client code in the common part. From what I've seen there are no big obstacles to doing this as  the generated code is cross platform and the ktor client is already cross platform.
- For the jvm builds, switch to using gson, which is the defacto serialization on Android instead of jackson.

**Describe alternatives you've considered**
- We could build our own client and simply use the same class name and packages so the imports for the generated code work. This should not be hard. Short term this is probably what we will do.

- Alternatively, the generated code could be changed to simply require the user to provide their own ktor httpclient instance. It would then be up to them to configure that properly and the generated code would have no com.expedia.* dependencies at all. Given how small this single class is, there is actually limited value in having a (wrong) default implementation there. At best its a convenience that users will probably want to override.

frankkoornstra:
Making this a multiplatform client would be amazing! I'm looking for a GraphQL client for a KotlinJS project and got very excited when I saw this project! I don't think there's much out there in terms of solid GraphQL HTTP clients so it'd might be an opportunity for this project to take a lead.

jillesvangurp:
@frankkoornstra you may want to check out the appollo gradle plugin which does something similar to this client for kotlin.

For this project, I've currently got a jvm only client. There are a few blockers that make multiplatform a bit tricky that I'm investigating workarounds for:

- the plugin only works with a jvm style project and breaks if you use it with a multiplatform project. A potential workaround would be letting it generate source code and then use a second gradle module that grabs the generated source folder and adds it to common/main in a multiplatform setup. The proper fix would be making the plugin be  a bit more flexible.
- you need a custom GraphqlClient implementation. The jar that contains that interface has a few rogue jvm specific dependencies. Workaround: just lose the dependency and copy the interface definition with the same package and name.
[...]

PR #1034, opened 2021-01-31
Migrate graphql client to kotlinx serialization

mpiniarski:
### :pencil: Description
In order to use `graphql-koltin-client` in multiplatform projects it's necessary to remove `jackson` dependency and use `kotlinx-serialization` instead.
This PR is an attempt to do this in separate subprojects (prefixed with `*-multiplatform`) based on `graphql-kotlin-types`, `graphql-kotlin-client` and `graphql-kotlin-ktor-client`. I've decided to create separate subprojects, because original ones are used by other parts of the library.
 
### :link: Related Issues
[#929](https://github.com/ExpediaGroup/graphql-kotlin/issues/929)
[#876](https://github.com/ExpediaGroup/graphql-kotlin/issues/876)


dariuszkuc:
Hello 👋 
Thanks for raising this but there is no need for separate projects. Instead we should be updating existing `graphql-kotlin-client` projects. I'm thinking that since we already need a client specific request object over the one in `graphql-kotlin-types` (see https://github.com/ExpediaGroup/graphql-kotlin/pull/1030) we probably should just create a client specific response as well - since Jackson is the most popular library for Spring I don't think it makes sense for server to use `kotlinx.serialization`. 

In order to migrate to `kotlinx.serialization` we need to make the biggest changes to the client generation logic. Currently we rely on Jackson annotations for handling polymorphic types, default enum values and scalar wrappers. We would need to update both client generation logic and actual client code at the same time.

PR #1040, opened 2021-02-11
Generate DataLoaders on every request

smyrick:
### :pencil: Description
With the introduction of the new `KotlinDataLoader` interface we made the way to get the data loaders a property. However taking the advice from `graphql-java` we should generate each `DataLoader` every request. The factory generates the registry for every request currently but the data loaders themselves would still be static.

This also updates the docs and examples around data loaders from the last PR

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/pull/1031


servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/KotlinDataLoader.kt
```diff
@@ -23,6 +23,6 @@ import org.dataloader.DataLoader
  * by return type and loading values in the data fetchers.
  */
 interface KotlinDataLoader<K, V> {
```
dariuszkuc, line 25: wondering whether it would make sense to change this to `abstract class` that extends the `DataLoader`

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -37,8 +38,6 @@ open class GraphQLRequestHandler(
      * Subscriptions require more specific server logic and will need to be handled separately.
      */
     open suspend fun executeRequest(request: GraphQLRequest, context: GraphQLContext? = null): GraphQLResponse<*> {
-        // We should generate a new registry for every request
-        val dataLoaderRegistry = dataLoaderRegistryFactory?.generate()
```
dariuszkuc, line 41: i thought that data loader registry should be re-generated per request?

smyrick: From graphql-java: 

Per Request Data Loaders
If you are serving web requests then the data can be specific to the user requesting it. If you have user specific data then you will not want to cache data meant for user A to then later give it to user B in a subsequent request.

The scope of your DataLoader instances is important. You will want to create them per web request to ensure data is only cached within that web request and no more. It also ensures that a dispatch call only affects that graphql execution and no other.

DataLoaders by default act as caches. If they have seen a value before for a key then they will automatically return it in order to be efficient.

If your data can be shared across web requests then you might want to change the caching implementation of your data loaders so they share data via a caching layer say like memcached or redis.

You still create data loaders per request, however the caching layer will allow data sharing (if that’s suitable).

smyrick: So based off of that we may not actually want to use the new format and users shouldn't create data loaders as beans?

smyrick: This is where again I think there is no need for data loaders in our library but I understand some users want to use pre-built caching solutions.

dariuszkuc: I'm thinking we could keep the KotlinDataLoader interface but change from val field to fun to generate the dataloader. We could then easily recreate the registry per request.

dariuszkuc: And yeah we probably should keep the registry factory

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/KotlinDataLoader.kt
smyrick: DataLoader is a class, not an interface so we would have to implement one of the many constructors. I think I better option is making the `val dataLoader` into a function

smyrick:
It is a breaking change from the last PR but we haven't released yet so not a breaking API change from last release

docs/server/data-loaders.md
```diff
@@ -0,0 +1,86 @@
...
+To help in the registration of these various `DataLoaders`, we have created a basic interface `KotlinDataLoader`:
+
+```kotlin
+interface KotlinDataLoader<K, V> {
+    val dataLoaderName: String
+    fun getDataLoader(): DataLoader<K, V>
+}
+```
+
+This allows for library users to still have full control over the creation of the `DataLoader` and its various configuraiton options,
```
dariuszkuc, line 41:
```suggestion
This allows for library users to still have full control over the creation of the `DataLoader` and its various configuration options,
```

dariuszkuc:
Changes look fine but since we no longer delete the factory can you update the title and description?

smyrick:
Updated PR info

dariuszkuc (approved):

PR #1041, opened 2021-02-11
Delete older alpha docs

smyrick:
### :pencil: Description
The release job automatically does a cut of the docs, however there is no need to keep different versions of the alpha docs around. If someone is on the alpha branch they should keep up-to-date. The next PR I am making though is to rename the old 3.6.1 to 3.x.x to make it clear that we can not update 3.x.x docs even though we can still update the code there. That is because the docs are always the ones from the master branch so even if we update docs in the 3.x.x branch those change will never be seen anywhere


dariuszkuc:
"Technically" you can publish new version of 3.x docs but that is a manual process. Since in docusaurus v1 versions only store what changed it is troublesome and probably not worth the effort (I believe it was simplified in docusaurus v2 to always store all docs).

dariuszkuc (approved):

PR #1048, opened 2021-02-13
[client] update GraphQL client to use kotlinx.serialization

dariuszkuc:
### :pencil: Description
TODO

### :link: Related Issues
Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/929


plugins/client/graphql-kotlin-client-generator/build.gradle.kts
```diff
@@ -16,6 +17,7 @@ dependencies {
         exclude(group = "com.graphql-java", module = "java-dataloader")
     }
     api("com.squareup:kotlinpoet:$kotlinPoetVersion")
+    api("org.jetbrains.kotlinx:kotlinx-serialization-json:$kotlinxSerializationVersion")
     implementation("io.ktor:ktor-client-cio:$ktorVersion")
     implementation("io.ktor:ktor-client-jackson:$ktorVersion")
```
smyrick, line 22: I am assuming we will remove jackson then too?

dariuszkuc:
Kotlin 1.4 requires sealed classes (required for supporting polymorphism in `kotlinx-serialization`) to be declared at the top level (and all implementations to be in the same file) which breaks our client generation where we put all query specific data classes under that query request wrapper.

We could potentially move polymorphic types to their own files but since Kotlin 1.5 will fix the above issue - I'd suggest to just wait for Kotlin 1.5 release instead.

dariuszkuc:
PR is huge but majority of it are from client generation tests that were updated to use KotlinX serialization by default and moved from hard coded Strings inside test classes to a separate files

```
# running from plugins/client/graphql-kotlin-client-generator/src/test/data
# total number of test files
> find . -type f | wc -l
     149
# total lines of code
> wc -l `find . -type f`
...
    3191 total
```

dariuszkuc:
Superseded by #1066, #1069 and #1070

PR #1049, opened 2021-02-18
Update to docusaurus 2

smyrick:
### :pencil: Description
Update to docusaurus 2

https://v2.docusaurus.io/docs/migration

As part of the migration instead of making new docs version for every release, we are just going to have major versions. `4.x.x` `3.x.x`. This is done in the GitHub actions to get the current major version number at append `"${MAJOR_VERSION}.x.x"`


dariuszkuc:
I think you will have to fix the tabbing from custom script to use mdx (which would allow us to keep the same tabs selected on a page).

smyrick:
I created the PR, but then I saw the option to enable errors on bad links so I am fixing those in older versions now too

smyrick:
Updated all broken links and code tabs

smyrick:
Running locally

Screen Shot 2021-02-19 at 10 38 38 AM
Screen Shot 2021-02-19 at 10 38 47 AM

dariuszkuc (approved):
Moving those .md files and using new docusaurus is over 20K+/10K- line changes... Lets merge it and fix stuff going forward as needed.

Issue #1052, opened 2021-02-19
Support suspend functions for context factory


smyrick:
**Is your feature request related to a problem? Please describe.**
I would like to be able to make asynchronous calls out to a service when generating my context object.

**Describe the solution you'd like**
Update the interface to `suspend` function

```kotlin
interface GraphQLContextFactory<out Context : GraphQLContext, Request> {
    suspend fun generateContext(request: Request): Context?
}
```

**Describe alternatives you've considered**
We could force users to run the factory in a blocking fashion but since the server is already suspendable we can easily add this to the interface

**Additional context**
Initially reported here, it was moved to discussion but you can't move discussions back to issues https://github.com/ExpediaGroup/graphql-kotlin/issues/1050

Issue #945, opened 2020-11-27
Not possible to have a custom context for subscription methods


gerald24:
**Library Version**
com.expediagroup:graphql-kotlin-spring-server:4.0.0-alpha.7

**Describe the bug**
Unable to provide a custom context for subscriptions

**To Reproduce**
Subscription Example (it's not important here if context is of type GraphQLContext or any other, because currently a new instance of graphql.GraphQLContext will be provided anyway - see GraphQLRequest extension below):

```
import com.expediagroup.graphql.spring.operations.Subscription
import graphql.GraphQLContext
import org.springframework.stereotype.Component
import reactor.core.publisher.Flux
import java.time.Duration
import java.util.*

@Component
class NotificationSubscription : Subscription {

    fun notifications(context: GraphQLContext): Flux<String> {

        return Flux.interval(Duration.ofSeconds(5))
            .map {
                UUID.randomUUID().toString()
            }
    }
}
```

is not possible since on start we get an exception:

```
Caused by: com.expediagroup.graphql.exceptions.TypeNotSupportedException: Cannot convert graphql.GraphQLContext since it is not a valid GraphQL type or outside the supported packages
```

Adding "graphql" to `graphql.packages` config property is not desired and will result in an exception anyway:

```
Caused by: com.expediagroup.graphql.exceptions.EmptyInputObjectTypeException: Invalid GraphQLContextInput input object type - input object does not expose any fields.
```

But just noted - `graphql.GraphQLContext` is the default context, generated by the GraphQLRequest#toExecutionInput extension,  if the Mono-subscriber-context will not have one:

```
fun GraphQLRequest.toExecutionInput(graphQLContext: Any? = null, dataLoaderRegistry: DataLoaderRegistry? = null): ExecutionInput =
    ExecutionInput.newExecutionInput()
        .query(this.query)
        .operationName(this.operationName)
        .variables(this.variables ?: emptyMap())
        .context(graphQLContext ?: GraphQLContext.newContext().build())
        .dataLoaderRegistry(dataLoaderRegistry ?: DataLoaderRegistry())
        .build()
```

Here is the default implementation of the `SubscriptionHandler` (`SimpleSubscriptionHandler`):

```
    override fun executeSubscription(graphQLRequest: GraphQLRequest): Flux<GraphQLResponse<*>> = Mono.subscriberContext()
        .flatMapMany { reactorContext ->
[...]

PR #1053, opened 2021-02-19
Generate context with suspend and nullable in subscriptions

smyrick:
### :pencil: Description
There are two issues we are fixing here:
* The `GraphQLContextFactory` should support `suspend` functions
* Subscriptions need to support getting a null context from the factory

Supporting `suspend` functions is easy enough. We just updated the interface. However that also means in `ApolloSubscriptionProtocolHandler` we need to be able to run the suspend function in a class that deals with reactor `Flux`. Once we generate the context it could come back as `null`. If it does our old logic was to return a class `DefaultGraphQLContext`. However, this was now blowing up because if you had a custom context of type `MyCustomContext` and used that as an argument in your subscription functions, graphql-java was throwing an illegal argument exception because we were trying to pass `DefaultGraphQLContext` to the type `MyCustomContext`.

So what that means is that we really can't have a `DefaultGraphQLContext` instead, if the `GraphQLContextFactory` implementation could possibly return null, the schema functions need to have a parameter type defined as nullable

```kotlin
fun ticker(ctx: SubscriptionContext?): Flux<String> = Flux.just("${ctx?.value}:${Random.nextInt()}")
```

### :link: Related Issues
Default Context added here: https://github.com/ExpediaGroup/graphql-kotlin/pull/955
Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/945
Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1052

### DIFF
--- a/servers/graphql-kotlin-spring-server/build.gradle.kts
+++ b/servers/graphql-kotlin-spring-server/build.gradle.kts
@@ -35,7 +35,7 @@ tasks {
                 limit {
                     counter = "BRANCH"
                     value = "COVEREDRATIO"
-                    minimum = "0.74".toBigDecimal()
+                    minimum = "0.70".toBigDecimal()
                 }
             }
         }
--- a/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/ApolloSubscriptionHooks.kt
+++ b/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/ApolloSubscriptionHooks.kt
@@ -17,7 +17,6 @@ package com.expediagroup.graphql.server.spring.subscriptions
 
 import com.expediagroup.graphql.generator.execution.GraphQLContext
 import org.springframework.web.reactive.socket.WebSocketSession
-import reactor.core.publisher.Mono
 
 /**
  * Implementation of Apollo Subscription Server Lifecycle Events
@@ -33,8 +32,8 @@ interface ApolloSubscriptionHooks {
     fun onConnect(
         connectionParams: Map<String, String>,
         session: WebSocketSession,
-        graphQLContext: GraphQLContext
-    ): Mono<GraphQLContext> = Mono.just(graphQLContext)
+        graphQLContext: GraphQLContext?
+    ): GraphQLContext? = graphQLContext
 
     /**
      * Called when the client executes a GraphQL operation.
@@ -43,7 +42,7 @@ interface ApolloSubscriptionHooks {
     fun onOperation(
         operationMessage: SubscriptionOperationMessage,
         session: WebSocketSession,
-        graphQLContext: GraphQLContext
+        graphQLContext: GraphQLContext?
     ): Unit = Unit
 
     /**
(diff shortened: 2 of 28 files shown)


servers/graphql-kotlin-spring-server/build.gradle.kts
```diff
@@ -35,7 +35,7 @@ tasks {
                 limit {
                     counter = "BRANCH"
                     value = "COVEREDRATIO"
-                    minimum = "0.74".toBigDecimal()
+                    minimum = "0.70".toBigDecimal()
```
smyrick, line 38: With removal of the default context factory we don't have code we are testing much anymore. We could separate some of the routers to internal functions but there is not much going on anymore in the spring module other than defining the beans and subscriptions which are covered.

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/ApolloSubscriptionHooks.kt
```diff
@@ -33,8 +32,8 @@ interface ApolloSubscriptionHooks {
     fun onConnect(
         connectionParams: Map<String, String>,
         session: WebSocketSession,
-        graphQLContext: GraphQLContext
-    ): Mono<GraphQLContext> = Mono.just(graphQLContext)
+        graphQLContext: GraphQLContext?
+    ): GraphQLContext? = graphQLContext
```
smyrick, line 36: There is no need for a mono wrapper, we should just cache the context to be used later

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/ApolloSubscriptionProtocolHandler.kt
```diff
@@ -149,15 +149,19 @@ class ApolloSubscriptionProtocolHandler(
...
-        val acknowledgeMessage = Flux.just(acknowledgeMessage)
+        saveContext(operationMessage, session)
+        val acknowledgeMessage = Mono.just(acknowledgeMessage)
         val keepAliveFlux = getKeepAliveFlux(session)
         return acknowledgeMessage.concatWith(keepAliveFlux)
     }
 
+    private fun saveContext(operationMessage: SubscriptionOperationMessage, session: WebSocketSession) {
+        val connectionParams = getConnectionParams(operationMessage.payload)
+        val context = runBlocking { contextFactory.generateContext(session) }
```
smyrick, line 160: This line here is still ambiguous to me. If we wrapping this in `mono { }` then the code is not run unless the client subscribes and then it might not run first before the operations are exectuted

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/SpringGraphQLSubscriptionHandler.kt
```diff
@@ -34,8 +34,8 @@ import reactor.kotlin.core.publisher.toFlux
  */
 open class SpringGraphQLSubscriptionHandler(private val graphQL: GraphQL) {
 
-    fun executeSubscription(graphQLRequest: GraphQLRequest, graphQLContext: GraphQLContext?): Flux<GraphQLResponse<*>> = Flux.deferContextual {
-        graphQL.execute(graphQLRequest.toExecutionInput(graphQLContext))
+    fun executeSubscription(graphQLRequest: GraphQLRequest, graphQLContext: GraphQLContext?): Flux<GraphQLResponse<*>> {
+        return graphQL.execute(graphQLRequest.toExecutionInput(graphQLContext))
```
dariuszkuc, line 38: do we want to switch to expression body syntax (as it was before) instead?

dariuszkuc:
Can we also update the docs to reflect those changes?

dariuszkuc (approved):

PR #1059, opened 2021-02-20
Move federation docs under generator folder

smyrick:
### :pencil: Description
Move the docs to reflect the sidebar organization


generator/graphql-kotlin-federation/README.md
```diff
@@ -4,7 +4,7 @@
 
 `graphql-kotlin-federation` extends the functionality of `graphql-kotlin-schema-generator` and allows you to easily
 generate federated GraphQL schemas directly from the code. Federated schemas rely on a number of directives to
-instrument the behavior of the underlying graph, see [the documentation](https://expediagroup.github.io/graphql-kotlin/docs/federated/federated-directives).
+instrument the behavior of the underlying graph, see [the documentation](https://expediagroup.github.io/graphql-kotlin).
```
dariuszkuc, line 7: broken link?

dariuszkuc:
Sometimes it takes a bit for a new release so the if we remove the distinction it will be possible to have docs not matching released version.

generator/graphql-kotlin-federation/README.md
smyrick: This is in the README. Instead of always having to update we can just reference the docs in general

smyrick:
So would we rather have docs that may be ahead of released versions or have the distinct "unreleased" docs?

website/versioned_docs/version-4.x.x/plugins/maven-plugin.md
```diff
@@ -73,13 +73,13 @@ Generate GraphQL client code based on the provided GraphQL schema and target que
 
     ```xml
 
-    
-      
-        
+
```
dariuszkuc, line 76: i think with the migration xml got messed up -> see https://expediagroup.github.io/graphql-kotlin/docs/plugins/maven-plugin-usage#complete-configuration-example

dariuszkuc:
> So would we rather have docs that may be ahead of released versions or have the distinct "unreleased" docs?

Yeah I'd keep the "next"/"unreleased"

dariuszkuc:
Looks like both maven and gradle plugin doc sites are messed up -> https://expediagroup.github.io/graphql-kotlin/docs/next/plugins/gradle-plugin-tasks

website/versioned_docs/version-4.x.x/plugins/maven-plugin.md
smyrick: Something must have happened in IntelliJ git revert

smyrick:
I updated the docs copying values from before the docusaurus 2 conversion, and then fixing issues. Instead of updating `4.x.x` docs let's merge this and have the release job update them to verify everything is good there too

dariuszkuc (approved):

PR #1060, opened 2021-02-20
Delete versioned 4.x.x docs for renamed current docs

smyrick:
### :pencil: Description
The docusaurus workflow does not allow us to keep updating existing version of docs. Versions are supposed to be specific cuts and only maintained.

We will have the `current` version labeled as `4.x.x` so docs will always be the latest on the master branch. This means the docs may be slightly ahead of the released version but we shouldn't be making any significant breaking changes to docs that is not reflected in code once we are on stable releases.

See latest version docs config here:
* https://v2.docusaurus.io/docs/versioning/#figure-out-the-behavior-for-the-current-version
* https://v2.docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs

### :link: Related Issues
Feature request to add `--overwrite` flag closed: https://github.com/facebook/docusaurus/issues/3478


smyrick:
What the different versions dropdown and urls look like




Screen Shot 2021-02-20 at 12 35 25 PM

smyrick:
See the failed action here: https://github.com/ExpediaGroup/graphql-kotlin/actions/runs/584922400

dariuszkuc:
How will this work when we start releasing 5.0.0?

smyrick:
We will have to manually run the command `npm run version 4.x.x` in the website, then update the config to set the latest as `5.x.x`

dariuszkuc (approved):

dariuszkuc (approved):

PR #1064, opened 2021-02-27
[build] drop jcenter repositories

dariuszkuc:
### :pencil: Description

JCenter is being sunset as per https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/. We don't have any JCenter specific dependencies and we only had it configured in the build file for convenience in resolving artifacts.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,13 +24,11 @@ allprojects {
         repositories {
             mavenLocal()
             mavenCentral()
-            jcenter()
         }
     }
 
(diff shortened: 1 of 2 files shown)


dariuszkuc:
Looks like we are dependent on Detekt -> https://github.com/detekt/detekt/pull/3455

dariuszkuc:
it might take a while for other dependent repos to migrate so closing it for now

PR #1065, opened 2021-03-01
Add support for federated tracing

smyrick:
### :pencil: Description

This adds a new `FederatedGraphQLContext` to the `graphql-kotlin-federation` package. It also creates a new `SpringGraphQLContext` and `SpringGraphQLContextFactory` that implement these required methods.

If users to do not have a custom context, they just need to set `graphq.federation.tracing.enabled=true`.

If there is a custom context, it should extend `SpringGraphQLContext` and also still have the properties set.

Since we are changing the required type of `SpringGraphQLContextFactory` this is a major change in the `4.x.x` branch.

### :link: Related Issues
Reimplements https://github.com/ExpediaGroup/graphql-kotlin/pull/962 for merge conflicts
Support in 3.x.x added here: #960

### DIFF
--- a/examples/federation/base-app/src/main/resources/application.yml
+++ b/examples/federation/base-app/src/main/resources/application.yml
@@ -1,6 +1,9 @@
 graphql:
   federation:
     enabled: true
+    tracing:
+      enabled: true
+      debug: true
   packages:
     - "com.expediagroup.graphql.examples.federation.base"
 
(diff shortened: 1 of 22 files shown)


smyrick:
Running examples with new return extensions

Screen Shot 2021-02-28 at 8 37 10 PM

dariuszkuc (approved):

PR #1066, opened 2021-03-05
[client] create new GraphQL client serialization abstraction

dariuszkuc:
### :pencil: Description

This is a first PR in a series that will introduce support for both `Jackson` and `kotlinx.serialization` data formats for GraphQL Kotlin clients. This PR provides general refactoring of GraphQL client request and response types and introduces new GraphQL client serialization abstraction. We need this additional abstraction as we need to provide some default configurations for both `Jackson` and `kotlinx.serialization` serializers and we cannot rely on the provided defaults. By serializing requests to raw String (and deserializing responses from String) allows us to simplify the configuration logic that would be required to support both formats across Ktor and Spring based clients. If serializer is not explicitly specified, then default one will be automatically loaded from a classpath using `Java ServiceLoader` mechanism.

NOTE: clients still default to use Jackson which is currently only supported format from the generator (support for generation of `kotlinx.serialization` data models will be added in subsequent PRs)

Changes:

* Update GraphQL client request and response types to support both `Jackson` and `kotlinx.serialization` formats.
* Create new `GraphQLClientSerializer` abstraction that is used by the GraphQL clients to serialize requests and deserialize responses
* Create `Jackson` and `kotlinx.serialization` compatible GraphQL serializers
* Drop the generation of unnecessary `executeXQuery` extension functions

### :link: Related Issues

* https://github.com/ExpediaGroup/graphql-kotlin/pull/1048 - this is an attempt to split it up to more manageable pieces
* https://github.com/ExpediaGroup/graphql-kotlin/issues/929 - adds initial support for `kotlinx.serialization`


clients/graphql-kotlin-client-serialization/src/main/kotlin/com/expediagroup/graphql/client/serialization/types/KotlinXGraphQLError.kt
```diff
@@ -0,0 +1,30 @@
...
+
+package com.expediagroup.graphql.client.serialization.types
+
+import com.expediagroup.graphql.client.serialization.serializers.AnyKSerializer
+import com.expediagroup.graphql.client.serialization.serializers.GraphQLErrorPathSerializer
+import com.expediagroup.graphql.client.types.GraphQLError
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class KotlinXGraphQLError(
```
smyrick, line 25: Should we call this different errors or should we have the specific type just be in the package name? Should we keep as `KotlinGraphQLError`?

dariuszkuc: Named those implementations differently to avoid the name clash when importing both interface and implementation. That being said if you prefer same names we can definitely rename those - probably would drop the "client" part in the interface names as well.

smyrick (approved):

Issue #1067, opened 2021-03-05
What is the recommended way for configuring customScalars for the maven plugin?


ssullivan:
I am testing the latest alpha version of this library that I could find: 4.0.0-alpha.14

I've created a project: [https://github.com/ssullivan/graphql-kotlin-customscalar-exampl](https://github.com/ssullivan/graphql-kotlin-customscalar-example) based on the example maven-client project from this repository. I've a made few small mods to the maven setup and customized the queries, and graphql schema so I can test out the plugin.

The customScalars that are specified don't appear to be used by the generated code. It looks like only the type aliases defined defined in `GraphQLTypeAliases.kt` are being utilized.

I noticed that the `GenerateClientAbstractMojo.kt` has a parameter defined [here](https://github.com/ExpediaGroup/graphql-kotlin/blob/4a89a8b183defa6a1f25494f2f9f90753f2723b2/plugins/graphql-kotlin-maven-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/maven/GenerateClientAbstractMojo.kt#L73):

```kotlin
@Parameter(name = "converters")
private var customScalars: List<CustomScalar> mutableListOf()
```

I checked out the graphql-kotlin master branch and modified this section of code locally to look like the following

```kotlin
@Parameter(name = "customScalars")
private var customScalars: List<CustomScalar> mutableListOf()
```

After making this change my generated code is using the converter specified by the CustomScalar configuration.

Am I just configuring things wrong here or is this a bug?

dariuszkuc:
Hello. Thanks for raising this. It is indeed a bug (I hit the same problem when working on adding kotlinx serialization support).

PR #1068, opened 2021-03-06
WIP: Change the parameterName so it can be used in the maven plugin config…

ssullivan:
This resolves the maven configuration problem I asked about in issue #1067

### :pencil: Description

Changes the maven parameter name from `converters` to `customScalars`.

### :link: Related Issues
* #1067

### DIFF
--- a/plugins/graphql-kotlin-maven-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/maven/GenerateClientAbstractMojo.kt
+++ b/plugins/graphql-kotlin-maven-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/maven/GenerateClientAbstractMojo.kt
@@ -70,7 +70,7 @@ abstract class GenerateClientAbstractMojo : AbstractMojo() {
      * </customScalars>
      * ```
      */
-    @Parameter(name = "converters")
+    @Parameter(name = "customScalars")
     private var customScalars: List<CustomScalar> = mutableListOf()
 
     /**


dariuszkuc:
There is a failing test that imports wrong `UUID` -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/plugins/graphql-kotlin-maven-plugin/src/integration/complete-setup/src/test/kotlin/com/expediagroup/graphql/plugin/maven/GraphQLMavenPluginTest.kt#L20

i.e. tests was incorrect as it was importing typealias whereas it should be importing custom scalar `com.expediagroup.graphql.plugin.generated.ExampleQuery.UUID`

dariuszkuc:
Hello 👋 FYI I got a big refactoring logic of the `graphql-kotlin-client-generator` (https://github.com/ExpediaGroup/graphql-kotlin/pull/1069) that updates the generator logic to create data classes in separate files outside of the query files, since I have to touch the same files I also fixed the maven parameter name in the mojo

i.e. given `HelloWorldQuery`

* query specific files will be under package name matching that operation name, e.g. `helloworldquery.MyObject`
* shared definitions will be placed under their specific packages, e.g.
  * `inputs.MyInput`
  * `enums.MyCustomEnum`
  * `scalars.MyCustomScalar`

dariuszkuc:
Fixed in https://github.com/ExpediaGroup/graphql-kotlin/pull/1069

PR #1069, opened 2021-03-08
[plugin] update client generator to support kotlinx serialization

dariuszkuc:
### :pencil: Description

This is a second PR in a series to introduce support for both `Jackson` and `kotlinx.serialization` data formats for GraphQL Kotlin clients.

This PR updates `graphql-kotlin-client-generator` to allow generating data models supporting both formats. Due to the declaration limitations of Kotlin sealed classes (in Kotlin 1.4 sealed classes and their implementations have to be a top level definitions), generation logic was also updated to generate operation specific data models under package name matching that operation name. Common objects such as input objects, enums and custom scalars definitions will be shared across different queries.

Large size of PR is driven primarily with test cases - instead of comparing generated classes against a hard coded Strings, tests were updated to compare it against files. Generator tests were also updated to verify generated sources can be compiled.

### :link: Related Issues

* https://github.com/ExpediaGroup/graphql-kotlin/pull/1048 - this is an attempt to split it up to more manageable pieces
* https://github.com/ExpediaGroup/graphql-kotlin/issues/929 - add support for generating `kotlinx.serialization` data models
* https://github.com/ExpediaGroup/graphql-kotlin/pull/1066 - first PR in the series
* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1067


dariuszkuc:
PR is huge due to all the data classes used for verifying generation logic

```
# running from plugins/client/graphql-kotlin-client-generator/src/test/data
# total number of test files
> find . -type f | wc -l
     168
# total lines of code
> wc -l `find . -type f`
...
    3704 total
```

smyrick (approved):

Issue #1062, opened 2021-02-26
Handling of GraphQLErrors in KotlinDataFetcherExceptionHandler


hsudbrock:
[The KDoc](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/exception/KotlinDataFetcherExceptionHandler.kt) of `KotlinDataFetcherExceptionHandler` states:
> All exceptions thrown during GraphQL execution are logged and then returned as wrapped ExceptionWhileDataFetching error if they are not a valid GraphQLError

However, the actual implementation also wraps valid `GraphQLError` exceptions as an `ExceptionWhileDataFetching`, instead of passing those directly into the `DataFetcherExceptionHandlerResult`.

Do I misunderstand the KDoc, or is this a small issue that needs to be fixed? (If it needs to be fixed, I would of course be willing to provide a fix - I'd just like to clarify whether this actually is an issue first).

I think that *not* wrapping `GraphQLError` instances in an `ExceptionWhileDataFetching` would be beneficial, because it would allow implementors to define the error message displayed in the graphql response, while currently the error message from the original `GraphQLError` is modified by the `KotlinGraphQLError`.

dariuszkuc:
Hello 👋 
It looks like `KDoc` wasn't updated. The reason why we wrap the exceptions in `KotlinGraphQLError` is to ensure proper serialization of error response -> i.e. we rely on Jackson which uses reflections to serialize all fields that have getters. So if you throw any of the `graphql-java` `GraphQLErrors` your exception field will also be serialized. I think we probably should update it to check if it is already an instance of `KotlinGraphQLError` and only wrap it if it is not.

PR #1083, opened 2021-03-21
Do not wrap exceptions twice

smyrick:
### :pencil: Description
Do not wrap the exceptions twice if they are already a `GraphQLError`

This now also deletes the `KotlinGraphQLError` as it was no longer wrapping anything with required logic. We can instead just map a throwable, if it is not already a `GraphQLError` to a `GraphqlErrorException`

https://github.com/graphql-java/graphql-java/issues/1690

### :link: Related Issues
Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1062


servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/responseExtensions.kt
```diff
@@ -36,12 +35,17 @@ fun ExecutionResult.toGraphQLResponse(): GraphQLResponse<*> {
...
  */
-fun GraphQLJavaError.toGraphQLKotlinType() = GraphQLServerError(
-    message.orEmpty(),
-    locations?.map { it.toGraphQLKotlinType() },
-    path,
-    extensions
-)
+fun graphql.GraphQLError.toGraphQLKotlinType(): GraphQLServerError {
+    val newExtensions = this.extensions?.toMutableMap() ?: mutableMapOf()
+    newExtensions.putIfAbsent("classification", this.errorType.toSpecification(this))
```
dariuszkuc, line 40: classification in extension? i don't think this is "standard"

smyrick: This is info that is in the graphql-java error that is lost if we do a raw conversion

https://github.com/graphql-java/graphql-java/blob/479b7a43087ec997959e5c9ce3048842370c80fa/src/main/java/graphql/GraphqlErrorHelper.java#L42-L44

smyrick: If we didn't need to convert this to a `GraphQLServerError` and instead could just return a string we could use the built-in `GraphqlErrorHelper.toSpecification()`

dariuszkuc: we can keep the logic but you will have to check if error type is not null as otherwise you will get NPE (see https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/GraphqlErrorHelper.java#L35)

smyrick: Resolved

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/exception/KotlinGraphQLError.kt
```diff
@@ -22,26 +22,25 @@ import graphql.GraphQLError
...
+ * You can extend this class if you would like to create your own [GraphQLError]
+ * or just throw an existing [GraphQLError] implementation from your data fetcher.
  */
 open class KotlinGraphQLError(
     private val exception: Throwable,
     private val locations: List<SourceLocation>? = null,
     private val path: List<Any>? = null,
-    private val errorType: ErrorClassification = ErrorType.DataFetchingException
+    private val errorType: ErrorClassification = ErrorType.DataFetchingException,
+    private val extensions: Map<String, Any> = emptyMap()
```
dariuszkuc, line 35: I believe it should be `Map<String, Any?>?` - why do we default to empty map?

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/responseExtensions.kt
```diff
@@ -36,12 +35,20 @@ fun ExecutionResult.toGraphQLResponse(): GraphQLResponse<*> {
...
  */
-fun GraphQLJavaError.toGraphQLKotlinType() = GraphQLServerError(
-    message.orEmpty(),
-    locations?.map { it.toGraphQLKotlinType() },
-    path,
-    extensions
-)
+fun graphql.GraphQLError.toGraphQLKotlinType(): GraphQLServerError {
+    val locations = this.locations?.map { it.toGraphQLKotlinType() }
+    val newExtensions = this.extensions?.toMutableMap() ?: mutableMapOf()
```
dariuszkuc, line 40: if extensions are null and there is no error type information I'd assume the `newExtensions` should be null as well

smyrick: Not if we are going to immediately set the `classification` anyway

smyrick: I suppose now that we are no longer constructing a custom `GraphQLError` we can just use the extensions as is

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/SpringGraphQLSubscriptionHandler.kt
```diff
@@ -47,8 +47,8 @@ open class SpringGraphQLSubscriptionHandler(
             .toFlux()
             .map { result -> result.toGraphQLResponse() }
             .onErrorResume { throwable ->
-                val error = KotlinGraphQLError(throwable).toGraphQLKotlinType()
-                Flux.just(GraphQLResponse<Any?>(errors = listOf(error)))
+                val error = throwable.toGraphQLError()
+                Flux.just(GraphQLResponse<Any?>(errors = listOf(error.toGraphQLKotlinType())))
```
smyrick, line 51: This now also handles the wrapping of exceptions in subscriptions

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/responseExtensions.kt
```diff
@@ -33,14 +34,28 @@ fun ExecutionResult.toGraphQLResponse(): GraphQLResponse<*> {
     return GraphQLResponse(data, filteredErrors, filteredExtensions)
 }
 
+/**
+ * Turn a thrown exception into a graphql-java [GraphQLError] that we can return in the response
+ */
+fun Throwable.toGraphQLError(): GraphQLError {
+    return if (this is GraphQLError) {
```
smyrick, line 41: This is the big change. Everything stems from what we no longer need to do around error mapping

dariuszkuc, line 41: Lets change this to expression body syntax

dariuszkuc (changes requested):

dariuszkuc (approved):

Issue #1075, opened 2021-03-15
client codegen enums with 'name' or 'ordinal' property


FidasL:
**Library Version**
`3.7.0`

**Description**
When my GraphQL schema contains Enum with `name` or `ordinal` property name, it is not possible to use graphql-codegen in  my graphql client. 

<details>
  <summary>Example of generated Enum class</summary>
  
```
enum class EnumClass {
    createdAt,
    createdById,
    id,
    name,
    /**
     * This is a default enum value that will be used when attempting to deserialize unknown value.
     */
    @JsonEnumDefaultValue
    __UNKNOWN_VALUE
}
```
</details>

This generated class causes compile error, because Enums cannot have `name` or `ordinal` property. In my opinion, this is an unexpected behavior from two main reasons:
- there are lots of existing schemas with `name` property that cannot be easily modified. This is making your codegen for graphql client unusable in most use-cases. 
- best-practice is to use uppercase properties in Kotlin enums.

**Possible solution**
Using `@JsonProperty` in property definition can solve this problem. In example below, I defined enum properties in uppercase. With `@JsonProperty` annotation this property is exposed in requests in lowercase. 
<details>
  <summary>Example of possible solution</summary>

```
enum class EnumClass {
    @JsonProperty("name") 
    ID,

    @JsonProperty("name") 
    NAME,

    /**
     * This is a default enum value that will be used when attempting to deserialize unknown value.
     */
    @JsonEnumDefaultValue
    __UNKNOWN_VALUE
}
```
</details>

If you decide to use this solution, I can participate on creating PR for gradle codegen plugin. 

**Possible workaround**
We solved this problem easily with `finalizedBy` gradle task, that replaces all `name` properties in generated files with `@JsonProperty("name") NAME,` definition
<details>
  <summary>Example of workaround</summary>
 
 ```
task replaceEnumNames(type: Copy) {
    def oldPackageName = "com.graphql.generated"
    def newPackageName = "com.graphql.fixed"
    def oldJsonImport = "import com.fasterxml.jackson.annotation.JsonEnumDefaultValue"
    def newJsonImport = "import com.fasterxml.jackson.annotation.JsonEnumDefaultValue\nimport com.fasterxml.jackson.annotation.JsonProperty"

    def generatedDir = "$buildDir/generated/source/graphql/main/com/graphql/generated/"
    def fixedDir = "$buildDir/generated/source/graphql/main/com/graphql/fixed/"

    from generatedDir
[...]

PR #1087, opened 2021-03-23
[client] generate enums in UPPERCASE and fix kotlinx request serialization

dariuszkuc:
### :pencil: Description

Update client generation logic to always generate enums in UPPERCASE and then use corresponding Jackson (`@JsonProperty`) or kotlinx-serialization (`@SerialName`) annotation to convert those back to expected schema values. This allows us to support enum values that cannot be used directly in Kotlin (i.e. `name` and `ordinal`) as they would clash with built in methods.

This PR also resolves incorrect serialization of requests when using `kotlinx-serialization`. We were applying our custom `AnySerializer` to serialize the requests which was in turn incorrectly serializing enums and custom scalars. I changed the logic to construct correct `KSerializer` based on the specified request.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1075

### DIFF
--- a/clients/graphql-kotlin-client-jackson/src/main/kotlin/com/expediagroup/graphql/client/jackson/GraphQLClientJacksonSerializer.kt
+++ b/clients/graphql-kotlin-client-jackson/src/main/kotlin/com/expediagroup/graphql/client/jackson/GraphQLClientJacksonSerializer.kt
@@ -18,6 +18,7 @@ package com.expediagroup.graphql.client.jackson
 
 import com.expediagroup.graphql.client.jackson.types.JacksonGraphQLResponse
 import com.expediagroup.graphql.client.serializer.GraphQLClientSerializer
+import com.expediagroup.graphql.client.types.GraphQLClientRequest
 import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.JavaType
 import com.fasterxml.jackson.databind.ObjectMapper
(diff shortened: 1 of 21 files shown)


clients/graphql-kotlin-client-jackson/src/main/kotlin/com/expediagroup/graphql/client/jackson/GraphQLClientJacksonSerializer.kt
```diff
@@ -35,7 +36,9 @@ class GraphQLClientJacksonSerializer(private val mapper: ObjectMapper = jacksonO
         mapper.enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)
     }
 
-    override fun serialize(request: Any): String = mapper.writeValueAsString(request)
```
dariuszkuc, line 38: Split it up so dont have to check whether it is a list in kotlinx serializer

clients/graphql-kotlin-client-serialization/src/main/kotlin/com/expediagroup/graphql/client/serialization/GraphQLClientKotlinxSerializer.kt
```diff
@@ -43,7 +44,14 @@ class GraphQLClientKotlinxSerializer(private val jsonBuilder: JsonBuilder.() ->
         encodeDefaults = true
     }
 
-    override fun serialize(request: Any): String = json.encodeToString(AnyKSerializer, request)
+    override fun serialize(request: GraphQLClientRequest<*>): String = json.encodeToString(requestSerializer(request), request)
+
+    override fun serialize(requests: List<GraphQLClientRequest<*>>): String {
+        val serializedRequests = requests.map { request ->
```
dariuszkuc, line 50: Kotlinx serializer is pretty bad with dynamic types - this was the simplest way I could think of how to correctly serialize list of generic types

smyrick (approved):

PR #1089, opened 2021-03-24
[build] update to use latest dependencies

dariuszkuc:


dariuszkuc:
Dependencies not updated in this PR:
* kotlinpoet -> new version generates code in [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors) that requires some generator changes
* dokka -> we should look into upgrading to new version
* publishing plugins -> one of the plugins was deprecated so should tackle the upgrade together

smyrick (approved):

Issue #1097, opened 2021-03-27
GraphQL endpoints observe the Reactive Spring Boot context path, but Playground tries to reach /graphql


joaodias14:
**Library Version**
3.6.8

**Describe the bug**
GraphQL endpoints observe the Reactive Spring Boot context path, but Playground tries to reach /graphql without including the context path. I haven't tried it but I would say the same is true for the Servlet stack.

**To Reproduce**
1. Create a Reactive Spring Boot application and configure its context path using `spring.webflux.base-path`;
2. Access Playground in the browser using `{domain}/{context-path}/playground`;
3. Playground opens but with errors;
4. Open the browser console and check that it is trying to reach `{domain}/graphql` instead of `{domain}/{context-path}/graphql`.

**Expected behavior**
Playground should reach `{domain}/{context-path}/graphql` instead of `{domain}/graphql`.

dariuszkuc:
Resolved in https://github.com/ExpediaGroup/graphql-kotlin/pull/1098

PR #1098, opened 2021-03-27
Reactive context-path added to Playground

joaodias14:
### :pencil: Description
Reactive context-path added to Playground Route configuration.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1097

### DIFF
--- a/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/PlaygroundRouteConfiguration.kt
+++ b/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/PlaygroundRouteConfiguration.kt
@@ -32,13 +32,22 @@ import org.springframework.web.reactive.function.server.html
 @Configuration
 class PlaygroundRouteConfiguration(
     private val config: GraphQLConfigurationProperties,
-    @Value("classpath:/graphql-playground.html") private val playgroundHtml: Resource
+    @Value("classpath:/graphql-playground.html") private val playgroundHtml: Resource,
+    @Value("server.servlet.context-path") private val servletContextPath: String?,
+    @Value("spring.webflux.base-path") private val reactiveContextPath: String?
 ) {
 
     private val body = playgroundHtml.inputStream.bufferedReader().use { reader ->
+        val contextPath = when {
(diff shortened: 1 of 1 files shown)


servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/PlaygroundRouteConfiguration.kt
```diff
@@ -32,13 +32,22 @@ import org.springframework.web.reactive.function.server.html
 @Configuration
 class PlaygroundRouteConfiguration(
     private val config: GraphQLConfigurationProperties,
-    @Value("classpath:/graphql-playground.html") private val playgroundHtml: Resource
+    @Value("classpath:/graphql-playground.html") private val playgroundHtml: Resource,
+    @Value("server.servlet.context-path") private val servletContextPath: String?,
```
dariuszkuc, line 36: we only support webflux in the `spring-server` so this should be removed

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/PlaygroundRouteConfiguration.kt
```diff
@@ -32,13 +32,17 @@ import org.springframework.web.reactive.function.server.html
 @Configuration
 class PlaygroundRouteConfiguration(
     private val config: GraphQLConfigurationProperties,
-    @Value("classpath:/graphql-playground.html") private val playgroundHtml: Resource
+    @Value("classpath:/graphql-playground.html") private val playgroundHtml: Resource,
+    @Value("spring.webflux.base-path") private val contextPath: String?
```
dariuszkuc, line 36: does this actually work? I think it should be `@Value("\${spring.webflux.base-path}")`

dariuszkuc:
Thanks for fixing this!

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/PlaygroundRouteConfiguration.kt
joaodias14: Yes, you are right. I will update it.

dariuszkuc (approved):

PR #1116, opened 2021-04-09
Abstract out FlowSubscriptionExecutionStrategy

josephlbarnett:
Put common methods in a base class and add an implementation
for a NativeFlowSubscriptionExecutionStrategy that returns
Flows instead of Publishers.

### :pencil: Description
Refactors FlowSubscriptionStrategy to allow for a NativeFlowSubscriptionStrategy
without too much copy/paste (although the base bits are still copy/pasting from graphql-java).
Possibly FlowSubscriptionStrategy could be renamed PublisherFlowSubscriptionStrategy but
not sure what makes the most sense there.  Also the new test is basically a copy/paste of the
existing test, with `getData<Publisher<>>` changed to `getData<Flow<>>`, not sure if there's
a good way to clean that up.

### :link: Related Issues
#358
#972


dariuszkuc:
Reverted the changes from #972 in #1120

PR #1134, opened 2021-04-27
Fix Classname of ktor client

huehnerlady:
### :pencil: Description

While updating the graphql client to the new version I ran into a small confusion with the documentation. After some research I would assume that there was a copy/paste error and the Class was not renamed. So I thought I will create a Pull Request for this :) 

Client serialization

### DIFF
--- a/website/docs/client/client-serialization.mdx
+++ b/website/docs/client/client-serialization.mdx
@@ -366,7 +366,7 @@ graphql {
 By default, `ServiceLoader` mechanism will load the first available GraphQL client serializer from the classpath.
 
 ```kotlin
-val client = GraphQLWebClient(
+val client = GraphQLKtorClient(
   url = "http://localhost:8080/graphql"
   serializer = GraphQLClientKotlinxSerializer()
 )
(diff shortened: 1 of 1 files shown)


website/docs/client/client-serialization.mdx
```diff
@@ -366,7 +366,7 @@ graphql {
 By default, `ServiceLoader` mechanism will load the first available GraphQL client serializer from the classpath.
 
 ```kotlin
-val client = GraphQLWebClient(
+val client = GraphQLKtorClient(
   url = "http://localhost:8080/graphql"
```
dariuszkuc, line 370: Nice catch (yeah it looks like copy and paste)! Can you also fix the constructor (Spring one takes String url vs URL object on Ktor)? i.e 

```kotlin
val client = GraphQLKtorClient(
  url = URL("http://localhost:8080/graphql")
```

I'm thinking for 5.0 we should update constructors on those clients to just accept a String...

huehnerlady: I did not see that, but amended it now :)

dariuszkuc (approved):

PR #1136, opened 2021-04-28
Client Generation - Allow schema to be on the classpath

AlexRiedler:
### :pencil: Description

Simply allow loading the schema file from your classpath, this allows you to include a maven dependency that stores your graphql SDLs as a package.

```xml
<plugin>
	<groupId>com.expediagroup</groupId>
	<artifactId>graphql-kotlin-maven-plugin</artifactId>
	<version>4.0.0-SNAPSHOT</version>
	<dependencies>
		<dependency>
			<groupId>com.expedia.schemas</groupId>
			<artifactId>schemas</artifactId>
		</dependency>
	</dependencies>
	<executions>
		<execution>
			<goals>
				<goal>generate-client</goal>
			</goals>
			<configuration>
				<packageName>com.expedia.graphql.generated.foobarservice</packageName>
				<schemaFile>expedia-schemas/foobar-service/schema.graphql</schemaFile>
				<queryFileDirectory>${project.basedir}/src/main/resources/queries</queryFileDirectory>
			</configuration>
		</execution>
	</executions>
</plugin>

```

I would love to understand how to test this for Gradle as well since it changed, I have tested for maven only so far!

Feedback appreciated :)

### DIFF
--- a/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
+++ b/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
@@ -21,13 +21,7 @@ import com.expediagroup.graphql.plugin.client.generator.GraphQLClientGeneratorCo
 import com.expediagroup.graphql.plugin.client.generator.GraphQLSerializer
 import com.expediagroup.graphql.plugin.client.generator.GraphQLScalar
 import com.squareup.kotlinpoet.FileSpec
-import graphql.schema.idl.SchemaParser
 import java.io.File
-import java.io.Reader
-import java.io.StringReader
-import java.net.URI
-import java.nio.charset.Charset
-import java.nio.file.Files
 
 /**
  * Generate GraphQL client data classes from specified queries and target schema.
@@ -37,7 +31,7 @@ fun generateClient(
     allowDeprecated: Boolean = false,
     customScalarsMap: List<GraphQLScalar> = emptyList(),
     serializer: GraphQLSerializer = GraphQLSerializer.JACKSON,
-    schema: String, // this is the file location, not the contents
+    schemaPath: String, // this is the file location, not the contents
     queries: List<File>
 ): List<FileSpec> {
     val customScalars = customScalarsMap.associateBy { it.scalar }
@@ -47,19 +41,6 @@ fun generateClient(
         customScalarMap = customScalars,
         serializer = serializer
     )
-    val graphQLSchema = SchemaParser().parse(getSchemaReader(schema))
-    val generator = GraphQLClientGenerator(graphQLSchema, config)
+    val generator = GraphQLClientGenerator(schemaPath, config)
     return generator.generate(queries)
 }
-
-private fun getSchemaReader(schemaPath: String): Reader =
-    if (schemaPath.startsWith("classpath://")) {
-        val resource = object {}.javaClass.getResource(URI(schemaPath).path) ?: throw RuntimeException("specified GraphQL schema not found, $schemaPath")
-        StringReader(resource.readText())
-    } else {
-        val file = File(schemaPath)
-        if (!file.isFile) {
-            throw RuntimeException("specified GraphQL schema is not a file, $schemaPath")
-        }
-        Files.newBufferedReader(file.toPath(), Charset.defaultCharset())
-    }
--- a/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
+++ b/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
@@ -31,7 +31,7 @@ fun generateClient(
     allowDeprecated: Boolean = false,
     customScalarsMap: List<GraphQLScalar> = emptyList(),
     serializer: GraphQLSerializer = GraphQLSerializer.JACKSON,
-    schemaPath: String, // this is the file location, not the contents
+    schemaPath: String,
     queries: List<File>
 ): List<FileSpec> {
     val customScalars = customScalarsMap.associateBy { it.scalar }
--- a/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
+++ b/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
@@ -23,6 +23,11 @@ import com.expediagroup.graphql.plugin.client.generator.GraphQLScalar
 import com.squareup.kotlinpoet.FileSpec
 import graphql.schema.idl.SchemaParser
 import java.io.File
+import java.io.Reader
+import java.io.StringReader
+import java.net.URI
+import java.nio.charset.Charset
+import java.nio.file.Files
 
 /**
  * Generate GraphQL client data classes from specified queries and target schema.
@@ -32,7 +37,7 @@ fun generateClient(
     allowDeprecated: Boolean = false,
     customScalarsMap: List<GraphQLScalar> = emptyList(),
     serializer: GraphQLSerializer = GraphQLSerializer.JACKSON,
-    schema: File,
+    schema: String, // this is the file location, not the contents
     queries: List<File>
 ): List<FileSpec> {
     val customScalars = customScalarsMap.associateBy { it.scalar }
@@ -42,7 +47,19 @@ fun generateClient(
         customScalarMap = customScalars,
         serializer = serializer
     )
-    val graphQLSchema = SchemaParser().parse(schema)
+    val graphQLSchema = SchemaParser().parse(getSchemaReader(schema))
     val generator = GraphQLClientGenerator(graphQLSchema, config)
     return generator.generate(queries)
 }
+
+private fun getSchemaReader(schemaPath: String): Reader =
+    if (schemaPath.startsWith("classpath://")) {
+        val resource = object {}.javaClass.getResource(URI(schemaPath).path) ?: throw RuntimeException("specified GraphQL schema not found, $schemaPath")
+        StringReader(resource.readText())
+    } else {
+        val file = File(schemaPath)
+        if (!file.isFile) {
+            throw RuntimeException("specified GraphQL schema is not a file, $schemaPath")
+        }
+        Files.newBufferedReader(file.toPath(), Charset.defaultCharset())
+    }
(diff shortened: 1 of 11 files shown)


plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
```diff
@@ -32,7 +37,7 @@ fun generateClient(
     allowDeprecated: Boolean = false,
     customScalarsMap: List<GraphQLScalar> = emptyList(),
     serializer: GraphQLSerializer = GraphQLSerializer.JACKSON,
-    schema: File,
+    schema: String, // this is the file location, not the contents
```
dariuszkuc, line 40: lets rename this field to `schemaPath: String` and drop the comment

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
```diff
@@ -42,7 +47,19 @@ fun generateClient(
...
     )
-    val graphQLSchema = SchemaParser().parse(schema)
+    val graphQLSchema = SchemaParser().parse(getSchemaReader(schema))
     val generator = GraphQLClientGenerator(graphQLSchema, config)
     return generator.generate(queries)
 }
+
+private fun getSchemaReader(schemaPath: String): Reader =
+    if (schemaPath.startsWith("classpath://")) {
+        val resource = object {}.javaClass.getResource(URI(schemaPath).path) ?: throw RuntimeException("specified GraphQL schema not found, $schemaPath")
```
dariuszkuc, line 57: i think anonymous objects like this might be throwing some detekt violations -> instead lets move the logic of reading from classpath to the `GraphQLClientGenerator` (i.e. change its constructor from accepting type registry to a path and initialize type registry in init block)

```
private val graphQLSchema: TypeDefinitionRegistry
init {
    graphQLSchema = parseSchema(schemaPath)
}

private fun parseSchema(path: String): TypeDefinitionRegistry {
    val schemaFile = File(path)
    return if (schemaFile.isFile) {
        SchemaParser().parse(schemaFile)
    } else {
        val schemaInputStream = this.javaClass.classLoader.getResourceAsStream(path) ?: throw RuntimeException("specified schema file=$path does not exist")
       SchemaParser().parse(schemaInputStream)
    }
}
```

we'll need some unit tests around this

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
```diff
@@ -42,7 +47,19 @@ fun generateClient(
...
         serializer = serializer
     )
-    val graphQLSchema = SchemaParser().parse(schema)
+    val graphQLSchema = SchemaParser().parse(getSchemaReader(schema))
     val generator = GraphQLClientGenerator(graphQLSchema, config)
     return generator.generate(queries)
 }
+
+private fun getSchemaReader(schemaPath: String): Reader =
+    if (schemaPath.startsWith("classpath://")) {
```
dariuszkuc, line 56: `classpath://` is not a valid URL protocol so we should not code against this (yes Spring supports it but it is non-standard)

plugins/graphql-kotlin-maven-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/maven/GenerateClientAbstractMojo.kt
```diff
@@ -99,17 +99,23 @@ abstract class GenerateClientAbstractMojo : AbstractMojo() {
 
     override fun execute() {
         log.debug("generating GraphQL client")
-        val graphQLSchemaFile = schemaFile ?: File(project.build.directory, "schema.graphql")
-        validateGraphQLSchemaExists(graphQLSchemaFile)
+
+        val schemaPath =
```
dariuszkuc, line 103: you can rewrite this as 
```suggestion
        val schemaPath = schemaFile ?: File(project.build.directory, "schema.graphql").path
```
which avoids the `!!`

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/tasks/AbstractGenerateClientTask.kt
```diff
@@ -66,7 +66,7 @@ abstract class AbstractGenerateClientTask : DefaultTask() {
      */
     @InputFile
     @Optional
-    val schemaFile: RegularFileProperty = project.objects.fileProperty()
+    val schemaFile: Property<String> = project.objects.property(String::class.java)
```
dariuszkuc, line 69: there already is `schemaFileName` property for that use case, this should remain a `RegularFileProperty`

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/tasks/AbstractGenerateClientTask.kt
```diff
@@ -147,13 +147,10 @@ abstract class AbstractGenerateClientTask : DefaultTask() {
         logger.debug("generating GraphQL client")
 
         val graphQLSchema = when {
-            schemaFile.isPresent -> schemaFile.get().asFile
-            schemaFileName.isPresent -> File(schemaFileName.get())
+            schemaFile.isPresent -> schemaFile.get() as String
```
dariuszkuc, line 150: lets rename the enclosing variable `graphQLSchema` -> `graphQLSchemaPath`

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/parameters/GenerateClientParameters.kt
```diff
@@ -37,7 +37,7 @@ interface GenerateClientParameters : WorkParameters {
     /** Type of JSON serializer that will be used to generate the data classes. */
     val serializer: Property<GraphQLSerializer>
     /** GraphQL schema file that will be used to generate client code. */
-    val schemaFile: Property<File>
+    val schemaFile: Property<String>
```
dariuszkuc, line 40: lets also update javadoc to reflect this is a path to a schema file

```suggestion
    val schemaPath: Property<String>
```

dariuszkuc (changes requested):
Thanks for raising this.

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generateClient.kt
```diff
@@ -32,7 +31,7 @@ fun generateClient(
     allowDeprecated: Boolean = false,
     customScalarsMap: List<GraphQLScalar> = emptyList(),
     serializer: GraphQLSerializer = GraphQLSerializer.JACKSON,
-    schema: File,
+    schemaPath: String, // this is the file location, not the contents
```
dariuszkuc, line 34: lets remove the comment as it is unnecessary

```suggestion
    schemaPath: String,
```

plugins/client/graphql-kotlin-client-generator/src/test/kotlin/com/expediagroup/graphql/plugin/client/generator/GraphQLTestUtils.kt
```diff
@@ -52,12 +50,7 @@ internal fun locateTestFiles(directory: File): Pair<List<File>, Map<String, File
     return queries to expectedFiles
 }
 
-internal fun testSchema(): TypeDefinitionRegistry {
-    val schemaFileStream = ClassLoader.getSystemClassLoader().getResourceAsStream("testSchema.graphql") ?: throw RuntimeException("unable to locate test schema")
-    return schemaFileStream.use {
-        SchemaParser().parse(schemaFileStream)
-    }
-}
+internal fun testSchema(): String = "testSchema.graphql"
```
dariuszkuc, line 53: since this is a constant string lets change it to 

```suggestion
internal const val TEST_SCHEMA = "testSchema.graphql"
```

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt
```diff
@@ -104,15 +104,15 @@ class GraphQLGradlePlugin : Plugin<Project> {
                         introspectSchemaTask.headers.convention(project.provider { extension.clientExtension.headers })
                         introspectSchemaTask.timeoutConfig.convention(project.provider { extension.clientExtension.timeoutConfig })
                         generateClientTask.dependsOn(introspectSchemaTask.path)
-                        generateClientTask.schemaFile.convention(introspectSchemaTask.outputFile)
+                        generateClientTask.schemaFile.convention(introspectSchemaTask.outputFile.get())
```
dariuszkuc, line 107: lets revert changes in this file as we should rely on the provider

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/GraphQLClientGenerator.kt
```diff
@@ -196,6 +202,16 @@ class GraphQLClientGenerator(
         val rootType = operationNames[operationDefinition.operation.name]
         return graphQLSchema.getType(rootType).get() as ObjectTypeDefinition
     }
+
+    private fun parseSchema(path: String): TypeDefinitionRegistry {
+        val schemaFile = File(path)
+        return if (schemaFile.isFile) {
+            SchemaParser().parse(schemaFile)
+        } else {
+            val schemaInputStream = this.javaClass.classLoader.getResourceAsStream(path) ?: throw RuntimeException("specified schema file=$path does not exist")
```
dariuszkuc, line 211: lets extract this exception to explicit `SchemaUnavailableException` (or something like that) under `com.expediagroup.graphql.plugin.client.generator.exceptions`

dariuszkuc (changes requested):
Can we also add a unit test for `GraphQLClientGenerator` to verify it will blow up with the exception when invalid path is specified?

dariuszkuc:
> One thing is the user no longer needs to specify classpath:// ; not sure if we should require that on the 2nd path; what do you think?

Nah, current logic of just checking if it is a file and/or resource on classpath should be fine. We could technically go fancy with support for URLs (so you could point to a http resource) but that might be an overkill.

Issue #1151, opened 2021-05-18
[client] support optional undefined arguments


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
When using parameterized arguments, optional arguments can be omitted (undefined), `null` or have a specific value. Currently `graphql-kotlin-client`s do not support undefined arguments which requires specifying redundant queries to handle   case when arguments should be skipped/omitted.

**Describe the solution you'd like**
Generated GraphQL client code should provide mechanism for handling optional undefined arguments.

**Describe alternatives you've considered**
Redundant queries can be specified that omit the unnecessary arguments but it means that we might end up writing an extra query for every possible permutation of optional argument.

**Additional context**
See [discussion](https://github.com/ExpediaGroup/graphql-kotlin/discussions/1147) for more details

dariuszkuc:
Hello 👋 
We are always open to contributions! 

I was playing around with optional input serializers and got a draft PR open -> https://github.com/ExpediaGroup/graphql-kotlin/pull/1158. Sadly it looks like we won't be able to support this functionality for `kotlinx-serialization` until https://github.com/Kotlin/kotlinx.serialization/issues/1091 is resolved (or someone has a better idea how to solve it).

jozef-slezak:
@dariuszkuc, that is great! We just need Jackson. Is it possible to separate Jackson only solution from `kotlinx-serialization`? We can test that on our use case.

PR #1158, opened 2021-05-25
[client] support optional input for jackson serializer

dariuszkuc:
### :pencil: Description

Wrap optional input arguments in new `OptionalInput` sealed class which can either be `Undefined` (omitted) or `Defined` (value or explicit null). Unfortunately it looks like we won't be able to support this functionality for `kotlinx-serialization` until https://github.com/Kotlin/kotlinx.serialization/issues/1091 is resolved.

### :link: Related Issues

Related: https://github.com/ExpediaGroup/graphql-kotlin/issues/1151

### DIFF
--- a/clients/graphql-kotlin-client-jackson/build.gradle.kts
+++ b/clients/graphql-kotlin-client-jackson/build.gradle.kts
@@ -14,7 +14,7 @@ tasks {
                 limit {
                     counter = "INSTRUCTION"
                     value = "COVEREDRATIO"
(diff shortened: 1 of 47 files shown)


smyrick:
Code changes look good, but this would actually be a breaking change to the generated code the clients would use. Can we instead add a plugin flag to enable this feature so we don't force a breaking change. 

Then for 5.0 we can enable it by default or even just remove the config all together

smyrick (approved):

PR #1159, opened 2021-05-25
pass code registry to all directive environments

koenpunt:
### :pencil: Description

To override datafetchers for all fields in an object using a directive, the environment should expose the `GraphQLCodeRegistry`.

Or at least the only way of overriding datafetchers from a directive on an object was by iterating over the fields and setting the datafetcher.

Something like this I expect to work:

```kt
override fun onObject(environment: KotlinSchemaDirectiveEnvironment<GraphQLObjectType>): GraphQLObjectType {
    val obj = environment.element
    val reg = environment.codeRegistry

    val expressionArgument = environment.directive.getArgument("expression")
    val expressionString = expressionArgument?.value as? String ?: return obj

    val authorizationManager = authorizationManagerForExpression(expressionString)

    obj.fieldDefinitions.forEach { field ->
        val fetcher = reg.getDataFetcher(obj, field)
        val authDataFetcher = AuthDataFetcher(fetcher, authorizationManager)
        reg.dataFetcher(obj, field, authDataFetcher)
    }

    return obj
}
```

Which I think should be possible with the provided change.

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/directives/KotlinDirectiveWiringFactory.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/directives/KotlinDirectiveWiringFactory.kt
@@ -35,7 +35,7 @@ open class KotlinDirectiveWiringFactory(
     /**
      * Wire up the directive based on the GraphQL type.
      */
-    fun onWire(graphQLSchemaElement: GraphQLSchemaElement, coordinates: FieldCoordinates? = null, codeRegistry: GraphQLCodeRegistry.Builder? = null): GraphQLSchemaElement {
+    fun onWire(graphQLSchemaElement: GraphQLSchemaElement, coordinates: FieldCoordinates? = null, codeRegistry: GraphQLCodeRegistry.Builder): GraphQLSchemaElement {
         if (graphQLSchemaElement !is GraphQLDirectiveContainer) return graphQLSchemaElement
 
         return wireDirectives(graphQLSchemaElement, coordinates, graphQLSchemaElement.getAllDirectives(), codeRegistry)
@@ -51,7 +51,7 @@ open class KotlinDirectiveWiringFactory(
         element: GraphQLDirectiveContainer,
         coordinates: FieldCoordinates?,
         directives: List<GraphQLDirective>,
-        codeRegistry: GraphQLCodeRegistry.Builder?
+        codeRegistry: GraphQLCodeRegistry.Builder
     ): GraphQLDirectiveContainer {
         var modifiedObject = element
         for (directive in directives) {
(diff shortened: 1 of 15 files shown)


dariuszkuc:
Hello 👋 
Can you elaborate on your use case? Why would you annotate the object vs the fields? i.e. given a schema (assuming there is `@authenticated` directive)

```
type Query {
  foo: Foo @authenticated
}

type Foo {
  bar: Int!
  baz: String!
}
```

In order to retrieve `Foo` you have to execute `foo` resolver which is already marked as `@authenticated`

koenpunt:
For the use case of an `@auth` directive; if a type is used multiple times in the schema, then for every time the type is used the field should be annotated. With support for adding it to the object you don't have to;

<table>
<tr>
<td>

```graphql
type Query {
  other: OtherType
  foo: Foo @auth
}

type OtherType {
  foo: Foo @auth
}

type Foo {
  bar: Int!
  baz: String!
}
```

</td>
<td>VS</td>
<td>

```graphql
type Query {
  other: OtherType
  foo: Foo
}

type OtherType {
  foo: Foo
}

type Foo @auth {
  bar: Int!
  baz: String!
}
```

</td>
</tr>
</table>

Or if all the fields of a type require authorization, except one, you could annotate the object, and the field with the exception;

```graphql
type OtherType @auth {
  foo: Foo
  something: String
  other: OtherType @public
}
```

The latter example would probably require a less trivial implementation, but should still be doable.

The Node implementation of Apollo Server also supports a similar directive, as documented here: https://www.apollographql.com/docs/apollo-server/schema/creating-directives/#enforcing-access-permissions

And apart from authorization, this could also be used to map all fields through a specific datafetcher, for whatever purpose it might be necessary.

generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/directives/KotlinDirectiveWiringFactoryTest.kt
```diff
@@ -224,19 +224,19 @@ class KotlinDirectiveWiringFactoryTest {
...
-            .type { context, visitor -> context.thisNode().accept(context, visitor) }
-            .description("My Field Description")
-            .withDirective(graphQLLowercaseDirective)
-            .build()
-
-        assertFailsWith(InvalidSchemaDirectiveWiringException::class) {
-            SimpleWiringFactory().onWire(graphQLSchemaElement = myTestField, coordinates = mockk(), codeRegistry = null)
-        }
-    }
+//    @Test
```
dariuszkuc, line 227: can you remove this commented out code

dariuszkuc (changes requested):

generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/directives/KotlinDirectiveWiringFactoryTest.kt
koenpunt: Done

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/directives/KotlinDirectiveWiringFactory.kt
```diff
@@ -65,7 +65,8 @@ open class KotlinDirectiveWiringFactory(
             } else {
                 KotlinSchemaDirectiveEnvironment(
                     element = modifiedObject,
-                    directive = directive
+                    directive = directive,
+                    codeRegistry = codeRegistry ?: throw InvalidSchemaDirectiveWiringException("Unable to wire directive on an object due to a missing code registry")
```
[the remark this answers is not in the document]
koenpunt: Done

dariuszkuc (approved):

dariuszkuc:
Bump to major as it is a breaking change to hook signature.

Issue #1084, opened 2021-03-21
Examples build failing with exception


smyrick:
**Library Version**
Latest master branch

**Describe the bug**
The build is failing to compile the library on the task `client-examples:startWireMock`.

Running the gradle task from the command line doesn't trigger a failure. An exception is logged but the build is still marked as successful. Running from the IDE fails and stops the build.

```
Exception in thread "ktor-cio-dispatcher-worker-2" java.lang.NoClassDefFoundError: kotlinx/coroutines/CoroutinesInternalError
        at kotlinx.coroutines.DispatchedTask.handleFatalException(DispatchedTask.kt:144)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:115)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
Caused by: java.lang.ClassNotFoundException: kotlinx.coroutines.CoroutinesInternalError
        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:471)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:588)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
        ... 6 more
```

**To Reproduce**
```
cd examples
gw clean assemble
```

**Expected behavior**
There is no exception logged.

dariuszkuc:
While I did encounter those exceptions occasionally they were not resulting in a build failure.

smyrick:
Correct, they don't result in a failure if run from the command line, but do if run from my IDE

huehnerlady:
I have the same issue. I found out that I can reproduce them by doing a 'clean' beforehand.

While this does not result in a build failure for me either, I do feel this should be looked at

PR #1162, opened 2021-05-26
[plugin] change tasks HttpClient engine from CIO to Apache

dariuszkuc:
### :pencil: Description

It looks like using CIO engine causes some issues with the classpath (https://github.com/ExpediaGroup/graphql-kotlin/issues/1084). Unsure what is the root cause of the classpath issue - it looks like CIO engine relies on coroutines-core `native-mt` vs regular artifact that we pull in. Switching to a different engine seems to resolve the classpath issue.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1084


dariuszkuc:
Ugh forgot that we build and release using Java 8... that new HttpClient is available from Java 11.

I'll update the PR to use other engine.

smyrick (approved):

PR #1170, opened 2021-05-29
WIP: Demo returning query root

smyrick:
### :pencil: Description
Still WIP. Demo of how we could return the query root object from a mutation field. This would actually be a backwards compatible feature addition. I would just want to clean some things up and add tests

Screen Shot 2021-05-29 at 12 00 10 PM

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/discussions/1169


smyrick:
See details here: https://github.com/ExpediaGroup/graphql-kotlin/discussions/1169#discussioncomment-822669

smyrick:
This is now supported with `@GraphQLType`

https://github.com/ExpediaGroup/graphql-kotlin/discussions/1169#discussioncomment-3882078

PR #1171, opened 2021-06-01
[generator] remove extended scalars

dariuszkuc:
### :pencil: Description

Extended scalars were deprecated in `graphql-java` v15 in favor of separate `graphql-java-extended-scalars` project. Extended scalars are problematic as clients have to explicitly know how to handle them and what they represent. Due to the all the drawbacks, instead of automatically providing support for deprecated functionality, users should explicitly opt-in into this functionality by using custom hooks.


smyrick (approved):

smyrick:
I approve of the change but as you mentioned this would be a major change. Do we want to cut 5.0 already?

dariuszkuc:
We already merged a breaking change in https://github.com/ExpediaGroup/graphql-kotlin/pull/1159 which probably wouldn't affect that many people but it is a breaking change nonetheless.

PR #1190, opened 2021-06-21
Add local schema file support to Gradle plugin

foo4u:
### :pencil: Description
Add support to the Gradle plugin for client generation from a local schema file.

### :link: Related Issues

#1136: Allow schema to be on the classpath

### DIFF
--- a/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt
+++ b/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt
@@ -59,6 +59,8 @@ open class GraphQLPluginClientExtension {
     var endpoint: String? = null
     /** GraphQL server SDL endpoint that will be used to download schema. Alternatively you can run introspection query against [endpoint]. */
     var sdlEndpoint: String? = null
+    /** GraphQL schema file location. Can be used instead of [endpoint] or [sdlEndpoint]. */
+    var schemaFile: File? = null
     /** Target package name to be used for generated classes. */
     var packageName: String? = null
     /** Optional HTTP headers to be specified on an introspection query or SDL request. */
--- a/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt
+++ b/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt
@@ -115,8 +115,11 @@ class GraphQLGradlePlugin : Plugin<Project> {
                         generateClientTask.dependsOn(downloadSDLTask.path)
                         generateClientTask.schemaFile.convention(downloadSDLTask.outputFile)
                     }
+                    extension.clientExtension.schemaFile != null -> {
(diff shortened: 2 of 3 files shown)


plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt
```diff
@@ -59,6 +59,8 @@ open class GraphQLPluginClientExtension {
     var endpoint: String? = null
     /** GraphQL server SDL endpoint that will be used to download schema. Alternatively you can run introspection query against [endpoint]. */
     var sdlEndpoint: String? = null
+    /** GraphQL schema file location. Can be used instead of [endpoint] or [sdlEndpoint]. */
+    var schemaFile: File? = null
```
smyrick, line 63: Do we want to use a file here or load the file for users and just accept a string? @dariuszkuc any comments?

dariuszkuc:
Hello 👋 
Thanks for raising this but this is already supported on the individual Gradle tasks (see `schemaFile` and `schemaFileName` on [graphqlGenerateClient] task). `graphql` extension is just a convenient shortcut for accessing some of the common functionality (like downloading/introspecting schema followed by generation of the the client using that info). In case of referencing local schema file we recommend to use the task directly.

foo4u:
@dariuszkuc, that doesn't seem to be true. If you provide a `graphql { client { .. } }` configuration, either `sdlEndpoint` or `endpoint` must be accessible otherwise client code generation fails.

dariuszkuc:
Yes, using `graphql` extension requires one of the endpoints to be specified. What I meant is instead of using

```kotlin
graphql {  client { ... } }
```

instead configure the task directly (https://expediagroup.github.io/graphql-kotlin/docs/plugins/gradle-plugin-usage#generating-client)

```kotlin
val graphqlGenerateClient by tasks.getting(GraphQLGenerateClientTask::class) {
  packageName.set("com.example.generated")
  schemaFileName.set("mySchema.graphql")
}
```

If referencing a local schema file (when generating a client) is a common pattern we can revisit and add `schemaFileName: String` to the extension.

dariuszkuc:
reopening after talking about it with @smyrick

@foo4u can you update docs to reflect changes in the extension?
- https://github.com/ExpediaGroup/graphql-kotlin/blob/master/plugins/graphql-kotlin-gradle-plugin/README.md
- https://github.com/ExpediaGroup/graphql-kotlin/blob/master/website/docs/plugins/gradle-plugin-tasks.mdx

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLPluginExtension.kt
dariuszkuc: Yes, this should be `var schemaFileName: String? =null` (matching https://github.com/ExpediaGroup/graphql-kotlin/blob/master/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/tasks/AbstractGenerateClientTask.kt#L60). Lets also update javadocs for the `endpoint`/`sdlEndpoint` to mention that you can specify file directly as well.

dariuszkuc:
@foo4u will you be able to make the changes?

dariuszkuc:
Superseded by https://github.com/ExpediaGroup/graphql-kotlin/pull/1212

PR #1193, opened 2021-06-22
Add GraphQLAbstractType to support custom schema types from libraries

smyrick:
### :pencil: Description

```kotlin
```

```graphql
  foo: String!
}

  foo: String!
}

  foo: String!
  bar: String!
}

```

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/annotations/GraphQLAbstractType.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/annotations/GraphQLAbstractType.kt
@@ -1,0 +1,6 @@
+package com.expediagroup.graphql.generator.annotations
+
+import kotlin.reflect.KClass
+
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.FUNCTION)
+annotation class GraphQLAbstractType(val possibleTypes: Array<KClass<*>>, val name: String, val description: String)
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
@@ -61,8 +61,10 @@ internal fun KClass<*>.isInterface(): Boolean =
     this.java.isInterface || this.isAbstract || this.isSealed
 
 internal fun KClass<*>.isUnion(): Boolean =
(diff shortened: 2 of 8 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/annotations/GraphQLAbstractType.kt
```diff
@@ -0,0 +1,6 @@
+package com.expediagroup.graphql.generator.annotations
+
+import kotlin.reflect.KClass
+
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.FUNCTION)
+annotation class GraphQLAbstractType(val possibleTypes: Array<KClass<*>>, val name: String, val description: String)
```
dariuszkuc, line 6: guess technically this would apply also to interfaces but since they have to declare some common fields -> should this be just applicable to the unions?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
```diff
@@ -61,7 +61,7 @@ internal fun KClass<*>.isInterface(): Boolean =
     this.java.isInterface || this.isAbstract || this.isSealed
 
 internal fun KClass<*>.isUnion(): Boolean =
-    this.isInterface() && this.declaredMemberProperties.isEmpty() && this.declaredMemberFunctions.isEmpty()
+    (this.isInterface() && this.declaredMemberProperties.isEmpty() && this.declaredMemberFunctions.isEmpty())  || this.isInstance(Any::class)
```
dariuszkuc, line 64: this is a little too loose -> in 99.9% of cases using `Any` as return type should blow up, this should check if it is `Any` and has annotation

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/annotations/GraphQLAbstractType.kt
smyrick: Yea, as I was walking through more required changes, it is going to only work with unions. We will need a different annotation that we can put on both the classes and the return type to implement custom interfaces

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
smyrick: Working on a better check

PR #1194, opened 2021-06-23
Add GraphQLUnion for custom schema modification

smyrick:
### :pencil: Description

```kotlin
```

```graphql
  foo: String!
}

  foo: String!
}

  foo: String!
  bar: String!
}

```


generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/internal/types/GenerateUnionTest.kt
```diff
@@ -94,4 +98,15 @@ internal class GenerateUnionTest : TypeTestHelper() {
         assertEquals(1, result.directives.size)
         assertEquals("simpleDirective", result.directives.first().name)
     }
+
+    @Test
```
HGF, line 102: should also add a test when the annotation has an empty `possibleTypes` along with `name`

generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/test/integration/CustomAbstractTypeTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        fun odd(first: Boolean): Any = if (first) One("1") else Three("3")
+
+        @GraphQLUnion(name = "Number", possibleTypes = [One::class, Two::class, Three::class, Four::class])
+        fun number(): Any = One("1")
+
+        @GraphQLUnion(name = "Number", possibleTypes = [One::class, Two::class, Three::class, Four::class])
+        fun listNumbers(): List<Any> = listOf(One("1"), Two("2"))
+    }
+
+    class InvalidQuery {
```
HGF, line 74: Why are these 2 invalid? The return type is still `Any`, the code is just only ever returning 1 of the possible types which is still valid.

HGF (changes requested):

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateUnion.kt
```diff
@@ -29,7 +31,29 @@ import graphql.schema.GraphQLUnionType
...
+        generateUnionFromKClass(generator, kClass)
+    }
+}
+
+private fun generateUnionFromAnnotation(generator: SchemaGenerator, kClass: KClass<*>, unionAnnotation: GraphQLUnion): GraphQLUnionType {
+    val builder = GraphQLUnionType.newUnionType()
+    builder.name(unionAnnotation.name)
+
+    val possibleTypes = unionAnnotation.possibleTypes.toList()
+    val types = if (kClass.isInstance(Any::class) || (possibleTypes.isNotEmpty() && generator.classScanner.getSubTypesOf(kClass).containsAll(possibleTypes))) {
```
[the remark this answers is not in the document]
smyrick: We only ever call `generateUnion` have we have called `kClass.isUnion` so it is already validated. This can be removed all together now

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/state/TypesCache.kt
```diff
@@ -80,25 +101,30 @@ internal class TypesCache(private val supportedPackages: List<String>) : Closeab
...
 
-        return when {
-            kClass.isListType() -> null
-            kClass.isSubclassOf(Enum::class) -> kClass.getSimpleName()
-            isTypeNotSupported(type) -> throw TypeNotSupportedException(type, supportedPackages)
-            else -> type.getSimpleName(cacheKey.inputType)
+            when {
+                kClass.isListType() -> null
+                kClass.isSubclassOf(Enum::class) -> kClass.getSimpleName()
+                kClass.isInstance(Any::class) -> null
```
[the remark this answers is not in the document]
smyrick: Yep, correct. Will remove now

generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/test/integration/CustomAbstractTypeTest.kt
smyrick: Each union on their own is valid, however when you have them both used in the schema then it should be invalid because you have the first one:

```graphql
union Number = One | Two
```

And the second one clashing definitions

```graphql
union Number = Three | Four
```

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/state/TypesCache.kt
```diff
@@ -64,6 +71,20 @@ internal class TypesCache(private val supportedPackages: List<String>) : Closeab
         return null
     }
 
+    private fun getCacheKey(type: KType, inputType: Boolean, annotations: List<Annotation> = emptyList()): TypesCacheKey {
+        val unionAnnotation = annotations.getUnionAnnotation()
+
+        return if (unionAnnotation != null) {
```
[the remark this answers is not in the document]
smyrick: This is handled already before when we call `kClass.isUnion`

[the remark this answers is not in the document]
HGF: I don't see where that is happening. This is called by `get(type: KType, inputType: Boolean, annotations: List<Annotation>)`, which is called by `objectFromReflection`, which is called by any number of potential objects and properties.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateUnion.kt
```diff
@@ -29,17 +31,45 @@ import graphql.schema.GraphQLUnionType
...
+        generateUnionFromAnnotation(generator, unionAnnotation)
+    } else {
+        generateUnionFromKClass(generator, kClass)
+    }
+}
+
+private fun generateUnionFromAnnotation(generator: SchemaGenerator, unionAnnotation: GraphQLUnion): GraphQLUnionType {
     val builder = GraphQLUnionType.newUnionType()
-    builder.name(kClass.getSimpleName())
+    builder.name(unionAnnotation.name)
```
HGF, line 44: I mentioned it in the comment on the units tests, but we should also ensure that `unionAnnotation.name` is not blank. Or will setting a blank name be an exception when building the type?

HGF: graphql-java already does actually along with validating the types are not empty. I think it would be better to rely on that and those specific exceptions: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/GraphQLUnionType.java#L55

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateUnion.kt
```diff
@@ -29,17 +31,45 @@ import graphql.schema.GraphQLUnionType
...
-    generator.classScanner.getSubTypesOf(kClass)
-        .map { generateGraphQLType(generator, it.createType()) }
+    val types = generator.classScanner.getSubTypesOf(kClass)
+
+    return createUnion(generator, builder, types, name)
+}
+
+private fun createUnion(generator: SchemaGenerator, builder: GraphQLUnionType.Builder, types: List<KClass<*>>, name: String): GraphQLUnionType {
+    if (types.isEmpty()) {
+        throw UnionContainsNoPossibleTypesException(name)
```
HGF, line 69: Same as other comment above:
graphql-java actually already validates the types are not empty. I think it would be better to rely on that and those specific exceptions for both cases: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/GraphQLUnionType.java#L55

dariuszkuc:
@smyrick can you add documentation?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/annotationExtensions.kt
```diff
@@ -30,6 +31,12 @@ internal fun KAnnotatedElement.getDeprecationReason(): String? = this.findAnnota
 
 internal fun KAnnotatedElement.isGraphQLIgnored(): Boolean = this.findAnnotation<GraphQLIgnore>() != null
 
+internal fun List<Annotation>.getUnionAnnotation(): GraphQLUnion? = this.filterIsInstance(GraphQLUnion::class.java).firstOrNull()
+
+/**
+ * The java code called here may return null
+ */
+@Suppress("RedundantNullableReturnType")
```
dariuszkuc, line 39: is this true? Deprecated `message` is non-nullable String so this would end up as an empty string and never null

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateUnion.kt
```diff
@@ -29,25 +30,49 @@ import graphql.schema.GraphQLUnionType
...
+
+private fun generateUnionFromAnnotation(generator: SchemaGenerator, unionAnnotation: GraphQLUnion): GraphQLUnionType {
     val builder = GraphQLUnionType.newUnionType()
-    builder.name(kClass.getSimpleName())
+    builder.name(unionAnnotation.name)
+    builder.description(unionAnnotation.description)
+
+    val possibleTypes = unionAnnotation.possibleTypes.toList()
+
+    return createUnion(generator, builder, possibleTypes, unionAnnotation.name)
```
dariuszkuc, line 48: with this new annotation it looks like it is not possible to specify union directives?

smyrick: Yea, not in this current state. I wonder though if there is some other work around we could do. Would this be better?

```kotlin
@GrapQLUnion(possibleTypes = [SharedModel::class, ServiceModel::class])
interface CustomUnion

@Suppress("CAST_NEVER_SUCCEEDS")
fun model(): CustomUnion = SharedModel("foo") as CustomUnion
```

dariuszkuc: since `SharedModel` does not implement this marker interface this won't work -> `Any` workaround was added specifically for that

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/state/TypesCache.kt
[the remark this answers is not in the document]
smyrick: @HGF Are you sure? Because calling `objectFromReflection` run the util `kClass.isUnion` which checks for `Any`.

Can you provide a unit test that fails?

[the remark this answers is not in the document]
smyrick: If I run with this code, the schema gets generated as an `Int` not a union

```
class InvalidReturnType {
        @GraphQLUnion(name = "Number", possibleTypes = [Int::class])
        fun number1(): Int = 1
    }
```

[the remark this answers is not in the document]
HGF: it's called in `generateGraphQLType` which is called after checking the cache, but the cache key will be wrong. I'll get a test written

HGF (approved):

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/annotations/GraphQLUnion.kt
```diff
@@ -0,0 +1,10 @@
+package com.expediagroup.graphql.generator.annotations
```
dariuszkuc, line 1: missing copyright

website/docs/schema-generator/writing-schemas/unions.md
```diff
@@ -43,3 +48,57 @@ type Query {
   whichHand(whichHand: String!): BodyPart!
 }
 ```
+
+## `@GraphQLUnion`
```
dariuszkuc, line 52: can you add `known limitations` (or issues/or something like that) and list that it is not possible to declare union directives?

dariuszkuc (approved):

PR #1213, opened 2021-07-09
[plugin] support custom scalar type aliases

dariuszkuc:
### :link: Related Issues

Discussion: https://github.com/ExpediaGroup/graphql-kotlin/discussions/1206


dariuszkuc:
While Jackson provide defaults for handling some objects like `BigInteger` we won't be able to support it generically as other objects might require default constructors or annotated methods/fields and resulting calls would fail at runtime. This is also problematic for `kotlinx.serialization` which requires all objects to be marked with `@Serializable` annotation (or be one of the supported primitives like `Int`/`String`/etc).

It might be suboptimal but current approach of using explicit converters seems like the most scalable solution.

PR #1225, opened 2021-07-22
Remove ClassScanner validation at initialization. Closes #1224

### LINKED ISSUE #1224: ClassScanner not working when building with GraalVM native-image
**Library Version**
 \>= 4.1.1

**Describe the bug**
As we know dynamic class loading is not working at runtime in native image. So ClassGraph call in ClassScanner is not working too, it not find any classes.

**To Reproduce**
Steps to reproduce the behavior. 
- compile with native-image and run, you get error: 
 "The configured packages do not contain any valid classes: "$packages" 
 Error is here https://github.com/ExpediaGroup/graphql-kotlin/blob/a6b957c999398c741cdfbc955fb86243ea211094/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt#L61

**Expected behavior**
It should working.

Possible fix is remove ClassScanner validation code, or make it optional. ClassScanner is used in a rare case for search classes by annotation.

navrocky:
### :pencil: Description
Remove validation of ClassScanner at initialization because it not working when application built with native-image (GraalVM)

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1224

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
@@ -54,15 +54,6 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) : Closeab
     internal val directives = ConcurrentHashMap<String, GraphQLDirective>()
 
     /**
-     * Validate that the supported packages contain classes
-     */
-    init {
-        if (classScanner.isEmptyScan()) {
-            throw InvalidPackagesException(config.supportedPackages)
-        }
-    }
-
-    /**
      * Generate a schema given a list of objects to parse for the queries, mutations, and subscriptions.
      */
     open fun generateSchema(


dariuszkuc:
While removing this validation may let you start your GraphQL server -> since we rely on `ClassScanner` to create polymorphic types (and federated schemas) your resulting schema will be severely limited.

navrocky:
May be move out this code to the configuration so we can choose execute automatic scan or manually provide a list of classes?

navrocky:
Also I tried this workaround (https://github.com/SoftInstigate/classgraph-on-graalvm) but with no luck.

I add these parameters to native-image cli invokation, to execute class scan at build phase:
```
tasks {
    nativeImage {
        args("--initialize-at-build-time=hello.world.ClassGraphInit,io.github.classgraph.,nonapi.io.github.classgraph.")
    }
```

ClassGraphInit.java
```
package hello.world;

import io.github.classgraph.ClassGraph;

public class ClassGraphInit {
    static {
        new ClassGraph()
                .disableModuleScanning() // added for GraalVM
                // .disableDirScanning() // added for GraalVM
                // .disableNestedJarScanning() // added for GraalVM
                // .disableRuntimeInvisibleAnnotations() // added for GraalVM
                .addClassLoader(ClassLoader.getSystemClassLoader())
                .enableAllInfo()
                .acceptPackages("hello.world")
                .scan();
    }
}
```

Then I get this error:
```
Error: type is not available in this platform: com.oracle.svm.hosted.NativeImageGenerator
Detailed message:
Trace: Object was reached by 
        indexing into array
                constant java.lang.Class[]@37711028 reached by 
        reading field nonapi.io.github.classgraph.classpath.CallStackReader.callStack
```

smyrick:
This exception was added just to catch the issues of invalid packages early on. We can remove it we will just get a more ambiguous error later on that all the types are not support because they are not in the supported packages.

If we remove this what would be the next step to support a Graal image? Do we need some way for users to pass in a custom` ClassScanner` or `ClassScanner.ScanResult`?

smyrick:
I have started a test branch here but I don't know if this would work to solve the GraalVM issues: https://github.com/ExpediaGroup/graphql-kotlin/compare/master...smyrick:move-init-to-hooks?expand=1

navrocky:
> That would mean exposing the internals of the generator and the fact we are using ClassGraph

You can wrap ClassGraph to hide internals

smyrick:
> You can wrap ClassGraph to hide internals

I don't think we want to expose all of `ClassScanner` and require you to reimplement that

dariuszkuc:
Closing stale PR as this does not solve the underlying problem that GraalVM needs to know which files are accessed by classgraph.

PR #1230, opened 2021-07-26
Fix deprecations and compile warnings

smyrick:
### :pencil: Description
With the update to Kotlin 1.5 there are some new deprecations and that we can fix and updating Gradle to use the same kotlin version under the hood

### :link: Related Issues
https://docs.gradle.org/7.1.1/userguide/upgrading_version_7.html#changes_7.1.1

### DIFF
--- a/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/GraphQLClientGenerator.kt
+++ b/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/GraphQLClientGenerator.kt
@@ -95,9 +95,9 @@ class GraphQLClientGenerator(
         }
 
         val fileSpecs = mutableListOf<FileSpec>()
-        val operationFileSpec = FileSpec.builder(packageName = config.packageName, fileName = queryFile.nameWithoutExtension.capitalizeWithDefaultLocale())
+        val operationFileSpec = FileSpec.builder(packageName = config.packageName, fileName = queryFile.nameWithoutExtension.capitalizeFirstChar())
         operationDefinitions.forEach { operationDefinition ->
-            val capitalizedOperationName = operationDefinition.name?.capitalizeWithDefaultLocale() ?: queryFile.nameWithoutExtension.capitalizeWithDefaultLocale()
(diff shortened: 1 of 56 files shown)


plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/GraphQLClientGenerator.kt
```diff
@@ -228,6 +229,13 @@ class GraphQLClientGenerator(
     }
 }
 
+/**
+ * This is the reccommended approach now with the deprecation of String.capitalize from the
+ * Kotlin stdlib in version 1.5.
+ */
+internal fun String.capitalizeWithDefaultLocale(): String =
+    replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() }
```
dariuszkuc, line 237: should this be also using `uppercase` instead of `titlecase` (for consistency)

dariuszkuc (changes requested):

dariuszkuc:
Can you also bump up the examples -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/gradle/wrapper/gradle-wrapper.properties

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/GraphQLClientGenerator.kt
smyrick: Yep, this was autogenerated from deprecation warning

website/docs/client/client-customization.mdx
```diff
@@ -3,6 +3,7 @@ id: client-customization
 title: Client Customization
 ---
 
+
```
dariuszkuc, line 6: just curious why do we need 2 blank lines on top of docs?

dariuszkuc (approved):

website/docs/client/client-customization.mdx
smyrick: 🤷🏻  That was added via the intellij "optimize imports"

Issue #1203, opened 2021-06-30
Plugin [id: 'com.expediagroup.graphql'] was not found in any of the following sources


justvil:
Hi,

I'm trying to integrate the graphql-kotlin lib into my Android app, but having trouble on the initial setup step. I used https://github.com/ExpediaGroup/graphql-kotlin/blob/master/examples/client/gradle-client/build.gradle.kts as my reference file, but after adding all the required fields I'm getting the following error:

```
org.gradle.internal.exceptions.LocationAwareException: Build file '.../mobileapp/android/app/build.gradle.kts' line: 5
Plugin [id: 'com.expediagroup.graphql'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Plugin Repositories (plugin dependency must include a version number for this source
Caused by: org.gradle.api.plugins.UnknownPluginException: Plugin [id: 'com.expediagroup.graphql'] was not found in any of the following sources:
- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Plugin Repositories (plugin dependency must include a version number for this source)
```

It's probably related to the version number, but I don't see a version number declared in your sample file. I'm probably missing something obvious here...

Thank you for your help!

dariuszkuc:
Hello 👋 
We are using composite build (example project is a composite build depending on the library project) so it automatically points to the current version. When you integrate it to your project you have to specify the plugin version.

See https://opensource.expediagroup.com/graphql-kotlin/docs/plugins/gradle-plugin-tasks

justvil:
Thank you @dariuszkuc ! Gradle is now able to locate the Plugin correctly.

By the way, since the task graphqlGenerateClient requires a task named compileKotlin to configure itself, does it mean that for Android it's only to perform the configuration explicitly? (Android projects do not contain a compileKotlin task, but compileDebugKotlin, etc.).

Thanks again!

dariuszkuc:
Indeed we look explicitly for `compileKotlin` task (https://github.com/ExpediaGroup/graphql-kotlin/blob/master/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt#L145). I'm unfamiliar with Android builds -> can you share sample project so I can take a look?

justvil:
[...]

PR #1231, opened 2021-07-27
[plugin] support android compile targets

dariuszkuc:
### :pencil: Description

Default (JVM) build use `compileKotlin`/`compileTestKotlin` tasks. Android builds use variants which has their own targets (e.g. `compileDebugKotlin`). Since our tasks were finalized by the compile tasks we couldn't apply it on android builds as default JVM tasks don't exist.

Cleaned up plugin logic for the default JVM builds by removing unnecessary `finalizeBy` clauses as adding client generating task `OutputDirectory` to the main/test sources automatically configures the necessary dependencies. Unfortunately same logic cannot be applied on Android projects as Android plugin is eagerly configured so configuring additional sources happens after this resolution. Until Android plugin is updated to support lazy configuration, we need to explicitly configure compile tasks to depend on our client generation tasks.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1203


plugins/graphql-kotlin-gradle-plugin/src/test/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePluginAndroidIT.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.jupiter.params.provider.MethodSource
+import java.io.File
+import java.nio.file.Path
+import kotlin.test.assertEquals
+
+class GraphQLGradlePluginAndroidIT {
+
+    @ParameterizedTest
+    @MethodSource("pluginTests")
+    @EnabledIfEnvironmentVariable(named = "ANDROID_SDK_ROOT", matches = ".+")
```
dariuszkuc, line 35: While Github Actions runners come preconfigured with Android setup, this ensures it will only execute if `ANDROID_SDK_ROOT` is set (so devs not touching this code don't have to setup Android env).

smyrick (approved):

PR #1234, opened 2021-08-04
Backport GraphQLUnion to 4.x.x

smyrick:
### :pencil: Description

```kotlin
```

```graphql
  foo: String!
}

  foo: String!
}

  foo: String!
  bar: String!
}

```

### :link: Related Issues
PR into master branch: https://github.com/ExpediaGroup/graphql-kotlin/pull/1194


dariuszkuc (changes requested):
Can you update 4.x docs as well?

Issue #1232, opened 2021-07-28
Generated sources not rebuilt when file changes


sproctor:
**Library Version**
4.1.1 and 5.0.0-alpha.0

**Describe the bug**
The gradle task `graphqlGenerateClient` does not generate the client if it already exists. Changing the query file and running `gradle graphqlGenerateClient` does nothing, running a clean first solves the problem.

**To Reproduce**
build.gradle.kts:
```kotlin
graphql {
    client {
        val GRAPHQL_SECRET: String by project
        packageName = "com.example.generated"
        endpoint = "https://dev.scraptickets.com/v1/graphql"
        
        allowDeprecatedFields = false
        headers = mapOf("X-Hasura-Admin-Secret" to GRAPHQL_SECRET)
        customScalars = listOf(
            com.expediagroup.graphql.plugin.gradle.config.GraphQLScalar(
                "UUID",
                "java.util.UUID",
                "com.example.UUIDScalarConverter"
            )
        )
        serializer = com.expediagroup.graphql.plugin.gradle.config.GraphQLSerializer.KOTLINX
    }
}
```
Run `gradle graphqlGenerateClient`, modify main/resources/queries.graphql and run the same command again.

**Expected behavior**
I expect the generate Kotlin files to be rebuilt.

PR #1235, opened 2021-08-07
[plugin] update generateClientTasks to correctly monitor inputs

dariuszkuc:
### :pencil: Description

By changing `queryFileDirectory` input from a String to a Directory and dropping support for providing schema file path in favor of explicit schema file, Gradle will now be able to correctly monitor task inputs to to determine whether the task is up-to-date.

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1232


dariuszkuc:
This PR fixes handling of query file directory, other outstanding issues related to inputs and UP-TO-DATE checks 
* generate client tasks can accept schema file path that can point to a local file or a classpath resource -> since it is a String value it has the same issue as query file directory fixed above (partially mitigated for classpath resources as changing dependencies should invalidate all the tasks)
* downloadSDL / introspectSchema tasks are always up to date as inputs (Strings) don't change - maybe we could do a timestamp based expiration?

smyrick (approved):

PR #1253, opened 2021-08-27
[build] dependency updates

dariuszkuc:
### :pencil: Description

Updates to the latest version of the libs.

This change does not include Kotlin update to 1.5.30 due to following issues:
* Gradle Kotlin compile task presents a warning that Java compile task (we don't have any Java sources) uses different JVM target (Java version defaults to your local JVM default)
* Above warning can be fixed by explicitly configuring Java tasks (again we don't have any Java sources) to use Java 8 target OR through toolchain configuration - at least it should be, when using toolchain configuration `kapt` blows up with unable to find the toolchain JVM
* Android test project used as an integration test for Gradle plugin blows up with `Cannot have abstract method KotlinJavaToolchain.getJdk()`


dariuszkuc:
Failure is from Gradle Android plugin -> starting with version `7.0.0` (which should match Gradle version) they no longer support Java 8

smyrick (approved):

dariuszkuc:
side note: those plugin integration tests are getting unwieldy, once IntelliJ adds support for Version Catalog we should move those IT's to their own directories under `integration` dir and use version catalog for managing dependencies.

Issue #1220, opened 2021-07-21
Cannot use ClosedRange as a property type


rocketraman:
**Library Version**
4.1.1

**Describe the bug**
I have a data class used as an input field which has a property `ClosedRange<LocalDate>`. This property cannot be easily deserialized by graphql-kotlin.

I created a coercer, which is executing properly.

However, after the coercer runs, Jackson attempts to instantiate the abstract ClosedRange type anyway, which clearly cannot work.

There is a discussion about this on Slack with @dariuszkuc : https://kotlinlang.slack.com/archives/CQLNT7B29/p1626875369050400 where I detail some of the things I've tried, including creating a custom Jackson deserializer.

**To Reproduce**

```
  override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier as? KClass<*>) {
    ClosedRange::class -> {
      when(type.arguments[0].type?.classifier as? KClass<*>) {
        LocalDate::class -> graphqlPeriodType
        else -> null
      }
    }
    else -> null
  }
```

```
typealias Period = ClosedRange<LocalDate>
```

```
val graphqlPeriodType: GraphQLScalarType = GraphQLScalarType.newScalar()
  .name("Period")
  .description("""A period of local date to local date, inclusive on both ends i.e. a closed range.""")
  .coercing(PeriodCoercing)
  .build()

object PeriodCoercing : Coercing<Period, String> {
  override fun parseValue(input: Any): Period = coerceSafe(input) {
    input.toString().parseAsPeriod()
  }

  override fun parseLiteral(input: Any?): Period? = coerceLiteralSafe(input) {
    (this as? StringValue)?.value?.parseAsPeriod()
  }

  override fun serialize(dataFetcherResult: Any): String = coerceSafe(dataFetcherResult) { toString() }

  private fun String.parseAsPeriod() = split("..").let {
    if (it.size != 2) error("Cannot parse input $this as Period")
    LocalDate.parse(it[0])..LocalDate.parse(it[1])
  }
}
```

```
class FooQuery: Query {
  fun foo(input: FooInput): String { TODO() }
}

data class FooInput(
  val foo: String,
  val range: Period,
)
```

**Expected behavior**
I wouldn't expect there to be an issues with this.

rocketraman:
This appears to be a workaround for this issue, though the value is going through several rounds of serialization and deserialization:

```
[...]

PR #1255, opened 2021-09-03
Drop Jackson, avoid duplicate deserialization

smyrick:
### :pencil: Description
Remove Jackson for converting input objects to Kotlin function arguments. graphql-java will have already run scalar coercion so the only thing we have to handle is List and Map. With Kotlin reflection we can get access to the function argument `KClass` and therefore call the constructor for that class.

This removes the need to use Jackson at all and means that we don't double deserialize objects

We are changing the public API by removing the need to have `objectMapper` as an argument. We could keep that around for backwards compatibility of others extending the logic if we wanted, but that might not be needed

### :link: Related Issues
Should help resolve https://github.com/ExpediaGroup/graphql-kotlin/issues/1220

I still need to more testing

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentToObject.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentToObject.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2021 Expedia, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.generator.execution
+
+import com.expediagroup.graphql.generator.exceptions.CouldNotConstructAValidKotlinObject
+import com.expediagroup.graphql.generator.internal.extensions.getKClass
+import com.expediagroup.graphql.generator.internal.extensions.getTypeOfFirstArgument
+import com.expediagroup.graphql.generator.internal.extensions.isOptionalInputType
+import graphql.schema.DataFetchingEnvironment
+import kotlin.reflect.KClass
+import kotlin.reflect.KParameter
+import kotlin.reflect.KType
+import kotlin.reflect.full.primaryConstructor
+
+/**
+ * Convert the input object [argumentValue] given from graphql-java to a class we can pass to the Kotlin function.
+ */
+internal fun convertArgumentToObject(
+    param: KParameter,
+    environment: DataFetchingEnvironment,
+    argumentName: String,
+    argumentValue: Any?
+): Any? = when {
+    param.type.isOptionalInputType() -> {
+        when {
+            !environment.containsArgument(argumentName) -> OptionalInput.Undefined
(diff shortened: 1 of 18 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentToObject.kt
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * The value may already be parsed, so we can perform some checks to avoid duplicate deserialization
+ */
+private fun convertValue(
+    paramType: KType,
+    argumentValue: Any?
+): Any? {
+    // The input given is already a list, iterate over each value to return a parsed list
+    if (argumentValue is Iterable<*>) {
```
dariuszkuc, line 60: instead of multiple ifs -> should this be one `when` selection instead?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentToObject.kt
```diff
@@ -0,0 +1,106 @@
...
+            val nestedTarged = it.type.getKClass()
+            @Suppress("UNCHECKED_CAST")
+            val subValue = mapToKotlinObject(input as Map<String, *>, nestedTarged)
+            constructorValues.add(subValue)
+        } else {
+            constructorValues.add(input)
+        }
+    }
+
+    return targetConstructor.call(*constructorValues.toTypedArray())
```
dariuszkuc, line 105: will this work with default values?

smyrick:
It looks like that if we want to keep array deserialization we would need to keep jackson around. It looks like DGS gets around this too by just not supporting arrays for input: https://netflix.github.io/dgs/datafetching/#using-inputargument-with-lists

Everything gets mapped to `java.util.ArrayList`

smyrick:
This would be quite the change. So I think we should push this update to a later release

dariuszkuc:
I'm somewhat leaning towards dropping the support of arrays - my gut feeling is that the fix for the deserialization logic might be more useful than supporting arrays for input.

wickedev:
I am using `com.fasterxml.jackson.databind.annotation.JsonDeserialize` to implement the schema of the GraphQL relay specification. Will this code work even after this change is applied?

```kotlin
private val encoder = Base64.getUrlEncoder().withoutPadding()

@JsonDeserialize(using = JacksonRelayNodeIdDeserializer::class)
data class ID(@get:JsonIgnore val type: String, @get:JsonIgnore val value: String) {

    companion object {
        val Empty = ID("", "")
    }

    constructor(value: String) : this("", value)

    @JsonValue
    fun toJSON(): String = encoder.encodeToString((type + ":" + this.value).toByteArray(StandardCharsets.UTF_8))

    fun toGlobalId(type: String): ID = ID(type, value)

    fun toGlobalId(type: KClass<*>): ID = ID(type.simpleName ?: "", value)

    override fun toString(): String = if (type.isEmpty()) value else "$type:$value"
}

```

rocketraman:
> This would be quite the change. So I think we should push this update to a later release

Just want to ping on this now that the mainline is on 6.0 SNAPSHOT. :-)

rocketraman:
If it helps, I've rebased this pull request here: https://github.com/rocketraman/graphql-kotlin/tree/avoid-duplicate-deserialization.

The array input tests are failing (as expected by @smyrick 's comments above), but also some `OptionalInput` tests are failing, and the reason why isn't clear to me.

rocketraman:
> It looks like this PR also breaks enums.

I've attempted a hacky fix for this on the same branch as above, commit cb783599.

dariuszkuc:
superseded by https://github.com/ExpediaGroup/graphql-kotlin/pull/1379

dariuszkuc:
@wickedev if you are using that `ID` for output type then it should work the same, if you are using that as an input then it won't work with defaults (which would no longer depend on jackson). Currently we are facing issue with custom scalar input due to the double deserialization (first by graphql-java and then by jackson). 

Jackson will still be a default serialization library used by `graphql-kotlin-spring-server`.

PR #1298, opened 2021-11-04
Custom types with annotation

smyrick:
### :pencil: Description
This adds a new annotation `@GraphQLType` which you can use to specify the return type of any field to be different than the Kotlin type. It explicitly takes a string for the type name and not just another `KClass` reference, because what we can do is return the field type as a type reference, but then allow devs to specify custom additional types that they have built up with their own GraphQLType builders.

The prime use case for this is what we were trying to accomplish with `@GraphQLUnion`. In the scenario where you want to define a new union that is many different Kotlin classes but maybe those come from other libraries and you can't modify the class to have it implement the empty interface, this can solve that solution. But now you can also specify directives, descriptions, and deprecations on the type. Plus this solution scales beyond just unions and can work for any custom type that you can define in Kotlin code. 

If we feel like this is a decent approach, I would even consider deprecating the `@GraphQLUnion` annotation in favor of this generic approach that anyone can use and even extend now that we have added the ability to pass in `GraphQLType`s in the config.


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/KTypeInfo.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.generator.internal.types
+
+/**
+ * Internal metadata class we can use to forward info about the type we are generating.
+ * If there is no metadata to add, create the class with default values.
+ */
+internal data class KTypeInfo(
```
dariuszkuc, line 23: just wondering whether this should also hold the actual type information

website/package-lock.json
```diff
@@ -2479,14 +2479,14 @@
       "dependencies": {
         "events": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/events/-/events-1.1.1.tgz",
+          "resolved": "https://artylab.expedia.biz/api/npm/public-npm-virtual/events/-/events-1.1.1.tgz",
```
dariuszkuc, line 2482: update to point to public registry

website/docs/schema-generator/customizing-schemas/custom-type-reference.md
```diff
@@ -0,0 +1,79 @@
+---
+id: custom-type-reference
+title: Custom Types
+---
+
+Code-first has many advantages and removes duplication. However, one downside is that the types defined have to match
```
dariuszkuc, line 6: can you elaborate more on the target use case? i.e. from the example below it is unclear why you would ever need to do it? I think that a line or two specifying that the `Foo`/`Bar` objects are defined in some dependent library JAR (so you cannot modify them in the app) would clarify the intent

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/KTypeInfo.kt
smyrick: We could, it would just require a lot more code changes. I figured it was more clear if we kept the `KType` argument separate from the custom class we had for more metadata about how and where the `KType` was used

dariuszkuc: maybe rename it to `KTypeMetadata` then? also is it possible to create an instance without a field name (currently it defaults to null)?

smyrick: I updated the type name to `GraphQLKTypeMetadata`. And yes it is possible to create the class without a field name in the use of additional types, ie you may never return an interface implementation explicitly in a field

lsittersexp:
Any update on this PR?

tapaderster (approved):

dariuszkuc (approved):

PR #1301, opened 2021-11-11
[server] concurrent execution for batched queries

samuelAndalon:
### :pencil: Description
This adds the logic to execute Queries in a concurrent instead of a sequential way, this will only apply for batched operations that are only queries, if there is a mutation, the batch will still be executed in a sequential way.


servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/requestExtensions.kt
```diff
@@ -34,3 +34,5 @@ fun GraphQLRequest.toExecutionInput(graphQLContext: Any? = null, dataLoaderRegis
         }
         .dataLoaderRegistry(dataLoaderRegistry ?: DataLoaderRegistry())
         .build()
+
+fun GraphQLRequest.isMutation(): Boolean = query.contains("mutation ")
```
smyrick, line 38: This would pass if someone just had a field called `mutation` as well....

Are we going to have to parse the request?

dariuszkuc: Technically it would also catch any `mutation` field suffix , e.g. `permutation`. I don't think it matters much -> we care if it is NOT a mutation so we can parallelize the calls, otherwise it will fallback to current sequential behavior

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/collectionExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package com.expediagroup.graphql.server.extensions
+
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.supervisorScope
+
+suspend fun <A, B> Iterable<A>.concurrentMap(transform: suspend (A) -> B): List<B> = supervisorScope {
+    map { async { transform(it) } }.awaitAll()
```
dariuszkuc, line 8: if any of the requests fail (they shouldn't as underlying failures should be handled), we should create generic response with error

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/requestExtensions.kt
samuelAndalon: did an investigation to see if its possible to to "cache" a `parsedDocument`, however graphql-java performs parsing and validation (syntactical and semantical executions) together, for this use case we only care about parsing to now if the query inside the document is a `mutation`,

is still possible to achieve this by implementing the  [PreparsedDocumentProvider](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/preparsed/PreparsedDocumentProvider.java), and parseAndValidate the query there, and then store the result in the ExecutionInput local context, but that would imply duplicating code from the [GraphQL API ](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/GraphQL.java)

specifically adding this part 
https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/GraphQL.java#L541


i think is not worth it, and we can just do the `contains` check, and log when a query has "mutation" as a fielf suffix .

servers/graphql-kotlin-server/build.gradle.kts
```diff
@@ -1,13 +1,44 @@
...
 }
 
+// Benchmarks
+
+// Create a separate source set for benchmarks.
+sourceSets.create("benchmarks")
+
+kotlin.sourceSets.getByName("benchmarks") {
+    dependencies {
+        implementation("org.jetbrains.kotlinx:kotlinx-benchmark-runtime:0.3.1")
```
dariuszkuc, line 23: can you add this version to properties?

servers/graphql-kotlin-server/build.gradle.kts
```diff
@@ -1,13 +1,44 @@
...
+        implementation(sourceSets.main.get().output)
+        implementation(sourceSets.main.get().runtimeClasspath)
+    }
+}
+
+benchmark {
+    targets {
+        register("benchmarks") {
+            this as JvmBenchmarkTarget
+            jmhVersion = "1.21"
```
dariuszkuc, line 33: should this be also a property?

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/collectionExtensions.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.server.extensions
+
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.supervisorScope
+
+suspend fun <A, B> Iterable<A>.concurrentMap(
```
dariuszkuc, line 23: while this is pretty generic, I wonder whether this should be exposed as part of the lib (currently it is public) -> maybe it should be internal (or private method in `GraphQLServer`)?

dariuszkuc: Iets move it inside the `GraphQLServer` - this should not be a public API

servers/graphql-kotlin-server/build.gradle.kts
```diff
@@ -1,13 +1,44 @@
...
+        }
+    }
+}
+
 tasks {
+    withType<org.gradle.jvm.tasks.Jar> {
+        duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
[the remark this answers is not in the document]
samuelAndalon: While running benchmark tests was facing an error 
```
Execution failed for task ':graphql-kotlin-server:benchmarksBenchmarkJar'.
> Entry META-INF/LICENSE is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.3.3/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.
```
reported [here](https://github.com/Kotlin/kotlinx-benchmark/issues/39),

seems that a PR with this same fix was added, will try using a newest version (currently using 0.3.1)

servers/graphql-kotlin-server/build.gradle.kts
```diff
@@ -3,9 +3,10 @@ import kotlinx.benchmark.gradle.JvmBenchmarkTarget
 description = "Common code for running a GraphQL server in any HTTP server framework"
 
 val kotlinCoroutinesVersion: String by project
+val kotlinxBenchmarkVersion: String by project
 
 plugins {
-    id("org.jetbrains.kotlinx.benchmark") version "0.3.1"
+    id("org.jetbrains.kotlinx.benchmark") version "0.4.0"
```
dariuszkuc, line 9: plugins should be managed in root settings gradle scrript

PR #1306, opened 2021-11-23
Add deprecation warning at runtime when using GraphQLContext injection

jbrwn:
### :pencil: Description

This PR updates `FunctionDataFetcher` so that it logs a warning when injecting a `GraphQLContext`.  This functionality is [deprecated](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/website/docs/schema-generator/execution/contextual-data.md#interface-injection-deprecated).  Unlike the `GraphQLContext` interface itself, which is annotated `@deprecated, there is no compile or runtime time warning when using `GraphQLContext` interface injection in a query or mutation.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/discussions/1305

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
@@ -32,6 +32,7 @@ import graphql.schema.DataFetchingEnvironment
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.future.future
+import org.slf4j.LoggerFactory
 import java.lang.reflect.InvocationTargetException
 import java.util.concurrent.CompletableFuture
 import kotlin.coroutines.CoroutineContext
@@ -58,6 +59,7 @@ open class FunctionDataFetcher(
     private val objectMapper: ObjectMapper = jacksonObjectMapper(),
     private val defaultCoroutineContext: CoroutineContext = EmptyCoroutineContext
 ) : DataFetcher<Any?> {
+    private val logger = LoggerFactory.getLogger(FunctionDataFetcher::class.java)
 
     /**
      * Invoke a suspend function or blocking function, passing in the [target] if not null or default to using the source from the environment.
@@ -106,7 +108,10 @@ open class FunctionDataFetcher(
      */
     protected open fun mapParameterToValue(param: KParameter, environment: DataFetchingEnvironment): Pair<KParameter, Any?>? =
         when {
-            param.isGraphQLContext() -> param to environment.getContext()
+            param.isGraphQLContext() -> {
+                logger.warn("GraphQLContext interface injection is deprecated. Please use DataFetchingEnvironment to retrieve ${param.getName()}.")
+                param to environment.getContext()
+            }
             param.isDataFetchingEnvironment() -> param to environment
             else -> {
                 val name = param.getName()
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
@@ -32,7 +32,6 @@ import graphql.schema.DataFetchingEnvironment
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.future.future
-import org.slf4j.LoggerFactory
 import java.lang.reflect.InvocationTargetException
 import java.util.concurrent.CompletableFuture
 import kotlin.coroutines.CoroutineContext
@@ -59,7 +58,6 @@ open class FunctionDataFetcher(
     private val objectMapper: ObjectMapper = jacksonObjectMapper(),
     private val defaultCoroutineContext: CoroutineContext = EmptyCoroutineContext
 ) : DataFetcher<Any?> {
-    private val logger = LoggerFactory.getLogger(FunctionDataFetcher::class.java)
 
     /**
      * Invoke a suspend function or blocking function, passing in the [target] if not null or default to using the source from the environment.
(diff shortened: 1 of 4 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
```diff
@@ -106,7 +108,10 @@ open class FunctionDataFetcher(
      */
     protected open fun mapParameterToValue(param: KParameter, environment: DataFetchingEnvironment): Pair<KParameter, Any?>? =
         when {
-            param.isGraphQLContext() -> param to environment.getContext()
+            param.isGraphQLContext() -> {
+                logger.warn("GraphQLContext interface injection is deprecated. Please use DataFetchingEnvironment to retrieve ${param.getName()}.")
```
dariuszkuc, line 112: Depending on how often this feature is used this could end up spamming the logs A LOT. I'm wondering whether we should log the warnings during the schema generation process instead.

jbrwn: Agree and i was actually just thinking the same thing.  I am unsure where that code lives but will do some hunting.  If you had a place in mind please let me know.

jbrwn: I refactored the PR so that the warning is added during schema generation.  Let me know what you think @dariuszkuc

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
```diff
@@ -16,12 +16,23 @@
...
-internal fun KFunction<*>.getValidArguments(): List<KParameter> =
-    this.valueParameters
-        .filterNot { it.isGraphQLContext() }
-        .filterNot { it.isGraphQLIgnored() }
-        .filterNot { it.isDataFetchingEnvironment() }
+private val logger: Logger = LoggerFactory.getLogger("schemaGenerator")
+
+internal fun KFunction<*>.getValidArguments(): List<KParameter> {
+    return this.valueParameters.mapNotNull {
+        when {
```
smyrick, line 29: This will also essentially be logged on every usage of the context class. I would like to keep this code in the generator and not in the server.

We may need to just keep a state value of if we have logged yet so we only log a message once

dariuszkuc: usage here is fine, it is invoked from `generateFunction`

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
```diff
@@ -16,12 +16,23 @@
...
+            it.isGraphQLIgnored() || it.isDataFetchingEnvironment() -> null
+            it.isGraphQLContext() -> {
+                logger.warn("GraphQLContext interface injection is deprecated. Please use DataFetchingEnvironment to retrieve ${it.getName()}.")
```
dariuszkuc, line 32: can we add some additional info to the log? i.e. `it.getName()` refers to parameter name, can we add function name and potentially the parent object name instead?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
smyrick: Your previous commit had the log happen on every execution of a request. This log will only happen at generation time, so either at start up or building of the schema. It will log for every usage still of the type but only once on each generation of the schema. 

I guess I was asking if we should only log once for the entire schema if you are using the context interface at all but this is fine as it would allow you to track the usages in logs too

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
smyrick: The function name should be available through `this` since it is an extension function

jbrwn: Great call out.  I added function name and parent name per your suggestions. Thanks.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
```diff
@@ -16,12 +16,23 @@
...
 import kotlin.reflect.full.valueParameters
 
-internal fun KFunction<*>.getValidArguments(): List<KParameter> =
-    this.valueParameters
-        .filterNot { it.isGraphQLContext() }
-        .filterNot { it.isGraphQLIgnored() }
-        .filterNot { it.isDataFetchingEnvironment() }
+private val logger: Logger = LoggerFactory.getLogger("schemaGenerator")
+
+internal fun KFunction<*>.getValidArguments(parentName: String? = null): List<KParameter> {
```
dariuszkuc, line 27: can you make this non-nullable?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
```diff
@@ -16,12 +16,23 @@
...
+    return this.valueParameters.mapNotNull {
```
dariuszkuc, line 28: can we change this to expression body syntax?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kFunctionExtensions.kt
```diff
@@ -16,12 +16,23 @@
...
+        when {
+            it.isGraphQLIgnored() || it.isDataFetchingEnvironment() -> null
+            it.isGraphQLContext() -> {
+                logger.warn("GraphQLContext interface injection is deprecated. Please use DataFetchingEnvironment to retrieve context. Parent: $parentName, function: ${getFunctionName()}, context: ${it.getName()}")
```
dariuszkuc, line 32:
```suggestion
                logger.warn("$parentName.${getFunctionName()} relies on GraphQLContext injection which is deprecated. Please update it to retrieve context from DataFetchingEnvironment instead.")
```

dariuszkuc:
Merged in https://github.com/ExpediaGroup/graphql-kotlin/pull/1323

Issue #1282, opened 2021-10-25
[client] support query directives


dariuszkuc:
### Discussed in https://github.com/ExpediaGroup/graphql-kotlin/discussions/1281

<div type='discussions-op-text'>

<sup>Originally posted by **FidasL** October 25, 2021</sup>
Hi!
I have custom graphql-server and graphql-client both developed with `graphql-kotlin` library. On the client side, I want to use schema directives on topLevelObject:
```
query {
  students @include(if: false) {
    id
    firstName
    lastName
  }
}
```

But after execute query like this from client, request serialization fails, because it generates non-nullable types in `Result` class:
```
  public data class Result(
    public val students: List<Student> // -> this type should be nullable
  )
```

Please, is this an issue, or do I need to configure something differently? Our client version is `4.1.1`. Thank you :)</div>

PR #1327, opened 2021-12-20
[generator] support toplevel Include and Skip directives

FidasL:
Generate nullable Result properties when Include or Skip directives provided in query

### :pencil: Description
Checking if query file contains toplevel`skip` or `include` directive and generating nullable properties in `Result` data class if directive provided

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1282

I am open to suggestions if some additional tests / documentation / examples are needed

### DIFF
--- a/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generatePropertySpecs.kt
+++ b/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generatePropertySpecs.kt
@@ -30,6 +30,8 @@ import com.squareup.kotlinpoet.ParameterizedTypeName
 import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeName
 import graphql.Directives.DeprecatedDirective
+import graphql.Directives.IncludeDirective
+import graphql.Directives.SkipDirective
 import graphql.language.Field
 import graphql.language.FieldDefinition
 import graphql.language.NonNullType
@@ -59,7 +61,10 @@ internal fun generatePropertySpecs(
             throw MissingArgumentException(context.operationName, objectName, selectedField.name, missingRequiredArguments)
         }
 
-        val kotlinFieldType = generateTypeName(context, fieldDefinition.type, selectedField.selectionSet)
+        val optional = selectedField.directives.any {
+            it.name == SkipDirective.name || it.name == IncludeDirective.name
+        }
+        val kotlinFieldType = generateTypeName(context, fieldDefinition.type, selectedField.selectionSet, optional = optional)
         val fieldName = selectedField.alias ?: fieldDefinition.name
 
         val propertySpecBuilder = PropertySpec.builder(fieldName, kotlinFieldType)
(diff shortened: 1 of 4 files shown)


plugins/graphql-kotlin-gradle-plugin/src/test/resources/templates/JUnit.mustache
```diff
@@ -26,6 +27,16 @@ class GraphQLMavenPluginTest {
         assertTrue(path.toFile().exists(), "graphql client was generated")
     }
 
+    @Test
+    fun `verify topLevelDirective fields are nullable`() {
+        assertTrue(
+            JUnitQuery.Result::class.declaredMemberProperties.first { it.name == "includeComplexObjectQuery" }.returnType.isMarkedNullable
```
FidasL, line 33: Maybe this can be written in some more handy way than using reflection

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generateTypeName.kt
```diff
@@ -51,8 +51,13 @@ import kotlinx.serialization.Serializable
 /**
  * Generate [TypeName] reference to a Kotlin class representation of an underlying GraphQL type.
  */
-internal fun generateTypeName(context: GraphQLClientGeneratorContext, graphQLType: Type<*>, selectionSet: SelectionSet? = null): TypeName {
-    val nullable = graphQLType !is NonNullType
+internal fun generateTypeName(
+    context: GraphQLClientGeneratorContext,
+    graphQLType: Type<*>,
+    selectionSet: SelectionSet? = null,
+    _nullable: Boolean = false
```
dariuszkuc, line 58:
```suggestion
    optional: Boolean = false
```

lets rename it to optional to avoid name clash and the _

plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generatePropertySpecs.kt
```diff
@@ -59,7 +61,10 @@ internal fun generatePropertySpecs(
             throw MissingArgumentException(context.operationName, objectName, selectedField.name, missingRequiredArguments)
         }
 
-        val kotlinFieldType = generateTypeName(context, fieldDefinition.type, selectedField.selectionSet)
+        val nullable = selectedField.directives.any {
```
dariuszkuc, line 64: lets rename it to optional (as its optional include/skip)

dariuszkuc (approved):

FidasL:
Failing pipeline is probably caused by missing default `null` value for optional fields (in kotlinx serialization). I'll have to add default constructor to creating data class `Result` which will contain default values for optional fields. 
```
@Generated
  public data class Result(
    /**
     * Query that returns wrapper object with all supported scalar types
     */
    public val scalarQuery: ScalarWrapper,
    /**
     * Query returning list of simple objects
     */
    public val listQuery: List<BasicObject>,
    /**
     * Query returning an object that references another object
     */
    public val complexObjectQuery: ComplexObject,
    /**
     * Query returning an object that references another object
     */
    public val includeComplexObjectQuery: ComplexObject2? = null, // this default null value is missing
  )
```

dariuszkuc:
Added unit and integration tests in https://github.com/ExpediaGroup/graphql-kotlin/pull/1341

PR #1336, opened 2022-01-05
EntityResolver interface

rubencagnie-toast:
### :pencil: Description
The `EntityResolver` class uses `GlobalScope.future` to dispatch the load for the entity resolving. It is easy to accidentally create resource or memory leaks when GlobalScope is used.

This PR abstracts out the actual entity resolving logic as an interface. The logic has moved to a `resolve` method, that can be called from any class that implements this interface.

The EntityResolver can be injected into the `FederatedSchemaGeneratorHooks` class, in case this needs to have a different way of being called. There is a default implementation (`GlobalScopeEntityResolver`), which keeps the current behavior in tact. A convenience `constructor` is added to maintain the current initialization with an array of `FederatedTypeResolver` and is using that `GlobalScopeEntityResolver` to maintain backwards compatibility

### DIFF
--- a/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/FederatedSchemaGeneratorHooks.kt
+++ b/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/FederatedSchemaGeneratorHooks.kt
@@ -28,6 +28,7 @@ import com.expediagroup.graphql.generator.federation.directives.PROVIDES_DIRECTI
 import com.expediagroup.graphql.generator.federation.directives.REQUIRES_DIRECTIVE_TYPE
 import com.expediagroup.graphql.generator.federation.execution.EntityResolver
 import com.expediagroup.graphql.generator.federation.execution.FederatedTypeResolver
+import com.expediagroup.graphql.generator.federation.execution.GlobalScopeEntityResolver
 import com.expediagroup.graphql.generator.federation.extensions.addDirectivesIfNotPresent
 import com.expediagroup.graphql.generator.federation.types.ANY_SCALAR_TYPE
 import com.expediagroup.graphql.generator.federation.types.ENTITY_UNION_NAME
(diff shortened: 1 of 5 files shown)


dariuszkuc:
Hello 👋 
Thanks for opening the PR. It's interesting idea but depending on the integration with `spring-graphql` we might be able to simplify this logic and rely on Spring to propagate all contextual data. AFAIK `spring-graphql` is targeting Spring '22 for the GA.

rubencagnie:
Thanks for the reply @dariuszkuc. We are not using spring, so that suggestion wouldn't apply to us :(

As this would behave the exact same way as before when you don't specify a EntityResolver, would it be possible to get this in? This would drastically simply our workflow as today we use the hooks to achieve the same, but we had to copy over a part of the code, which is obviously not ideal...

dariuszkuc:
👋 I think might be a good idea abstract this logic away and while this part does not directly depend on Spring integration, it is still a major change that might be impacted with the upcoming changes due to potential `graphql-spring` integration. I'll try to block some time this week to look into how those integrations might look like to better understand the execution model there.

rubencagnie-toast:
That would be great, @dariuszkuc 

Overall, I think it would be great if this framework has a base interface of defining the coroutine contexts that the resolvers are executed in. Then you can add the specific implementation depending on what framework you run in.

Thanks for looking into this and let me know if I can be of any help

rubencagnie-toast:
@dariuszkuc - friendly poke to see if there is any status change on this...

dariuszkuc:
👋 I'm updating the way that we handle context to rely on the new `GraphQLContextMap` to propagate the scope information. Unfortunately it is a breaking change so will start cutting new alpha version as well.

High level logic
1. users can provide custom context map through a factory
2. this map can contain an entry for custom coroutine context (e.g. setting up MDC, thread locals etc)
3. graphql server (before executing the query) creates supervisor scope using the current context + custom one from the user and stores it in the context map
4. data fetchers/resolvers will then attempt to retrieve the scope from the context (if not available fallback to global scope) and use it to execute suspending functions

This is similar to what `spring-graphql` project does and should provide easy interop in the future.

dariuszkuc:
Superseded by https://github.com/ExpediaGroup/graphql-kotlin/pull/1349

PR #1338, opened 2022-01-07
Updated schema generator to support return DataFetcherResult of list of DataFetcherResult

### LINKED ISSUE #1337: Support for DataFetcherResult<List<DataFetcherResult<Entity>>>
We have some code similar to this

```
class EntityQuery(private val repository: Repository) : Query {
  fun entities() : DataFetcherResult<List<DataFetcherResult<Entity>>> {
    val result = repository.getEntities()
    return when (result) {
      is Success -> {
        val entityDataFetcherResults = result.value.map { entity ->
          DataFetcherResult.newResult<Entity>()
            .data(entity)
            .localContext(entity.id) // actually more involved then this in the real case
            .build()
        }
        DataFetcherResult.newResult<List<DataFetcherResult<Entity>>>()
           .data(entityDataFetcherResults)
           .build()
      }
      is Failure -> {
        DataFetcherResult.newResult<List<DataFetcherResult<Entity>>()
          .error(result.reason.toGraphQLError())
          .build()
      }
    }
  }
}
```

The first DataFetcherResult in the type signature is used to map a Result type from our domain into success or error cases, without resorting to throwing. The second DataFetcherResult is used to provide localContext to each individual entity.

When we try to run this example we are getting an error similar to `Cannot convert graphql.execution.DataFetcherResult<##REDACTED##> since it is not a valid GraphQL type or outside the supported packages "[##REDACTED##]"`

If we change the signature to return `DataFetchResult<List<Entity>>` or `List<DataFetcherResult<Entity>>` these both produce schemas.

At the moment we are working around this issue by changing the signature to return `List<DataFetcherResult<Entity>>` and throwing in the result of failure; but this is not ideal. It is even more annoying at the lower levels of the hierarchy as it prevents us from returning partial results on partial failure.

dryan-cais:
Fixes #1337

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/utils/functionReturnTypes.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/utils/functionReturnTypes.kt
@@ -42,9 +42,11 @@ import kotlin.reflect.full.createType
  *      CompletableFuture<T>
  *      CompletableFuture<DataFetcherResult<T>>
  *      List<DataFetcherResult<T>>
+ *      DataFetcherResult<List<DataFetcherResult<T>>>
  */
 internal fun getWrappedReturnType(returnType: KType): KType {
     return when {
+        returnType.isSubclassOf(DataFetcherResult::class) && returnType.getTypeOfFirstArgument().isSubclassOf(List::class) -> getWrappedReturnType(returnType.getTypeOfFirstArgument())
         returnType.isSubclassOf(DataFetcherResult::class) -> returnType.getTypeOfFirstArgument()
         returnType.isSubclassOf(Publisher::class) -> { checkTypeForDataFetcherResult(returnType) }
         returnType.isSubclassOf(CompletableFuture::class) -> { checkTypeForDataFetcherResult(returnType) }
--- a/generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/internal/types/utils/FunctionReturnTypesKtTest.kt
+++ b/generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/internal/types/utils/FunctionReturnTypesKtTest.kt
@@ -32,6 +32,8 @@ class FunctionReturnTypesKtTest {
         val listOfListString: List<List<String>> = listOf(listOf(string))
         val dataFetcherResult: DataFetcherResult<String> = DataFetcherResult.newResult<String>().data(string).build()
         val listDataFetcherResult: List<DataFetcherResult<String>> = listOf(DataFetcherResult.newResult<String>().data(string).build())
+        val dataFetcherResultListDataFetcherResult: DataFetcherResult<List<DataFetcherResult<String>>> =
+            DataFetcherResult.newResult<List<DataFetcherResult<String>>>().data(listOf(DataFetcherResult.newResult<String>().data(string).build())).build()
         val dataFetcherResultListString: DataFetcherResult<List<String>> = DataFetcherResult.newResult<List<String>>().data(listOf(string)).build()
         val listDataFetcherResultListString: List<DataFetcherResult<List<String>>> = listOf(DataFetcherResult.newResult<List<String>>().data(listOf(string)).build())
         val publisher: Publisher<String> = Flowable.just(string)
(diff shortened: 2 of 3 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/utils/functionReturnTypes.kt
```diff
@@ -45,7 +45,7 @@ import kotlin.reflect.full.createType
  */
 internal fun getWrappedReturnType(returnType: KType): KType {
     return when {
-        returnType.isSubclassOf(DataFetcherResult::class) -> returnType.getTypeOfFirstArgument()
+        returnType.isSubclassOf(DataFetcherResult::class) -> getWrappedReturnType(returnType.getTypeOfFirstArgument())
```
dariuszkuc, line 48: I believe we probably would need to make a distinction between (and add handling for each)

* `DataFetcherResult<T>` - current logic
* `DataFetcherResult<List<T>>` - currently unsupported
* `DataFetcher<List<DataFetcher<T>>>` - currently unsupported

Haven't tested it but following two cases might cover it

```kotlin
  // if return type is DataFetcher<List<T>> check if T has to be unwrapped
  returnType.isSubclassOf(DataFetcherResult::class) && returnType.getTypeOfFirstArgument().isSubclassOf(List::class) -> {
    List::class.createType(listOf(KTypeProjection(KVariance.INVARIANT, getWrappedReturnType(returnType.getTypeOfFirstArgument().getTypeOfFirstArgument()))))
  }
  // keep current case as is
  returnType.isSubclassOf(DataFetcherResult::class) -> returnType.getTypeOfFirstArgument() is
```

dariuszkuc: Actually, I think just doing the `getWrappedReturnType()` as you suggested might do the trick (list types would be handled recursively). Can you check the test failures?

```
GenerateFunctionTest > DataFetcherResult of a CompletableFuture is invalid() FAILED
    org.opentest4j.AssertionFailedError at GenerateFunctionTest.kt:276

FunctionReturnTypesKtTest > getWrappedReturnType of DataFetcherResult() FAILED
    org.opentest4j.AssertionFailedError at FunctionReturnTypesKtTest.kt:66
```

dryan-cais: Strange. I thought all the tests were passing when I ran locally. There quite a lot of noise with warnings etc, so I might have failed to spot them.

Sorry about this, I will check this out straight away.

dryan-cais: Provided a fix for the tests. Can you please rerun the workflow?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/utils/functionReturnTypes.kt
```diff
@@ -45,6 +45,7 @@ import kotlin.reflect.full.createType
  */
 internal fun getWrappedReturnType(returnType: KType): KType {
     return when {
```
smyrick, line 47: Can we have a better recursive call here? Would this work?

```kotlin
internal fun getWrappedReturnType(returnType: KType): KType
return when {

   returnType.isSubclassOf(DataFetcherResult::class) -> getWrappedReturnType(returnType.getTypeOfFirstArgument())
   returnType.isSubclassOf(Publisher::class) -> { checkTypeForDataFetcherResult(returnType) }
   returnType.isSubclassOf(CompletableFuture::class) -> { checkTypeForDataFetcherResult(returnType) }
   returnType.isSubclassOf(List::class) -> {
            List::class.createType(listOf(KTypeProjection(KVariance.INVARIANT, getWrappedReturnType(returnType.getTypeOfFirstArgument()))))
        }
        else -> returnType
}
```

dryan-cais: I don't think this is possible without changing the existing behaviour. For example, currently for an invalid case, CompletableFuture<Publisher<String>> returns Publisher<String>; whereas the fully-recursive implementation would return String.

To be honest I am not sure why the function behaves as it does now. But because this is my first PR to the library, and because it fixes a pressing issue for us, I wanted to keep the scope of the change small, to just fix our use-case.

dariuszkuc (approved):

PR #1348, opened 2022-01-24
feat: transaction batcher module

samuelAndalon:
### :pencil: Description
`transaction-batcher` module is an alternative to `data-loaders`, using reactive stream to apply batching and deduplication of transactions needed by data fetchers, the implementation is completely agnostic of GraphQL and compatible with any asynchronous approach used to resolve data.
it also supports caching to avoid making a request to a previous completed transaction.


dariuszkuc:
@samuelAndalon can you add copyright to all new source files?

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLAutoConfiguration.kt
```diff
@@ -30,4 +30,5 @@ import org.springframework.context.annotation.Import
     PlaygroundRouteConfiguration::class,
     SdlRouteConfiguration::class
 )
-class GraphQLAutoConfiguration
+class
+GraphQLAutoConfiguration
```
dariuszkuc, line 34: i believe this change was unintentional?

samuelAndalon: yes, sorry. Just removed it

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
```diff
@@ -0,0 +1,100 @@
...
+ */
+@Suppress(
+    "ReactiveStreamsSubscriberImplementation",
+    "UNCHECKED_CAST"
+)
+interface TriggeredPublisher<TInput, TOutput> {
+    /**
+     * Given a input of type [TInput] create a cold [Publisher] that will produce a [Publisher] of type [TOutput] of n elements
+     * that maps to the size of the input [List] of [TInput]
+     * order is important so make sure ot produce elements in the same order of [input]
```
tapaderster, line 36:
```suggestion
     * Given an input of type [TInput] create a cold [Publisher] that will produce a [Publisher] of type [TOutput] of n elements
     * that maps to the size of the input [List] of [TInput]
     * order is important so make sure to produce elements in the same order of [input]
```

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
```diff
@@ -0,0 +1,100 @@
...
+     */
+    fun produce(input: List<TInput>): Publisher<TOutput>
+
+    /**
+     * Attempt to collect values from first [cache] and then [produce]
```
tapaderster, line 41:
```suggestion
     * Attempts to collect values from [cache] first and then [produce]
```

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/transaction/TransactionBatcher.kt
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Holds logic to apply batching, deduplication and caching of [BatcheableTransaction]
+ * if no [TransactionBatcherCache] implementation is provided it will use [DefaultTransactionBatcherCache]
+ */
+class TransactionBatcher(
+    private val cache: TransactionBatcherCache = DefaultTransactionBatcherCache()
+) {
+
+    private val queue = ConcurrentHashMap<
```
tapaderster, line 40: does order matter here?

executions/graphql-kotlin-transaction-batcher/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+val junitVersion: String by project
+val slf4jVersion: String by project
+val mockkVersion: String by project
+val reactorVersion: String by project
+val reactorExtensionsVersion: String by project
+
+dependencies {
+    implementation("org.reactivestreams:reactive-streams:$reactiveStreamsVersion")
+    implementation("org.slf4j:slf4j-api:$slf4jVersion")
+    testImplementation("org.junit.jupiter:junit-jupiter-api:$junitVersion")
```
dariuszkuc, line 13: just a nit but can you make the deps within `testImplementation` sorted in alphanumerical order?

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/transaction/TransactionBatcher.kt
dariuszkuc: queue implies FIFO and it doesn't look like this variable is used as a queue, maybe we need a better name?

executions/graphql-kotlin-transaction-batcher/src/test/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/AstronautService.kt
```diff
@@ -0,0 +1,66 @@
...
+    val produceArguments: MutableList<List<AstronautServiceRequest>> = mutableListOf()
+    val getAstronautCallCount: AtomicInteger = AtomicInteger(0)
+    private val astronautsPublisher = object : TriggeredPublisher<AstronautServiceRequest, Astronaut> {
+        override fun produce(input: List<AstronautServiceRequest>): Publisher<Astronaut> {
+            produceArguments.add(input)
+            return this@AstronautService.getAstronauts(input)
+        }
+    }
+
+    companion object {
```
dariuszkuc, line 43: nit -> per Kotlin code style companion objects should be declared last (https://kotlinlang.org/docs/coding-conventions.html#class-layout)

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
```diff
@@ -0,0 +1,100 @@
...
+                    this.subscription.request(1)
+                }
+
+                override fun onNext(result: TOutput) {
+                    results += result
+                    this.subscription.request(1)
+                }
+
+                override fun onError(throwable: Throwable) {
+                    throwable.printStackTrace()
```
dariuszkuc, line 82: we probably should use slf4j to log it

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/transaction/TransactionBatcher.kt
samuelAndalon: > does order matter here?

Order in the queue (will rename it to batch) does not matter, what it matters is each publisher needs to produce data in the same order of inputs

samuelAndalon: > queue implies FIFO and it doesn't look like this variable is used as a queue, maybe we need a better name?

i agree, probably just `batch` is a better name.

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.reactivestreams.Publisher
+import org.reactivestreams.Subscriber
+import org.reactivestreams.Subscription
+import org.slf4j.LoggerFactory
+
+/**
+ * Interface representing a publisher with input [TInput] type and output [TOutput] type
+ */
+@Suppress(
+    "ReactiveStreamsSubscriberImplementation",
```
dariuszkuc, line 30: why are we suppressing this warning? is this valid?

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
```diff
@@ -0,0 +1,105 @@
...
+     * if [TriggeredPublisher] is of type <Int, Int> and [cache] resolves [1, null, 3, null, 5, 6]
+     * we will attempt to produce elements for index 1 and 3
+     * when [produce] stream completes we will complete futures from either values resolved from [cache] or from [produce]
+     */
+    fun trigger(
+        batcheableTransactions: List<BatcheableTransaction<TInput, TOutput>>,
+        cache: TransactionBatcherCache
+    ) {
+
+        val values = batcheableTransactions.map { batcheableTransaction ->
```
dariuszkuc, line 54: I think we can simplify this to something like

```kotlin
val transactionsNotInCache = batchableTransactions.filterNot { transaction -> cache.get(transaction.key) != null }
```

*instead of doing the `cache.get` we could also add `cache.contains` which I think would clarify the intent better

dariuszkuc: actually I'd also apply `map(BatcheableTransaction<TInput, TOutput>::input)` at the same time so we would just pass a list

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
samuelAndalon: it is valid, given that `Subscriber` is an interface, so it's valid to implement it, However a warning appears in the IDE when trying to do it.

i can see a lot of `Subscriber` implementations that need to add that suppress as well
https://github.com/search?p=2&q=%22ReactiveStreamsSubscriberImplementation%22&type=Code

dariuszkuc: unsure, I am not getting any warning in my local

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
```diff
@@ -0,0 +1,99 @@
...
+                    transactionResults += result
+                    this.subscription.request(1)
+                }
+
+                override fun onError(throwable: Throwable) {
+                    logger.error("Error while producing data", throwable)
+                }
+
+                override fun onComplete() {
+                    val cachedTransactionResults = batchableTransactions.map { transaction -> cache.get(transaction.key) }
```
dariuszkuc, line 79: I think we can simplify this even further

```kotlin
val transactionResultsIterator = transactionResults.iterator()
batchableTransactions.forEach { transaction ->
    var result = cache.get(transaction.key) as? TOutput
    if (result == null) {
        result = transactionResultsIterator.next()
        cache.set(transaction.key, result as Any)
    }
    transaction.future.complete(result)
}
```

technicallly that if block could be simplified even further to something like `val result = cache.computeIfAbsent(transaction.key) { transactionResultsIterator.next() }` but unsure if it is worth it

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/transaction/DefaultTransactionBatcherCache.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package com.expediagroup.graphql.transactionbatcher.transaction
+
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Default implementation of [TransactionBatcherCache] using an in memory [cache]
+ * without eviction
+ */
+class DefaultTransactionBatcherCache : TransactionBatcherCache {
```
dariuszkuc, line 25: can we move `DefaultTransactionBatcherCache` and  `TransactionBatcherCache` to `com.expediagroup.graphql.transactionbatcher.transaction.cache` package?

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/publisher/TriggeredPublisher.kt
samuelAndalon: image

executions/graphql-kotlin-transaction-batcher/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/transaction/TransactionBatcher.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Trigger concurrently and asynchronously the instances of [TriggeredPublisher] that the [batch] holds
+     * at the end clear the queue
+     */
+    @Synchronized fun dispatch() {
+        batch.values.forEach { (triggeredPublisher, transactions) ->
+            triggeredPublisher.trigger(
+                transactions.map(Map.Entry<String, BatchableTransaction<Any, Any>>::value),
```
dariuszkuc, line 91: why not just `transactions.values` (collection) or `transaction.values.toList()`?

dariuszkuc (approved):

rocketraman:
Why does this use reactive-streams instead of kotlinx-coroutines Flow? Flow can be transformed to reactive-streams by users, if desired.

dariuszkuc:
We decided to use generic reactive streams `Publisher` for flexibility to allow users to use various implementations (including coroutines).

samuelAndalon:
reactive streams and all his implementations like project-reactor has 100% interoperabilty with other async technologies like Java Flow, coroutines, completableFuture etc.
So the most generic choice was RX, at the beggining we were scoping for java flow, but that requires target jvm to be java 9.
=

rocketraman:
Ok. Given this is a Kotlin project, I would have thought coroutines Flow would be the obvious choice as first-party implementation. As stated, it is easy to convert from Flow to other reactive streams implementations i.e. `flow.asPublisher()`.

Is there any documentation in the works for this feature? Seems really interesting.

samuelAndalon:
Once we add the instrumentation that will signal when to dispatch, we will add the documentation.
It will be pretty much similar to the `DataLoaderDispatcherInstrumentation` from graphql-java but applied to more than 1 operation is client sent a request with batched operations.

for now you can refer to the [unit tests](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/executions/graphql-kotlin-transaction-batcher/src/test/kotlin/com/expediagroup/graphql/transactionbatcher/transaction/TransactionBatcherTest.kt)

Issue #1318, opened 2021-12-02
context map rough edges


halfbaked:
**Library Version**
5.2.0

**Describe the bug**
I wonder if the `SpringGraphQLContextFactory` needs some finesse. It currently requires us to implement the deprecated `suspend fun generateContext(request: ServerRequest)` while leaving the new `suspend fun generateContextMap(request: ServerRequest): Map<*, Any>?` optional. In the current situation, using the IDE to implement the necessary methods in the subclass directs users to continue using the deprecated method. 

Also in the documentation for Generating GraphQL Context in spring server, the examples are for the deprecated method. To me, as a newcommer, it's a little confusing.

**Expected behavior**
Perhaps (for backwards compatibility) give both `generateContext` and `generateContextMap` default implementations in the abstract class, and update the spring server docs to use `generateContextMap`. 

Keep up the good work!

dariuszkuc:
Should be resolved by https://github.com/ExpediaGroup/graphql-kotlin/pull/1349

Issue #1257, opened 2021-09-15
Coroutine cancellation support


DrMetallius:
**Is your feature request related to a problem? Please describe.**
Currently `FunctionDataFetcher` launches suspend functions using `GlobalScope`. This means that it is disconnected from the scope the request is in and can't be canceled.

We discovered this when we were load testing the server. We can set a timeout for requests to drop the ones the server didn't handle in time, but this won't cancel the running coroutines started with `GlobalScope`. This means that when the load is above a certain threshold, it overwhelms the server, the memory is filled with uncancelled coroutine contexts from all the coroutine starts, and the server is doing the work for the clients which are no longer there.

**Describe the solution you'd like**
`GraphQLServer` should create a `supervisorScope`, run all the methods within it, and store a reference to that scope somewhere accessible to data fetchers. Judging by what GraphQL Java makes available, this should probably just be in `GraphQLContext`. `FunctionDataFetcher` can then get the scope and use it to execute the functions instead of `GlobalScope`.

**Describe alternatives you've considered**
Right now we're doing the same thing manually by overriding what GraphQL Kotlin offers us. However, we would prefer to have this problem fixed in the framework since it makes it more difficult to upgrade to newer versions as we have to reimplement the solution again every time.

I've seen issue #879, which describes an attempt to resolve this by making the whole context a scope, but that seems inflexible, also that commit leaves out the code required to actually create the scope. Speaking of the parent-child relationship of coroutines mentioned there, I don't see why it would be a blocker. `GlobalScope` doesn't handle this anyway, so it's not a regression. We could start with the basic support at the request level and make it more granular later if we need to, possibly using local context.

I really hope this issue gets addressed, it caused us a lot of trouble when we load tested our server.

dariuszkuc:
This should be resolved in https://github.com/ExpediaGroup/graphql-kotlin/pull/1349

PR #1349, opened 2022-01-25
standardize coroutine context propagation in the execution

dariuszkuc:
### :pencil: Description

Attempts to standardize how we can propagate the coroutine/reactor context created by the server when processing incoming requests to the data fetchers that would resolve the underlying fields.

`GraphQLContext` map was introduced in `graphql-java` 17 as means to standardize usage of GraphQL context (previously it could be any object). We can store the original GraphQL request context in the map and then use it from within the data fetchers.

This is a breaking change as we are changing signatures of `GraphQLContextFactory` and `FunctionDataFetcher`.

### :link: Related Issues

* https://github.com/ExpediaGroup/graphql-kotlin/pull/1336
* https://github.com/ExpediaGroup/graphql-kotlin/issues/1318
* https://github.com/ExpediaGroup/graphql-kotlin/discussions/1300
* https://github.com/ExpediaGroup/graphql-kotlin/issues/1257


servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLContextFactory.kt
```diff
@@ -16,23 +16,16 @@
 
 package com.expediagroup.graphql.server.execution
 
-import com.expediagroup.graphql.generator.execution.GraphQLContext
+const val GRAPHQL_COROUTINE_CONTEXT_KEY = "GRAPHQL_COROUTINE_CONTEXT"
```
smyrick, line 19: Is this supposed to be `GRAPHQL_COROUTINE_SCOPE`? Maybe we should just use the value defined in the generator package so we don't have it defined in two locations

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/subscriptions/ApolloSubscriptionSessionState.kt
```diff
@@ -31,43 +30,24 @@ internal class ApolloSubscriptionSessionState {
...
-        }
-    }
-
-    /**
-     * Save the context created from the factory and possibly updated in the onConnect hook.
-     * This allows us to include some intial state to be used when handling all the messages.
-     * This will be removed in [terminateSession].
-     */
-    fun saveContextMap(session: WebSocketSession, graphQLContext: Map<*, Any>?) {
+    fun saveContextMap(session: WebSocketSession, graphQLContext: Map<*, Any?>?) {
```
smyrick, line 41: If we are making the breaking change do we want to call it `saveGraphQLContext` to match conventions?

dariuszkuc: Yeah was just trying to get it compile first.

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLContextFactory.kt
dariuszkuc: Actually I'm thinking of dropping the String key in favor of just using classes as keys.

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLServer.kt
```diff
@@ -39,18 +45,21 @@ open class GraphQLServer<Request>(
...
         val graphQLRequest = requestParser.parseRequest(request)
 
-        return if (graphQLRequest != null) {
-            val context = contextFactory.generateContext(request)
-            val graphQLContext = contextFactory.generateContextMap(request)
+        if (graphQLRequest != null) {
+            val contextMap = contextFactory.generateContextMap(request) ?: emptyMap<Any, Any?>()
+
+            val customContext: CoroutineContext = contextMap[GRAPHQL_COROUTINE_CONTEXT_KEY] as? CoroutineContext ?: EmptyCoroutineContext
+            val graphQLExecutionScope = CoroutineScope(coroutineContext + customContext + SupervisorJob())
```
jgorman-exp, line 55: Not an expert on coroutines, but does this need to be cancelled at the end of request processing?  Was just reading through this:  https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html

dariuszkuc: When we return the value from this method then all children coroutines will already complete so there is nothing to cancel.

TiKevin83:
Ran into #1300 testing this library in a Kotlin project and was able to confirm that this PR resolves the issues we were seeing with the spring security context being empty when coming from the GraphQL endpoints!

wallind:
> Ran into #1300 testing this library in a Kotlin project and was able to confirm that this PR resolves the issues we were seeing with the spring security context being empty when coming from the GraphQL endpoints!

to be even more specific we were able to get things working with manually built jars specifically from this commit https://github.com/ExpediaGroup/graphql-kotlin/pull/1349/commits/87fc89ce7b9abaee2025b273c14a6f80af90c692

dariuszkuc:
Updated to keep the deprecated custom GraphQL context around (will drop it in v7) to provide easier migration path.

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/execution/SpringGraphQLContextFactory.kt
```diff
@@ -25,8 +26,13 @@ import org.springframework.web.reactive.function.server.ServerRequest
...
 /**
- * Basic implementation of [SpringGraphQLContextFactory] that returns [SpringGraphQLContext]
+ * Basic implementation of [SpringGraphQLContextFactory] that populates Apollo tracing header.
  */
-class DefaultSpringGraphQLContextFactory : SpringGraphQLContextFactory<SpringGraphQLContext>() {
-    override suspend fun generateContext(request: ServerRequest) = SpringGraphQLContext(request)
+open class DefaultSpringGraphQLContextFactory : SpringGraphQLContextFactory<SpringGraphQLContext>() {
+    override suspend fun generateContextMap(request: ServerRequest): Map<*, Any> = mutableMapOf<Any, Any>()
+        .also { map ->
+            request.headers().firstHeader(FEDERATED_TRACING_HEADER_NAME)?.let { headerValue ->
```
tapaderster, line 34: nit: a comment on why we are adding this header will be great.

tapaderster (approved):

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/execution/SpringGraphQLContextFactory.kt
dariuszkuc: its specified in the KDoc on the class -> `Basic implementation of [SpringGraphQLContextFactory] that populates Apollo tracing header.`

samuelAndalon (approved):

frozenice:
I'm interested in trying this out. I don't want to build and publish it somewhere for myself and our CI.

@dariuszkuc Do you publish snapshots of the `graphql-kotlin` packages somewhere? Or even better: any idea when this will be released?

dariuszkuc:
@frozenice we just released `6.0.0-alpha.0`

PR #1359, opened 2022-02-12
[generator] mark default PropertyDataFetcher as "trivial"

dariuszkuc:
### :pencil: Description
Various instrumentations (e.g. tracing) rely on `TrivialDataFetcher` distinction to skip their execution. Looks like we missed this when creating this data fetcher.

### :link: Related Issues
N/A

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/PropertyDataFetcher.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/PropertyDataFetcher.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Expedia, Inc
+ * Copyright 2022 Expedia, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/PropertyDataFetcher.kt
```diff
@@ -25,7 +26,7 @@ import kotlin.reflect.KProperty
  *
  * @param propertyGetter Kotlin property getter that will be invoked to resolve a field
  */
-class PropertyDataFetcher(private val propertyGetter: KProperty.Getter<*>) : DataFetcher<Any?> {
+class PropertyDataFetcher(private val propertyGetter: KProperty.Getter<*>) : DataFetcher<Any?>, TrivialDataFetcher<Any?> {
```
samuelAndalon, line 29: TrivialDataFetcher<T> already implements DataFetcher<T>, should we just 
`class PropertyDataFetcher(private val propertyGetter: KProperty.Getter<*>) : TrivialDataFetcher<Any?>` ?

dariuszkuc: Good callout. 👍

PR #1362, opened 2022-02-15
Add small paragraph when dealing with input classes

tobias-:
### :pencil: Description

Unfortunately, when deserializing attributes, `ctxt.parser.parsingContext.currentName` will be
the name of the enclosing class when the attribute the isn't defined/sent.

### :link: Related Issues

No related issue.


dariuszkuc:
Hi 👋
I could not repro the default `null` value and was always getting undefined even without default value. In https://github.com/ExpediaGroup/graphql-kotlin/pull/1367 I've updated optional docs with the link to the argument default values.

PR #1378, opened 2022-02-24
feat: transaction batcher instrumentation level

samuelAndalon:
### :pencil: Description
custom Instrumentation that will define when is the right moment to dispatch operations added in `TransactionBatcher`.

This instrumentation follows the same approach of the [DataLoaderDispatcherInstrumentation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/dataloader/DataLoaderDispatcherInstrumentation.java)

The main difference is that Instrumentation is applied by `Execution` aka GraphQL Operation, and this new Instrumentation accesses to an state in the GraphQL context to keep track of each execution state and the apply the same logic that the `DataLoaderDispatcherInstrumentation` uses, once a certain level of all executions was dispatched we signal the `TransactionBatcher` to dispatch.

image


settings.gradle.kts
```diff
@@ -52,6 +52,7 @@ include(":graphql-kotlin-spring-server")
 
 // Executions
 include(":graphql-kotlin-transaction-batcher")
+include(":graphql-kotlin-transaction-batcher-instrumentation-level")
```
dariuszkuc, line 55: this is a very long artifact name - we can drop the `-level` but wondering whether something like `graphql-kotlin-batch-instrumentation` is sufficient (probably it would make sense to keep naming consistent so maybe `transaction-batcher-instrumentation` or do we rename the `transaction-batcher` to something shorter?

dariuszkuc (changes requested):
Missing KDocs and new module README

executions/graphql-kotlin-transaction-batcher-instrumentation/README.md
```diff
@@ -0,0 +1,83 @@
+# GraphQL Kotlin Transaction Batcher Instrumentation
+[![Maven Central](https://img.shields.io/maven-central/v/com.expediagroup/graphql-kotlin-transaction-batcher.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22com.expediagroup%22%20AND%20a:%22graphql-kotlin-transaction-batcher%22)
+[![Javadocs](https://img.shields.io/maven-central/v/com.expediagroup/graphql-kotlin-transaction-batcher.svg?label=javadoc&colorB=brightgreen)](https://www.javadoc.io/doc/com.expediagroup/graphql-kotlin-transaction-batcher)
+
+`graphql-kotlin-transaction-batcher-instrumentation` is a custom instrumentation that will signal when is the right moment
+to dispatch transactions added in the `TransactionBatcher` located in the `GraphQLContext`
+
+This instrumentation follows the same approach of the [DataLoaderDispatcherInstrumentation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/dataloader/DataLoaderDispatcherInstrumentation.java)
```
dariuszkuc, line 8:
```suggestion
This instrumentation follows the same approach of the [DataLoaderDispatcherInstrumentation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/dataloader/DataLoaderDispatcherInstrumentation.java).
```

executions/graphql-kotlin-transaction-batcher-instrumentation/README.md
```diff
@@ -0,0 +1,83 @@
+# GraphQL Kotlin Transaction Batcher Instrumentation
+[![Maven Central](https://img.shields.io/maven-central/v/com.expediagroup/graphql-kotlin-transaction-batcher.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22com.expediagroup%22%20AND%20a:%22graphql-kotlin-transaction-batcher%22)
```
dariuszkuc, line 2: I believe those links should point to `graphql-kotlin-transaction-batcher-instrumentation` instead?

samuelAndalon: my bad, sorry.

executions/graphql-kotlin-transaction-batcher-instrumentation/README.md
```diff
@@ -0,0 +1,83 @@
+# GraphQL Kotlin Transaction Batcher Instrumentation
+[![Maven Central](https://img.shields.io/maven-central/v/com.expediagroup/graphql-kotlin-transaction-batcher-instrumentation.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22com.expediagroup%22%20AND%20a:%22graphql-kotlin-transaction-batcher-instrumentation%22)
+[![Javadocs](https://img.shields.io/maven-central/v/com.expediagroup/graphql-kotlin-transaction-batcher-instrumentation.svg?label=javadoc&colorB=brightgreen)](https://www.javadoc.io/doc/com.expediagroup/graphql-kotlin-transaction-batcher-instrumentation)
+
+`graphql-kotlin-transaction-batcher-instrumentation` is a custom instrumentation that will signal when is the right moment
+to dispatch transactions added in the `TransactionBatcher` located in the `GraphQLContext`.
+
+This instrumentation follows the same approach of the [DataLoaderDispatcherInstrumentation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/dataloader/DataLoaderDispatcherInstrumentation.java).
+
+The main difference is that `Instrumentation` is applied by `ExecutionInput` aka GraphQL Operation, and this new Instrumentation
```
dariuszkuc, line 10:
```suggestion
The main difference is that regular `Instrumentation`s are applied by a single `ExecutionInput` aka GraphQL Operation, whereas this new instrumentation is applied across a number of operations and stores its state in the `GraphQLContext`.
```

executions/graphql-kotlin-transaction-batcher-instrumentation/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
+description = "Transaction Batcher Instrumentation by level"
```
dariuszkuc, line 1: should we drop `by level`?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/execution/ExecutionLevelInstrumentation.kt
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.transactionbatcher.instrumentation.execution
+
+import graphql.ExecutionInput
+
+interface ExecutionLevelInstrumentation {
```
dariuszkuc, line 21: can we add KDoc to the interface as well?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ConcurrentHashMapExtensions.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * if [ConcurrentHashMap] contains a given [key] execute a given [block] function while
+ * holding the monitor of the given object lock
+ *
+ * @param key to check if there is a value associated with
+ * @param block function to execute over the associated value while holding his monitor
+ * @return result of [block] function that was executed over the associated values
+ */
+fun <K, V, R> ConcurrentHashMap<K, V>.synchronizeIfPresent(key: K, block: (V) -> R): R? =
```
dariuszkuc, line 29: this definitely shouldn't be a public API - maybe we should make it private and move it under `ExecutionLevelInstrumentationState`?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ExecutionContextExtensions.kt
```diff
@@ -0,0 +1,43 @@
...
+import graphql.analysis.QueryTraverser
+import graphql.analysis.QueryVisitorFieldEnvironment
+import graphql.execution.ExecutionContext
+import kotlin.math.max
+
+/**
+ * Calculate the longest path of the [ExecutionContext] AST Document from the root node to a leaf node
+ * @return the height of the AST Document
+ */
+fun ExecutionContext.getDocumentHeight(): Int {
```
dariuszkuc, line 28:
```suggestion
internal fun ExecutionContext.getDocumentHeight(): Int {
```

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/FieldValueInfosExtensions.kt
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.transactionbatcher.instrumentation.extensions
+
+import graphql.execution.FieldValueInfo
+
+fun List<FieldValueInfo>.getExpectedStrategyCalls(): Int {
```
dariuszkuc, line 21:
```suggestion
internal fun List<FieldValueInfo>.getExpectedStrategyCalls(): Int {
```

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/state/ExecutionState.kt
```diff
@@ -0,0 +1,152 @@
...
+package com.expediagroup.graphql.transactionbatcher.instrumentation.state
+
+import graphql.schema.DataFetcher
+
+enum class LevelState { NOT_DISPATCHED, DISPATCHED }
+
+/**
+ * Handle the state of an [graphql.ExecutionInput]
+ */
+class ExecutionState(documentHeight: Int) {
```
dariuszkuc, line 26: maybe rename to `BatchExecutionState`?

dariuszkuc (changes requested):

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/state/ExecutionState.kt
samuelAndalon: i would keep it as `ExecutionState` an instance of this class is associated with to an instance of `ExecutionInput`, in theory this Instrumentation can also apply to a single ExecutionInput (single GraphQLRequest), as it kind of covers what the DataLoaderInstrumentation does, it will batch operations even for a single Execution.

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ConcurrentHashMapExtensions.kt
samuelAndalon: what about `internal` as the other extension functions ?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/DataFetchingEnvironmentExtensions.kt
```diff
@@ -0,0 +1,31 @@
...
+import graphql.schema.DataFetchingEnvironment
+
+/**
+ * get an instance of [TransactionBatcher] from the GraphQLContext
+ * @return a [TransactionBatcher] instance or [MissingTransactionBatcherException] if there
+ * is not a [TransactionBatcher] instance in the GraphQLContext
+ */
+fun DataFetchingEnvironment.transactionBatcher(): TransactionBatcher =
+    this.graphQlContext
+        .get<TransactionBatcher>(TransactionBatcher::class)
```
dariuszkuc, line 30: do you actually need both the generic and the passed value? I thought our inline extension function should have simplified this?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/state/ExecutionState.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    private val levelsState: MutableMap<Level, LevelState> = mutableMapOf(
+        *Array(documentHeight) { number -> Level(number + 1) to LevelState.NOT_DISPATCHED }
+    )
+
+    private val expectedFetches: MutableMap<Level, Int> = mutableMapOf(
+        *Array(documentHeight) { number -> Level(number + 1) to 0 }
+    )
+    private val happenedFetches: MutableMap<Level, Int> = mutableMapOf(
```
dariuszkuc, line 35: `happened` seems kind of weird -> maybe we should rename to `completed` or something along those lines instead?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ConcurrentHashMapExtensions.kt
dariuszkuc: was just wondering - do we even need this? why not just use https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfPresent-K-java.util.function.BiFunction-? which should be thread safe?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/state/ExecutionState.kt
dariuszkuc: it is still related to batching even within single request so I think `ExecutionBatchState` is valid and more meaningful

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ConcurrentHashMapExtensions.kt
samuelAndalon: `computeIfPresent` and `computeIfAbsent` apply a lock on the whole map.

As documentation indicates: 
```
Some attempted update operations on this map by other threads may be blocked while computation is in progress
```
and we are interested on applying a lock on the monitor of the Value as it would be possible to have parallelism altering the concurrentHashMap but we need to be atomic at the value level

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/DataFetchingEnvironmentExtensions.kt
samuelAndalon: the extension function is defined in the generator library, in order to be able to use it i would need to import that library or define the extension function here, or maybe extract all these utility extensions in a dedicated artifact

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/state/ExecutionState.kt
samuelAndalon: in the GraphQLContext `completed` indicated completion stage of a fetch - this is when a completableFuture completes the async work, so maybe we could use a word related to dispatched

something like `fetchersToDispatch` and `dispatchedFetchers` ?

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ConcurrentHashMapExtensions.kt
samuelAndalon: having this extension function will allow to concurrently modify this map without locking it, but actually applying a lock over the value that we are attempting to modify

samuelAndalon: [graphql-java](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/dataloader/FieldLevelTrackingApproach.java#L127) synchronizes the access to the executionState, for the possible parallelism reason

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/state/ExecutionState.kt
dariuszkuc: maybe 
* `happenedFetches` -> `fetched`
* `happenedExecutionStrategies` -> `executedStrategies`
* `happenedOnFieldValueInfos` -> `onFieldValueInfos`

or you could keep the `dispatched` prefix

executions/graphql-kotlin-transaction-batcher-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/ConcurrentHashMapExtensions.kt
dariuszkuc: IMHO you are preemptively attempting to optimize thing that most likely doesn't matter and introduces unncessary complexity - you are just increasing the counts in the `ExecutionBatchState`. Default behavior from `computeIfPresent`/`computeIfAbsent` should be good enough and we should optimize it if there is actually a perf problem with it.

dariuszkuc (approved):

Issue #1370, opened 2022-02-17
Support value classes


rocketraman:
**Is your feature request related to a problem? Please describe.**
I often use value classes in my data models. However, while the schema generation appears to work correctly, at runtime graphql-kotlin fails when attempting to fetch the value class property. When serializing a class that uses them:

```
java.lang.IllegalArgumentException: argument type mismatch
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77) ~[?:?]
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:568) ~[?:?]
	at kotlin.reflect.jvm.internal.calls.InlineClassAwareCaller.call(InlineClassAwareCaller.kt:145) ~[kotlin-reflect-1.6.10.jar:1.6.10-release-923(1.6.10)]
	at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:108) ~[kotlin-reflect-1.6.10.jar:1.6.10-release-923(1.6.10)]
	at com.expediagroup.graphql.generator.execution.PropertyDataFetcher.get(PropertyDataFetcher.kt:34) ~[graphql-kotlin-schema-generator-5.3.2.jar:5.3.2]
	...
```

An explicit `GraphQLScalarType` with a `Coercing` implementation does not help.

**Describe the solution you'd like**
Value classes should not cause any issues.

dariuszkuc:
Hello 👋,
Guessing this is due to the https://github.com/FasterXML/jackson-module-kotlin/issues/199. I believe with the proposed refactor of how we parse the input arguments (https://github.com/ExpediaGroup/graphql-kotlin/pull/1379) we should be able to support it.

dariuszkuc:
#1379 added support for handling ID value class only, looks like we may need to make some additional changes to correctly handle it

Issue #1312, opened 2021-11-30
OptionalInput<List<UUID>> is parsed as OptionalInput<List<String>> resulting in runtime ClassCastException


rocketraman:
**Library Version**
5.2.0

**Describe the bug**
An input type containing `OptionalInput<List<UUID>>` is parsed by GraphQL as `OptionalInput<List<String>>`. The runtime type mismatch results in a runtime `ClassCastException` when the code expects to see a UUID, but actually gets a String.

A workaround that can be used in most situations is to accept `List<UUID>?` and treat an empty list vs a null list as the difference between provided+empty and absent. However, the field semantics may not always allow for this.

**To Reproduce**
Steps to reproduce the behavior. Please provide:

```
data class Foo(
  val l: OptionalInput<List<UUID>>
)
```

```
val graphqlUuidType: GraphQLScalarType = GraphQLScalarType.newScalar()
  // lower-case name matches Hasura
  .name("uuid")
  .description("A type representing a formatted UUID")
  .coercing(UUIDCoercing)
  .build()

object UUIDCoercing : Coercing<UUID, String> {
  override fun parseValue(input: Any): UUID = UUID.fromString(serialize(input))

  override fun parseLiteral(input: Any): UUID {
    val s = (input as StringValue).value
    return UUID.fromString(s)
  }

  override fun serialize(dataFetcherResult: Any): String = coerceSafe(dataFetcherResult) { toString() }
}

class CustomSchemaGeneratorHooks(config: SchemaConfig) : SchemaGeneratorHooks {
  override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier as? KClass<*>) {
    UUID::class -> graphqlUuidType
    else -> null
  }
}
```

**Expected behavior**
I would expect the runtime type to be the same as the compile-time type.

dariuszkuc:
Hello 👋 
This sounds like the same underlying issue as https://github.com/ExpediaGroup/graphql-kotlin/issues/1220

PR #1379, opened 2022-02-26
[generator] avoid duplicate argument deserialization

dariuszkuc:
### :pencil: Description

Update function data fetcher argument parsing logic to no longer depend on Jackson. Current logic is problematic as arguments passed to data fetchers are already coerced to appropriate type by `graphql-java` and Jackson is unaware of it leading to duplicate deserialization. This leads to custom scalar inputs deserialization problems.

Instead of relying on Jackson, argument mapping logic is updated to rely on reflection to instantiate objects based on their primary constructor.

NOTE: this change removes dependency on Jackson from `graphql-kotlin-schema-generator`. `graphql-kotlin-server` is still tied to Jackson for execution.

Breaking changes:
* `FunctionDataFetcher` no longer accepts Jackson `ObjectMapper` as an argument
* arrays are no longer supported by the schema generator

### :link: Related Issues

* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1220
* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1312
* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1370
* based on https://github.com/ExpediaGroup/graphql-kotlin/pull/1255


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/exceptions/CouldNotConstructAValidKotlinObject.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.generator.exceptions
+
+class CouldNotConstructAValidKotlinObject(input: Any?) : GraphQLKotlinException("Could not convert the input $input to a valid Kotlin object")
```
smyrick, line 19: You might want to have the input be the last item in the log message

`Could not convert the input to a valid Kotlin object: $input`

website/docs/schema-generator/writing-schemas/lists.md
```diff
@@ -35,22 +35,8 @@ type Query {
...
-| `kotlin.FloatArray`   |
-| `kotlin.DoubleArray`  |
-| `kotlin.CharArray`    |
-| `kotlin.BooleanArray` |
-
-:::note
-The underlying GraphQL types of primitive arrays will be corresponding to the built-in scalar types provided by `graphql-java`.
-:::
+Due to the [argument deserialization issues](https://github.com/ExpediaGroup/graphql-kotlin/pull/1379), primitive arrays
+are currently not supported.
```
smyrick, line 39: We should update the docs below as well for Unsupported Collection Types

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/exceptions/CouldNotConstructAValidKotlinObject.kt
dariuszkuc: This error was misleading, updated exception to indicate that we are missing public primary constructor. This exception thrown both at schema generation and as a safeguard all at runtime (which should never happen).

samuelAndalon (approved):

PR #1383, opened 2022-03-02
Class name in SchemaGeneratorHooksProvider example

ashu-dd:
The class name for the service provider is incorrect in example for SchemaGeneratorHooksProvider


dariuszkuc:
Thanks for the fix!

PR #1388, opened 2022-03-15
feat(server): add TransactionBatcher to graphQLContext

samuelAndalon:
### :pencil: Description
Adding `TransactionBatcher` and `ExecutionLevelInstrumentationState` instances to the GraphQLContext that will be used as a context for all `ExecutionInput`  aka `GraphQLOperation` in a Request (single or batched).

This way client only responsibility will be opt it or opt out from the optional `TransactionBatcherLevelInstrumentation`

### DIFF
--- a/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
+++ b/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
@@ -80,22 +80,26 @@ class GraphQLSchemaConfiguration {
             .also { builder ->
                 executionIdProvider.ifPresent(builder::executionIdProvider)
                 preparsedDocumentProvider.ifPresent(builder::preparsedDocumentProvider)
-            }
 
-        val instrumentations = mutableListOf<Instrumentation>()
-        if (config.batching.enabled) {
-            instrumentations.add(TransactionBatcherLevelInstrumentation())
-        }
-
-        providedInstrumentations.ifPresent { unorderedInstrumentations ->
-            instrumentations.addAll(
-                unorderedInstrumentations.sortedBy { instrumentation ->
-                    (instrumentation as? Ordered)?.order ?: DEFAULT_INSTRUMENTATION_ORDER
+                val instrumentations = mutableListOf<Instrumentation>()
+                if (config.batching.enabled) {
+                    builder.doNotAddDefaultInstrumentations()
+                    instrumentations.add(TransactionBatcherLevelInstrumentation())
                 }
-            )
-        }
 
-        graphQLBuilder.instrumentation(ChainedInstrumentation(instrumentations))
+                providedInstrumentations.ifPresent { unorderedInstrumentations ->
+                    instrumentations.addAll(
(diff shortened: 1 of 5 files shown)


servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -60,45 +61,43 @@ const val DEFAULT_INSTRUMENTATION_ORDER = 0
...
-                }
-                graphQL.instrumentation(ChainedInstrumentation(sorted))
+            .also { builder ->
+                executionIdProvider.ifPresent(builder::executionIdProvider)
+                preparsedDocumentProvider.ifPresent(builder::preparsedDocumentProvider)
             }
+
+        val instrumentations = mutableListOf<Instrumentation>()
+        if (config.batching.enabled) {
+            instrumentations.add(TransactionBatcherLevelInstrumentation())
```
dariuszkuc, line 87: should this also disable the default `DataLoaderInstrumentation`?

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -60,45 +61,43 @@ const val DEFAULT_INSTRUMENTATION_ORDER = 0
...
-        executionIdProvider.ifPresent {
-            graphQL.executionIdProvider(it)
-        }
-        preparsedDocumentProvider.ifPresent {
-            graphQL.preparsedDocumentProvider(it)
+
+        providedInstrumentations.ifPresent { unorderedInstrumentations ->
+            instrumentations.addAll(
+                unorderedInstrumentations.sortedBy { instrumentation ->
+                    (instrumentation as? Ordered)?.order ?: DEFAULT_INSTRUMENTATION_ORDER
```
dariuszkuc, line 93: while this is functionally the same IMHO previous version

```kotlin
if (instrumentation is Ordered) {
  instrumentation.order
} else {
  DEFAULT_INSTRUMENTATION_ORDER
}
```

is more readable so lets revert this back

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -60,45 +61,43 @@ const val DEFAULT_INSTRUMENTATION_ORDER = 0
...
-                        DEFAULT_INSTRUMENTATION_ORDER
-                    }
-                }
-                graphQL.instrumentation(ChainedInstrumentation(sorted))
+            .also { builder ->
+                executionIdProvider.ifPresent(builder::executionIdProvider)
+                preparsedDocumentProvider.ifPresent(builder::preparsedDocumentProvider)
             }
+
+        val instrumentations = mutableListOf<Instrumentation>()
```
dariuszkuc, line 85: if we already introducing that `also` block above -> should we move this inside as well?

PR #1389, opened 2022-03-17
feat: TransactionLoader to use DataLoaders

samuelAndalon:


samuelAndalon:
closing PR in favor of always using DataLoaders

Issue #1392, opened 2022-03-22
graphql-java 18.0 support


cryptoki:
I tried to bump up graphql-kotlin to graphql-java 18.0. It seems a bit more work. :) Is there any task already in doing?

dariuszkuc:
Hello 👋,
Thanks for taking the initiative with the update! Indeed the bump requires quite some changes (big change is the directive -> applied directive split). I'm working on it but still got to fix a number of unit tests.

PR #1393, opened 2022-03-23
[dependency] update to graphql-java 18

dariuszkuc:
### :pencil: Description

Update to `graphql-java` v18.

Adds support for **applied** schema directive. In previous versions, `GraphQLDirective` represented both a schema definition object as well as individual instances that were applied to a schema or query element. Schema generation logic was updated to continue generating `GraphQLDirective` schema definitions and then create a corresponding instance of `GraphQLAppliedDirective` that is applied to a target element.

Directives are now validated whether they are applicable to a target element at build time instead of runtime. If directive is applied in invalid location, new `InvalidDirectiveLocationException` is thrown. Previously this validation was (incorrectly) performed at runtime in the `KotlinDirectiveWiringFactory`.

see: https://github.com/graphql-java/graphql-java/releases/tag/v18.0

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1392


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateDirective.kt
```diff
@@ -70,38 +80,40 @@ private fun getDirective(generator: SchemaGenerator, directiveInfo: DirectiveInf
...
     } else {
-        directive
+        appliedDirective
     }
 }
 
-private fun generateDirectiveArgument(prop: KProperty<*>, directiveInfo: DirectiveInfo, generator: SchemaGenerator): GraphQLArgument {
+private fun generateDirectiveArgument(prop: KProperty<*>, generator: SchemaGenerator): GraphQLArgument {
     val propertyName = prop.name
-    val value = prop.call(directiveInfo.directive)
```
dariuszkuc, line 99: This never worked for default values as it is called on an instance of the annotation. Unsure how it worked for federation before (guessing values were not printed in the sdl).

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/types/_FieldSet.kt
```diff
@@ -45,9 +48,19 @@ private object FieldSetCoercing : Coercing<FieldSet, String> {
...
+        when (input) {
+            is FieldSet -> input
+            is StringValue -> FieldSet::class.constructors.first().call(input.value)
+            else -> {
+                throw CoercingParseLiteralException("Cannot parse $input to FieldSet. Expected AST type 'StringValue' but was '${input.javaClass.simpleName}'.")
+            }
+        }
+
+    override fun valueToLiteral(input: Any): Value<out Value<*>> {
+        if (input is FieldSet) {
```
samuelAndalon, line 60: for consistency with ^ method, would it worth to use `when` ?

samuelAndalon (approved):

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/types/_FieldSet.kt
dariuszkuc: for consistency changed to expression body syntax

PR #1395, opened 2022-03-24
[docs] documentation on inline value classes

dariuszkuc:
### :pencil: Description

It is often beneficial to create a wrapper around the underlying primitive type to better represent its meaning. Inline value classes can be used to optimize such use cases - Kotlin compiler will attempt to use underlying type directly whenever possible and only keep the wrapper classes whenever it is necessary.

In order to use inline value classes in your schema, you need to register it using hooks and also provide value unboxer that will be used by `graphql-java` when dealing with its wrapper object.

```kotlin
@JvmInline
value class MyValueClass(
    val value: String
)

class MyQuery : Query {
    fun inlineValueClassQuery(value: MyValueClass? = null): MyValueClass = value ?: MyValueClass("default")
}

class MySchemaGeneratorHooks : SchemaGeneratorHooks {
    override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier) {
        MyValueClass::class -> Scalars.GraphQLString
        else -> null
    }
}

class MyValueUnboxer : IDValueUnboxer() {
    override fun unbox(`object`: Any?): Any? = if (`object` is MyValueClass) {
        `object`.value
    } else {
        super.unbox(`object`)
    }
}

val config = SchemaGeneratorConfig(
    supportedPackages = listOf("com.example"),
    hooks = MySchemaGeneratorHooks()
)
val schema = toSchema(
    config = config,
    queries = listOf(TopLevelObject(MyQuery()))
)
val graphQL = GraphQL.newGraphQL(graphQLSchema)
    .valueUnboxer(IDValueUnboxer())
    .build()
```

This will generate the schema that exposes value classes as corresponding primitive types in the schema

```graphql
type Query {
  inlineValueClassQuery(value: String): String!
}
```

### :link: Related Issues

Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1370

### DIFF
--- a/website/docs/schema-generator/writing-schemas/scalars.md
+++ b/website/docs/schema-generator/writing-schemas/scalars.md
@@ -42,7 +42,7 @@ val graphQL = GraphQL.newGraphQL(graphQLSchema)
 ```
 
 :::note
-`IDValueUnboxer` is automatically configured by `graphql-kotlin-spring-server`.
+`IDValueUnboxer` bean is automatically configured by `graphql-kotlin-spring-server`.
 :::
 
 ```kotlin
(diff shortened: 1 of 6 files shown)


samuelAndalon (approved):

website/docs/schema-generator/writing-schemas/scalars.md
```diff
@@ -99,19 +99,98 @@ val graphqlUUIDType = GraphQLScalarType.newScalar()
...
+val config = SchemaGeneratorConfig(
+    supportedPackages = listOf("com.example"),
+    hooks = MySchemaGeneratorHooks()
+)
+val schema = toSchema(
+    config = config,
+    queries = listOf(TopLevelObject(MyQuery()))
+)
+val graphQL = GraphQL.newGraphQL(graphQLSchema)
+    .valueUnboxer(IDValueUnboxer())
```
violeta-co, line 170: hi, quick question: not entirely sure but should this be: `.valueUnboxer(MyValueUnboxer())` ?  haven't run this particular example locally, but tried a similar one

dariuszkuc: yeah, my bad yes it should :) copy & paste error, thanks for picking it up

website/docs/schema-generator/writing-schemas/scalars.md
```diff
@@ -99,19 +99,98 @@ val graphqlUUIDType = GraphQLScalarType.newScalar()
...
+}
+
+class MySchemaGeneratorHooks : SchemaGeneratorHooks {
+    override fun willGenerateGraphQLType(type: KType): GraphQLType? = when (type.classifier) {
+        MyValueClass::class -> Scalars.GraphQLString
+        else -> null
+    }
+}
+
+class MyValueUnboxer : IDValueUnboxer() {
```
rocketraman, line 153: @dariuszkuc `IDValueUnboxer` seems to be a Spring-specific bean meant for handling IDs. To handle arbitrary value classes, I think this should extend `DefaultValueUnboxer` instead, shouldn't it?

dariuszkuc: `IDValueUnboxer` is part of the schema generator lib and is not tied to Spring (by default our server does create bean for you). You can technically specify any `ValueUnboxer` as long as it handles all the inline value classes - e.g. if you need `ID` support then it will work as long you add the logic to unwrap the `ID` value (which is why we suggest to extend the `IDValueUnboxer`).

PR #1411, opened 2022-04-06
feat: remove transaction-batcher module

samuelAndalon:
### :pencil: Description
Removing `transaction-batcher` module in favor of using `graphql-java-dataloader` and DataLoaderRegistry, make the `TransactionLoaderLevelInstrumentation` work with `DataLoaderRegistrry`


executions/graphql-kotlin-transaction-loader-instrumentation/build.gradle.kts
```diff
@@ -1,12 +1,12 @@
...
+description = "Transaction Loader Instrumentations"
 
 val junitVersion: String by project
 val graphQLJavaVersion: String by project
 val reactorVersion: String by project
 val reactorExtensionsVersion: String by project
 
 dependencies {
-    api(project(path = ":graphql-kotlin-transaction-batcher"))
+    api(project(path = ":graphql-kotlin-server"))
```
dariuszkuc, line 9: it seems somewhat odd that instrumentation would depend on the server?

dariuszkuc:
Since we want to move towards the sync exhaustion instrumentation (#1407) -> is there any point in doing this work (vs just dropping the modules)?

executions/graphql-kotlin-transaction-loader-instrumentation/src/main/kotlin/com/expediagroup/graphql/transactionbatcher/instrumentation/extensions/DataFetchingEnvironmentExtensions.kt
```diff
@@ -16,16 +16,19 @@
...
 /**
- * get an instance of [TransactionBatcher] from the GraphQLContext
- * @return a [TransactionBatcher] instance or [MissingTransactionBatcherException] if there
- * is not a [TransactionBatcher] instance in the GraphQLContext
+ * get an instance of [DataLoaderRegistry] from the [GraphQLContext]
+ * @return [DataLoaderRegistry] instance
+ * @throws [MissingDataLoaderRegistryException] if there is not a [DataLoaderRegistry] instance in the [GraphQLContext]
  */
-fun DataFetchingEnvironment.transactionBatcher(): TransactionBatcher =
+fun <K, V> DataFetchingEnvironment.getContextDataLoader(key: String): DataLoader<K, V> =
```
dariuszkuc, line 30: wasn't sure what `context data loader` is -> maybe this should be renamed to something like `getDataLoaderFromContext`?

executions/graphql-kotlin-transaction-loader-instrumentation/build.gradle.kts
samuelAndalon: oops, forgot to specify it as a `testImplementation` main reason is because for the unit tests im using the `KotlinDataLoader`, we can 

1. Move GraphQL Kotlin DataLoader stuff to a different library
2. Use DataLoaders directly in tests

for now will just  specify `:graphql-kotlin-server`  as a testImplementation

samuelAndalon:
there is a discussion that i would like to have regarding the sync execution exhaustion. What i would like to do is for now keep by level instrumentation and have the discussion on the PR of sync exhaustion.

dariuszkuc (approved):

PR #1419, opened 2022-04-14
feat: DataLoader Sync Execution Exhaustion Instrumentation

samuelAndalon:
### :pencil: Description
Adding a Custom Instrumentation that will dispatch a [DataLoaderRegistry](https://github.com/graphql-java/java-dataloader/blob/master/src/main/java/org/dataloader/DataLoaderRegistry.java), when the synchronous execution of all operations sharing the same `GraphQLContext` was **exhausted**

by **exhausted** it means that all the AST paths were visited until finding a leaf scalar or a non-trivial dataFetcher that returns a `CompletableFuture`

This Instrumentation keeps track on the state of each [ExecutionStrategy](https://www.javadoc.io/doc/com.graphql-java/graphql-java/18.0/graphql/execution/ExecutionStrategy.html) associated with an operation using the following data structure

image

image

### :link: Related Issues
This PR is related to [this discussion opened in graphql-java](https://github.com/graphql-java/graphql-java/discussions/2715)

### DIFF
--- a/executions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/fixture/extensions/ListExtensions.kt
+++ b/executions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/fixture/extensions/ListExtensions.kt
@@ -1,0 +1,10 @@
+package com.expediagroup.graphql.dataloader.instrumentation.fixture.extensions
+
+import java.util.Optional
+
+fun <T : Any> List<Optional<T>>.toListOfNullables(): List<T?> = map { optional ->
+    when {
+        optional.isPresent -> optional.get()
+        else -> null
+    }
+}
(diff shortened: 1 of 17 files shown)


executions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/fixture/extensions/ListExtensions.kt
```diff
@@ -0,0 +1,10 @@
+package com.expediagroup.graphql.dataloader.instrumentation.fixture.extensions
+
+import java.util.Optional
+
+fun <T : Any> List<Optional<T>>.toListOfNullables(): List<T?> = map { optional ->
```
dariuszkuc, line 5:
```suggestion
internal fun <T : Any> List<Optional<T>>.toListOfNullables(): List<T?> = map { optional ->
```

dariuszkuc (changes requested):
Do we need to update README with new names?

dariuszkuc (approved):

PR #1423, opened 2022-04-18
feat: add dataLoader instrumentation configuration to server

samuelAndalon:
### :pencil: Description
Auto Configuration (**disabled** by default) to enable `DataLoaderLevelDispatchedInstrumentation`, `DataLoaderLevelDispatchedInstrumentation` instances to the GraphQLContext that will be used as a context for all ExecutionInput in a `GraphQLBatchRequest`

This way client only responsibility will be opt it or opt out from the optional instrumentations

```yml
graphql:
  batching: 
   enabled: true # or false 
   strategy: LEVEL_DISPATCHED # or SYNC_EXHAUSTION
```


servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -60,45 +62,52 @@ const val DEFAULT_INSTRUMENTATION_ORDER = 0
...
             .queryExecutionStrategy(AsyncExecutionStrategy(dataFetcherExceptionHandler))
             .mutationExecutionStrategy(AsyncSerialExecutionStrategy(dataFetcherExceptionHandler))
             .subscriptionExecutionStrategy(FlowSubscriptionExecutionStrategy(dataFetcherExceptionHandler))
             .valueUnboxer(idValueUnboxer)
+            .also { builder ->
+                executionIdProvider.ifPresent(builder::executionIdProvider)
+                preparsedDocumentProvider.ifPresent(builder::preparsedDocumentProvider)
+
+                val instrumentations = mutableListOf<Instrumentation>()
+                if (config.batching.enabled) {
```
samuelAndalon, line 86: add batch instrumentations if batching is enabled, otherwise stick to the default behavior which is adding the `DefaultDataLoaderDispatcherInstrumentation` from graphql-java, which applies to a single operation

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,136 @@
...
+                    execute(
+                        graphQLRequest.toExecutionInput(context, dataLoaderRegistryFactory?.generate(), graphQLContext)
+                    )
+                )
+            }
+            is GraphQLBatchRequest -> {
+                when {
+                    graphQLRequest.requests.none(GraphQLRequest::isMutation) -> {
+                        val dataLoaderRegistry = dataLoaderRegistryFactory?.generate()
+                        val contextWithDataLoaderInstrumentationState = graphQLContext + (
```
samuelAndalon, line 83: this part is tricky, given that at this layer we don't have access to the configuration, we need to check the `GraphQL` instance instrumentations to check if there is a `DataLoaderLevelDispatchedInstrumentation` or `DataLoaderSyncExecutionExhaustedInstrumentation`, if so, it means that batching is enabled and we need to add the corresponding `State` instance to the `GraphQLContext` before attempting to execute the operations in a concurrent way.

The DataLoaderRegistry needs to be created only once for all operations in a `GraphQLBatchRequest`

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,125 @@
...
+                                        )
+                                    )
+                                    else -> null
+                                }
+                            } ?: emptyMap()
+                        )
+                        GraphQLBatchResponse(
+                            execute(
+                                graphQLRequest.requests.map {
+                                    it.toExecutionInput(context, null, contextWithDataLoaderInstrumentationState)
```
samuelAndalon, line 102: if batching enabled either using by `level` or `sync execution exhaustion`, do not add the `DataLoaderRegistry` to the execution input, no need to, the `DataLoaderRegistry` is located in the shared `GraphQLContext`

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,125 @@
...
+import kotlinx.coroutines.supervisorScope
 
 open class GraphQLRequestHandler(
     private val graphQL: GraphQL,
-    private val dataLoaderRegistryFactory: DataLoaderRegistryFactory? = null
+    private val dataLoaderRegistryFactory: KotlinDataLoaderRegistryFactory? = null
 ) {
 
+    private val batchingInstrumentationStrategy: Class<out Instrumentation>? =
+        (graphQL.instrumentation as? ChainedInstrumentation)?.let { chainedInstrumentation ->
```
dariuszkuc, line 51: would it be possible to single instrumentation for batching and nothing else? i.e. is there a scenario where it wouldn't be a chained instrumentation?

samuelAndalon: the GraphQL instance is configured to use ChainedInstrumentation but, yes, if someone creates the instance on his own with a single instrumentation for batching, will add that case.

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/InstrumentationExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package com.expediagroup.graphql.server.extensions
+
+import com.expediagroup.graphql.dataloader.instrumentation.level.DataLoaderLevelDispatchedInstrumentation
+import com.expediagroup.graphql.dataloader.instrumentation.syncexhaustion.DataLoaderSyncExecutionExhaustedInstrumentation
+import graphql.execution.instrumentation.Instrumentation
+
+internal fun Instrumentation.isDataLoaderInstrumentation(): Boolean =
```
dariuszkuc, line 7:
```suggestion
internal fun Instrumentation.isBatchDataLoaderInstrumentation(): Boolean =
```

dariuszkuc, line 7: guess it doesn't really matter but based on the naming conventions this file name should start with lowercase

servers/graphql-kotlin-server/src/test/kotlin/com/expediagroup/graphql/server/execution/GraphQLServerTest.kt
```diff
@@ -55,7 +55,7 @@ class GraphQLServerTest {
 
         val server = GraphQLServer(mockParser, mockContextFactory, mockHandler)
 
-        runBlockingTest { server.execute(mockk()) }
+        runBlocking { server.execute(mockk()) }
```
dariuszkuc, line 58: based on deprecation notice should this be updated to `runTest` instead?

servers/graphql-kotlin-server/src/test/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandlerTest.kt
```diff
@@ -16,21 +16,32 @@
...
 import graphql.execution.AbortExecutionException
+import graphql.execution.instrumentation.ChainedInstrumentation
+import graphql.execution.instrumentation.Instrumentation
 import graphql.schema.DataFetchingEnvironment
 import graphql.schema.GraphQLSchema
 import io.mockk.every
 import io.mockk.mockk
-import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.runBlockingTest
+import kotlinx.coroutines.runBlocking
```
dariuszkuc, line 41: same as in other comment -> this should be `runTest` instead

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,134 @@
...
-    open suspend fun executeRequest(request: GraphQLRequest, context: GraphQLContext? = null, graphQLContext: Map<*, Any> = emptyMap<Any, Any>()): GraphQLResponse<*> {
-        // We should generate a new registry for every request
-        val dataLoaderRegistry = dataLoaderRegistryFactory?.generate()
-        val executionInput = request.toExecutionInput(context, dataLoaderRegistry, graphQLContext)
+    open suspend fun executeRequest(
+        graphQLRequest: GraphQLServerRequest,
+        context: GraphQLContext? = null,
+        graphQLContext: Map<*, Any> = emptyMap<Any, Any>()
+    ): GraphQLServerResponse =
+        when (graphQLRequest) {
```
dariuszkuc, line 75: I think we can simplify this by moving some stuff around


```kotlin
when (graphQLRequest) {
    is GraphQLRequest -> execute(graphQLRequest, context, graphQLContext)
    is GraphQLBatchRequest -> {
        if (graphQLRequest.containsMutation()) {
            executeBatch(graphQLRequest, context, graphQLContext)
        } else {
            executeBatchAsync(graphQLRequest, context, graphQLContext)
        }
    }
}
```

refactored methods

```kotlin
private suspend fun execute(request: GraphQLRequest, context: GraphQLContext?, graphQLContext: Map<*, Any>, dataLoaderRegistry: KotlinDataLoaderRegistry? = dataLoaderRegistryFactory?.generate()): GraphQLResponse<*> {
    val executionInput = request.toExecutionInput(context, dataLoaderRegistry, graphQLContext)
    return try {
        graphQL.executeAsync(executionInput).await().toGraphQLResponse()
    } catch (exception: Exception) {
        val error = exception.toGraphQLError()
        GraphQLResponse<Any?>(errors = listOf(error.toGraphQLKotlinType()))
    }
}

private suspend fun executeBatch(request: GraphQLBatchRequest, context: GraphQLContext?, graphQLContext: Map<*, Any>): GraphQLBatchResponse =
    GraphQLBatchResponse(request.requests.map { execute(it, context, graphQLContext) })

private suspend fun executeBatchAsync(request: GraphQLBatchRequest, context: GraphQLContext?, graphQLContext: Map<*, Any>): GraphQLBatchResponse {
    val (dataLoaderRegistry, batchContext) = batchingContext(request.requests.size)
    val batchGraphQLContext = graphQLContext + batchContext
    return supervisorScope {
        val responses = request.requests.map {
            async {
                execute(it, context, batchGraphQLContext, dataLoaderRegistry)
            }
        }
        GraphQLBatchResponse(responses.awaitAll())
    }
}

private fun batchingContext(batchSize: Int): Pair<KotlinDataLoaderRegistry?, Map<*, Any>> {
    val dataLoaderRegistry = dataLoaderRegistryFactory?.generate()
    val context = dataLoaderRegistry?.let {
        when (dataLoaderInstrumentationType) {
            DataLoaderLevelDispatchedInstrumentation::class.java -> mapOf(
                KotlinDataLoaderRegistry::class to dataLoaderRegistry,
                ExecutionLevelDispatchedState::class to ExecutionLevelDispatchedState(batchSize)
            )
            DataLoaderSyncExecutionExhaustedInstrumentation::class.java -> mapOf(
                KotlinDataLoaderRegistry::class to dataLoaderRegistry,
                SyncExecutionExhaustedState::class to SyncExecutionExhaustedState(batchSize, dataLoaderRegistry)
            )
            else -> null
        }
    } ?: emptyMap<Any, Any>()

    return if (context.isEmpty()) {
        dataLoaderRegistry to context
    } else {
        null to context
    }
}
```

*with new extension
```kotlin
fun GraphQLBatchRequest.containsMutation(): Boolean = requests.any { it.query.contains("mutation ") }
```

dariuszkuc (changes requested):

dariuszkuc:
We will also need documentation on how to use batching in the GH pages

servers/graphql-kotlin-server/src/test/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandlerTest.kt
samuelAndalon: i guess its fine, the main difference between `runBlocking` and `runTest` is that runTest will skip delays, i liked more runBlocking bc
1. we are not delaying
2. wanted to block only when needed, also `runBlocking` returns the result of the lambda argument which is not the case of `runTest`

samuelAndalon:
> We will also need documentation on how to use batching in the GH pages

will work on those.

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,139 @@
...
 import kotlinx.coroutines.future.await
+import kotlinx.coroutines.supervisorScope
 
 open class GraphQLRequestHandler(
     private val graphQL: GraphQL,
-    private val dataLoaderRegistryFactory: DataLoaderRegistryFactory? = null
+    private val dataLoaderRegistryFactory: KotlinDataLoaderRegistryFactory? = null
 ) {
 
+    private val dataLoaderInstrumentationType: Class<Instrumentation>? =
```
dariuszkuc, line 51:
```suggestion
    private val batchDataLoaderInstrumentationType: Class<Instrumentation>? =
```

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,139 @@
...
+                    executeBatchConcurrently(graphQLRequest, context, graphQLContext)
+                }
+            }
+        }
 
+    private suspend fun execute(
+        graphQLRequest: GraphQLRequest,
+        context: GraphQLContext? = null,
+        graphQLContext: Map<*, Any> = emptyMap<Any, Any>(),
+        dataLoaderRegistry: KotlinDataLoaderRegistry? = null
```
dariuszkuc, line 91: instead of `null` if it defaults to `dataLoaderRegistryFactory?.generate()` you would only need to pass it from batch async method

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,139 @@
...
+        request: GraphQLBatchRequest,
+        context: GraphQLContext?,
+        graphQLContext: Map<*, Any>
+    ): GraphQLBatchResponse {
+        val batchContext = getBatchContext(request, dataLoaderRegistryFactory?.generate())
+        val batchGraphQLContext = graphQLContext + batchContext
+        val responses = supervisorScope {
+            request.requests.map {
+                async {
+                    execute(it, context, batchGraphQLContext)
```
dariuszkuc, line 125: depending whether there is batch instrumentation or not you may have to provide the dataloader instance either in graphql context (batch instrumentation enabled) OR as part of the execution input (no batch)

dariuszkuc (changes requested):

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
samuelAndalon: i agree, and i think there is some value on passing the DataLoaderRegistry instance even if we need to execute sequentially, we are not going to batch but we can still use the `ValueCache` of the DataLoader, this to avoid retrieving data if it was already retrieved

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -117,12 +115,13 @@ open class GraphQLRequestHandler(
...
     ): GraphQLBatchResponse {
-        val batchContext = getBatchContext(request, dataLoaderRegistryFactory?.generate())
+        val dataLoaderRegistry = dataLoaderRegistryFactory?.generate()
+        val batchContext = getBatchContext(request, dataLoaderRegistry)
         val batchGraphQLContext = graphQLContext + batchContext
         val responses = supervisorScope {
             request.requests.map {
                 async {
-                    execute(it, context, batchGraphQLContext)
+                    execute(it, context, batchGraphQLContext, dataLoaderRegistry)
```
dariuszkuc, line 124: I don't think this is correct -> I believe that `batchContext` is empty then it should be NULL, otherwise yes its the one thats generated, see my previous comment

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt
```diff
@@ -16,37 +16,153 @@
...
             GraphQLResponse<Any?>(errors = listOf(error.toGraphQLKotlinType()))
         }
+
+    private suspend fun executeBatchSequentially(
+        batchRequest: GraphQLBatchRequest,
+        context: GraphQLContext?,
+        graphQLContext: Map<*, Any>
+    ): GraphQLBatchResponse {
+        val dataLoaderRegistry = dataLoaderRegistryFactory?.generate()
+        // only 1 operation will be executed at a time, still we can use batching and deduplication
```
dariuszkuc, line 106: should this be applied to all then? (i.e. even if it is a single call)

samuelAndalon: when executing batch sequentially only 1 operation will be executed at a time
the instrumentations have this mechanism in place to check when to dispatch, and for that it needs to know how
many operations will be executed concurrently, when doing it sequentially will always be 1

dariuszkuc: Yes it would be applied for single op. I thought that new instrumentation still provided some additional functionality over default dd instrumentation.

samuelAndalon: if clients configure server to use `DataLoaderSyncExecutionExhaustedInstrumentation` they will get the [additional functionality of sync exhaustion](cutions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/syncexhaustion/DataLoaderSyncExecutionExhaustedInstrumentationTest.kt#L229)

dariuszkuc: sure but that also implies the usage of the dataloader through the graphql context?

samuelAndalon: yes, unfortunately, lets see if we can add this instrumentations to graphql-java and always rely on the registry to be placed in the context instead of the execution input

website/docs/server/data-loaders-instrumentations.md
```diff
@@ -0,0 +1,195 @@
...
+    * `nasa` **DataFetcher** will return a `Nasa` object, so we can descend more that path.
+2. Start an `ExecutionStrategy` for the `nasa` field of the `root` field of the query to concurrently resolve `astronaut`, `address` and `phoneNumber`.
+    * `astronaut` **DataFetcher** will return a `CompletableFuture<Astronaut>` so we can consider this path exhausted
+    * `address` **DataFetcher** will return an object, so we can descend more that path.
+    * `phoneNumber` **DataFetcher** will return a scalar, so we can consider this path exhausted.
+3. Start an `ExecutionStrategy` for the `address` field of the `nasa` field to concurrently resolve `street` and `zipCode`.
+    * `street` **DataFetcher** will return a scalar, so we can consider this path exhausted.
+    * `zipCode` **DataFetcher** will return a scalar, so we can consider this path exhausted.
+
+The same logic will apply if we split that query in 2 because `graphql-kotlin-dataloader-instrumentations`
```
dariuszkuc, line 146:
```suggestion
By default, each GraphQL operation is processed independently of each other. Multiple operations can be processed together as if they were single GraphQL request if they are part of the same batch request and server is using one of the `graphql-kotlin-dataloader-instrumentation`s.
```

website/docs/server/data-loaders-instrumentations.md
```diff
@@ -0,0 +1,195 @@
...
+      }
+    """
+]
+
+val graphQLContext = mapOf(
+    KotlinDataLoaderRegistry::class to kotlinDataLoaderRegistry,
+    SyncExecutionExhaustedState::class to SyncExecutionExhaustedState(queries.size, kotlinDataLoaderRegistry)
+)
+
+val result1 = graphQL.executeAsync(ExecutionInput.newExecutionInput(queries[0]).graphQLContext(graphQLContext).build())
```
dariuszkuc, line 187: same comment for code as above

dariuszkuc (changes requested):

website/docs/server/data-loaders.md
```diff
@@ -51,49 +20,87 @@ interface KotlinDataLoader<K, V> {
...
 ```
 
-This allows for library users to still have full control over the creation of the `DataLoader` and its various configuration options,
-but then allows common server code to handle the registration, generation on request, and execution.
+This allows for library users to still have full control over the creation of the `DataLoader` and his configuration options
+and allows server code to handle the registration, generation and execution by request.
+
+## `KotlinDataLoaderRegistryFactory`
+
+The [GraphQLRequestHandler](graphql-request-handler.md) accepts an optional `KotlinDataLoaderRegistryFactory`.
```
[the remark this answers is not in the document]
samuelAndalon: will remove it

website/docs/server/data-loaders-instrumentations.md
```diff
@@ -0,0 +1,195 @@
...
+   of all paths completed up until a scalar leaf or dispatched a `CompletableFuture`.
+
+Given set of GraphQL queries, the server can execute them concurrently, by taking advantage of concurrency we can batch and deduplicate
+data fetching logic for multiple operations.
+
+## DataLoaderLevelDispatchedInstrumentation
+
+ Consider the following example:
+
+![Image of data loader level dispatched instrumentation](../assets/data-loader-level-dispatched-instrumentation.png)
```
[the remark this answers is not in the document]
samuelAndalon: will work on the diagrams

website/docs/server/data-loader/data-loader-instrumentation.md
```diff
@@ -0,0 +1,189 @@
...
+id: data-loader-instrumentation
+title: Data Loader Instrumentations
+---
+
+`graphql-kotlin-dataloader-instrumentation` is set of custom [Instrumentations](https://www.graphql-java.com/documentation/instrumentation/)
+that will calculate when is the right moment to dispatch `KotlinDataLoader`s across single and batch GraphQL operations.
+
+These custom instrumentations follow the similar approach as the default [DataLoaderDispatcherInstrumentation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/dataloader/DataLoaderDispatcherInstrumentation.java)
+from `graphql-java`, the main difference is that regular instrumentations apply to a single `ExecutionInput` aka [GraphQL Operation](https://www.graphql-java.com/documentation/execution#queries),
+whereas these custom instrumentations applies to multiple GraphQL operations (say a BatchRequest) and stores its state in the `GraphQLContext`.
```
dariuszkuc, line 11:
```suggestion
whereas these custom instrumentations apply to multiple GraphQL operations (say a BatchRequest) and store their state in the `GraphQLContext`.
```

website/docs/server/data-loader/data-loader-instrumentation.md
```diff
@@ -0,0 +1,189 @@
...
+allowing batching and deduplication of transactions across those multiple GraphQL operations.
+
+The `graphql-kotlin-dataloader-instrumentation` module contains 2 custom `DataLoader` instrumentations:
+
+1. **DataLoaderLevelDispatchedInstrumentation**
+
+   Dispatches all data loaders when a certain **Level** when a certain **Level**of all **ExecutionInputs** sharing a
```
dariuszkuc, line 18:
```suggestion
   Dispatches all data loaders when a certain **Level** of all **ExecutionInputs** sharing a
```

dariuszkuc (changes requested):

website/docs/server/data-loader/data-loader-instrumentation.md
```diff
@@ -0,0 +1,191 @@
...
+**Execution diagram**
+
+![Image of data loader level dispatched instrumentation](../../assets/data-loader-level-dispatched-instrumentation.png)
+
+* The `astronaut` `DataFetcher` uses a `AstronautDataLoader` which will be dispatched when **Level 1** of those 2 operations
+  is dispatched, causing the `AstronautDataLoader` to load 2 astronauts.
+* The `missions` `DataFetcher` uses a `MissionsByAstronautDataLoader` which will be dispatched when **Level 2** of those 2 operations
+  is dispatched, causing the `MissionsByAstronautDataLoader` to load 2 lists of missions by astronaut.
+
+### Usage on spring server
```
dariuszkuc, line 72: I think we should provide docs on integration outside of spring server (there could be a note below saying that our spring support simplifies the above setup and you can use this simple yaml configuration to enable it). Guess you don't need to specify the complete schema setup part -> main part is to highlight that this instrumentation requires population of GraphQL context map with the state (this comment applies to both instrumentations).

i.e. something along the lines

===
In order to enable batching by level, you need to configure your GraphQL instance with the `DataLoaderLevelDispatchedInstrumentation`

```kotlin
val graphQL = GraphQL
    .instrumentation(DataLoaderLevelDispatchedInstrumentation())
    // configure schema, type wiring, etc.
    .build()
```

This data loader instrumentation relies on a global state object that should be stored in the GraphQL context map

```kotlin
val graphQLContext = mapOf(
    KotlinDataLoaderRegistry::class to kotlinDataLoaderRegistry,
    SyncExecutionExhaustedState::class to ExecutionLevelDispatchedState(queries.size)
)
```

:::info
`graphql-kotlin-spring-server` provides convenient integration of batch loader functionality through simple configuration. Batching instrumentation can be enabled by configuring following properties

```yaml
graphql:
  batching:
   enabled: true
   strategy: LEVEL_DISPATCHED
```
:::

dariuszkuc (changes requested):

dariuszkuc (approved):

Issue #1426, opened 2022-04-20
GraphQLClientGenerator fails when KotlinPoet upgraded to latest stable version


bsorbo:
**Library Version**
At least from 5.2.0 on up.

**Describe the bug**
GraphQLClientGenerator, when using kotlin-poet to define data class constructor parameters, specifies `KModifier.OVERRIDE` on the addParameter call. In kotlin-poet v1.11.0, behavior was recently changed to disallow this, which causes the following error message when using that version:
```
Exception in thread "main" java.lang.IllegalArgumentException: Modifiers [OVERRIDE] are not allowed on Kotlin parameters. Allowed modifiers: [VARARG, NOINLINE, CROSSINLINE]
        at com.squareup.kotlinpoet.ParameterSpec.<init>(ParameterSpec.kt:41)
        at com.squareup.kotlinpoet.ParameterSpec.<init>(ParameterSpec.kt:29)
        at com.squareup.kotlinpoet.ParameterSpec$Builder.build(ParameterSpec.kt:157)
        at com.squareup.kotlinpoet.FunSpec$Builder.addParameter(FunSpec.kt:465)
        at com.expediagroup.graphql.plugin.client.generator.GraphQLClientGenerator.generate$graphql_kotlin_client_generator(GraphQLClientGenerator.kt:162)
        at com.expediagroup.graphql.plugin.client.generator.GraphQLClientGenerator.generate(GraphQLClientGenerator.kt:68)
        at com.expediagroup.graphql.plugin.client.GenerateClientKt.generateClient(generateClient.kt:47)
        at com.expediagroup.graphql.plugin.client.GenerateClientKt.generateClient$default(generateClient.kt:29)
        ...
```

This change in behavior in KotlinPoet is listed as a bug-fix in [the 1.11.0 release notes](https://github.com/square/kotlinpoet/releases/tag/1.11.0), and the maintainer confirmed this is the expected behavior from KotlinPoet in [this now-closed Github Issue](https://github.com/square/kotlinpoet/issues/1230), while also pointing out the fix.

**To Reproduce**
Change the `kotlinPoetVersion` in `gradle.properties` to `1.11.0`, then run `./gradlew check`. Various existing tests fail.

**Expected behavior**
I would expect tests to pass/schema generation to succeed when using the latest stable version of kotlin-poet.

bsorbo:
[...]

PR #1427, opened 2022-04-20
Upgrade to KotlinPoet v1.11.0, make corresponding required changes.

bsorbo:
### :pencil: Description
This PR updates the KotlinPoet dependency version to `v1.11.0`. There are two breaking changes [in this version](https://github.com/square/kotlinpoet/releases/tag/1.11.0) that affect graphql-kotlin:
- KotlinPoet now enforces only allowed parameter modifiers in ParameterSpec (i.e. crossinline, vararg, and noinline). This required an actual fix in `GraphQLClientGenerator`.
- KotlinPoet now emits trailing commas for multi-line parameters and annotations.

In most cases the second change wouldn't be breaking, but because graphql-kotlin's test suite does an exact equality match against golden master queries, these started failing. Luckily regex can catch all these cases fairly easily.

### :link: Related Issues
This resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1426


dariuszkuc:
Thanks for fixing this!

dariuszkuc:
@bsorbo all releases should be stable, we are using alpha tags to basically indicate that there still might be some breaking API changes coming up before the final release (e.g. looks like I might need to do a breaking change to support Apollo Federation V2 in the generator).

I don't think we can backport this to 5.x branch simply because this version of `kotlinpoet` relies on Kotlin 1.6 and we just started supporting it in v6.

bsorbo:
> @bsorbo all releases should be stable, we are using alpha tags to basically indicate that there still might be some breaking API changes coming up before the final release (e.g. looks like I might need to do a breaking change to support Apollo Federation V2 in the generator).
> 
> I don't think we can backport this to 5.x branch simply because this version of `kotlinpoet` relies on Kotlin 1.6 and we just started supporting it in v6.

Thanks, ok, makes sense.

PR #1428, opened 2022-04-21
Value type improvements

rocketraman:
### :pencil: Description

Value types can be represented in various ways:

1) Unwrapped, as the underlying type

2) Unwrapped, as a custom scalar type representing the value type

3) Wrapped, like a regular class

There are valid use cases for all 3, though in my opinion options number 2 and 3 are generally the best, as they maintain the extra type information provided by the value class in the schema.

Currently, the docs only discuss option 1, and option 2 is actually not possible.

This PR makes a minor change to the code to support option 2, and documents all 3 value class representations.

NOTE: While the code change works for me, I would like to add tests, but frankly I'm not sure of the best way to do so. Some guidance would be appreciated.

### :link: Related Issues

https://github.com/ExpediaGroup/graphql-kotlin/issues/1370
https://github.com/ExpediaGroup/graphql-kotlin/pull/1395

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
@@ -104,5 +104,7 @@ private fun mapToEnumValue(paramType: KType, enumValue: String): Enum<*> =
 
 private fun <T : Any> mapToInlineValueClass(value: Any?, targetClass: KClass<T>): T {
     val targetConstructor = targetClass.primaryConstructor ?: throw PrimaryConstructorNotFound(targetClass)
-    return targetConstructor.call(value)
+    // if the user has registered a coercer for the value class, it may already be an instance of the target type
+    @Suppress("UNCHECKED_CAST")
+    return if (targetClass.isInstance(value)) value as T else targetConstructor.call(value)
 }
--- a/website/docs/schema-generator/writing-schemas/scalars.md
+++ b/website/docs/schema-generator/writing-schemas/scalars.md
@@ -130,7 +130,9 @@ It is often beneficial to create a wrapper around the underlying primitive type 
 to optimize such use cases - Kotlin compiler will attempt to use underlying type directly whenever possible and only keep the wrapper classes
 whenever it is necessary.
 
-In order to use inline value classes in your schema, you need to register it using hooks and also provide value unboxer that will be used by
+#### Representing Unwrapped Value Classes in the Schema as the Underlying Type
+
+In order to represent unwrapped inline value classes in your schema as the underlying type, you need to register it using hooks and also provide value unboxer that will be used by
 `graphql-java` when dealing with its wrapper object.
 
 ```kotlin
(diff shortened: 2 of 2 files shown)


dariuszkuc:
Hello 👋 
Thanks for the updates to the docs. Just curious whats the use case for wrapping regular objects in inline value classes? i.e. due to the inlining properties it is probably most useful to wrap Kotlin primitives (which can potentially map to custom scalars, e.g. Long). If you have a complex object type why inline it vs just create another custom object type?

rocketraman:
> Just curious whats the use case for wrapping regular objects in inline value classes? i.e. due to the inlining properties it is probably most useful to wrap Kotlin primitives (which can potentially map to custom scalars, e.g. Long). If you have a complex object type why inline it vs just create another custom object type?

@dariuszkuc I didn't mean to suggest wrapping regular objects in value classes. I think the three options I am talking about are best explained with an example. Say I have a value type "Meters" that wraps a primitive `Int`:

```
value class Meters(val value: Int)
```

and say I am returning such in some arbitrary GraphQL response e.g.:

```
data class Trip(
  val travelDistance: Meters
)
```

Now, in each of the three options this would be represented like this in the schema:

1. Unwrapped, as the underlying type

```
type Trip {
  travelDistance: Int!
}
```

No preservation of type information: the frontend must read docs to know that `travelDistance` is in meters. The same problem we would have had in the backend model before the introduction of the value class!

2. Unwrapped, as a custom scalar type representing the value type

```
"Represents a number of Meters as an integer value."
scalar Meters

type Trip {
  travelDistance: Meters!
}
```

The frontend must now explicitly deal with a scalar type of `Meters`. The type information of the value class is preserved across the API boundary.

3. Wrapped, like a regular class

```
type Meters {
  value: Int!
}

type Trip {
  travelDistance: Meters!
}
```

The frontend must also here explicitly deal with a type of `Meters`. The type information of the value class is preserved across the API boundary.

Hopefully that clarifies my intent!

rocketraman:
@dariuszkuc Any follow-up on this?

dariuszkuc:
@rocketraman PR is still a draft, is it ready?

as for the use cases -> yes value classes could potentially be used for wrapping primitives classes but I am still unsure how the latter is useful for schema purposes (documentation showing that it is possible to do is fine), i.e. 
* with Kotlin wrapped primitives you can embed some additional information in your code but still rely on primitive performance optimizations
* similarly you can embed additional information in the graph when using custom scalars and if they are mapped to the Kotlin primitives then you get the same benefits as above - IMHO this is the best use case as you get additional semantic information on both server (from value class) and client (from schema)

What I am unsure is - what sort of benefit you would get from wrapping a class over providing a custom class, you wouldn't get perf optimizations from server nor you wouldn't expose the additional semantic info in the graph

rocketraman:
> @rocketraman PR is still a draft, is it ready?

I've marked it as Ready now.

> * with Kotlin wrapped primitives you can embed some additional information in your code but still rely on primitive performance optimizations

Right, that's the basic idea of value classes. Add type information, but without losing out on the perf advantages of primitives. In general, where we have a value class in the backend, ideally we would *always* have had a wrapper type there, but we chose not to for optimization reasons.

> similarly you can embed additional information in the graph when using custom scalars and if they are mapped to the Kotlin primitives then you get the same benefits as above - IMHO this is the best use case as you get additional semantic information on both server (from value class) and client (from schema)

Right, this is my use case number 2, which is actually not possible with the current code AFAIK. This PR makes it possible with the change to `convertArgumentValue.kt` (https://github.com/ExpediaGroup/graphql-kotlin/pull/1428/files#diff-8482e178e473e5cb8d38318d4ac00003d517224f3498d77c2522fcd625aad0de), and documents it.

> What I am unsure is - what sort of benefit you would get from wrapping a class over providing a custom class, you wouldn't get perf optimizations from server nor you wouldn't expose the additional semantic info in the graph

I'm likely not understanding your point here.

On the server, all 3 use cases wrap a primitive with a value class, so we always get that perf benefit on the server (assuming as a backend dev we've chosen the right use cases in which to use a value class).

The additional semantic info in the graph comes from use cases 2 and 3 which the current docs don't discuss at all, which is the whole point of this PR.

rocketraman:
@dariuszkuc Ping on this, thanks

samuelAndalon (approved):

Issue #1324, opened 2021-12-10
Avoid adding 'Input' in case of using INPUT_OBJECT annotation


adammach:
Hi,
is there any possibility to avoid adding 'Input' postfix to name of input objects? If object is already with INPUT_OBJECT annotation I think this is not necessarily. 

thanks

mervyn-mccreight:
Is this referring to schema generation? So you mean there should not be a post-fixed "Input" to class-names that are annotated with `INPUT_OBJECT` in the generated schema?

Or what is the desired behavior in this issue?
I'm asking because I'm thinking about contributing a PR :)

dariuszkuc:
Hello 👋 
@mervyn-mccreight yes. When object is annotated with `INPUT_OBJECT` as the only valid location, it is not possible to reuse it as an output type hence there is no need to explicitly add the suffix there (suffix is added by default to avoid collisions when same Kotlin type is used for both input and output types).

mervyn-mccreight:
Unfortunately I forgot to mention inside the PR description that it closes this issue, so one has to close it by hand I am afraid 😅

dariuszkuc:
No worries, thanks for the contribution!

bherrmann2:
What if an object needs to be changed to be both an input and output object? The same one now cannot be used.

dariuszkuc:
> What if an object needs to be changed to be both an input and output object? The same one now cannot be used.

While you can use same object in the code for both input and output type, GraphQL spec makes a distinction between input and output types and they have to have unique names. That is why if type is used for both input and output types, we append `Input` suffix to make a distinction between them. 

If you define `Foo` as input type in your schema, you won't be able to use the same name for output types - if you were to update locations to support both then your input type in the schema would change to `FooInput` which is a potentially breaking schema change.

bherrmann2:
Yes exactly. Is that something we're ok with? Before this change, you could change an object from input only to both input and output without any potential breaking change

samuelAndalon:
https://github.com/ExpediaGroup/graphql-kotlin/pull/1438

PR #1434, opened 2022-05-01
feat(generator): avoid adding `input`-suffix for input-only classes

mervyn-mccreight:
### :pencil: Description
Avoid adding the `Input`-suffix for classes restricted to being used as input objects only using the `@GraphQLValidObjectLocations`-annotation. Suffixing the type with `Input` is not necessary, since no name collision can happen here.

### :link: Related Issues
#1324

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
@@ -16,6 +16,7 @@
 
 package com.expediagroup.graphql.generator.internal.extensions
 
+import com.expediagroup.graphql.generator.annotations.GraphQLValidObjectLocations
 import com.expediagroup.graphql.generator.exceptions.CouldNotGetNameOfKClassException
 import com.expediagroup.graphql.generator.hooks.SchemaGeneratorHooks
 import com.expediagroup.graphql.generator.internal.filters.functionFilters
@@ -28,6 +29,7 @@ import kotlin.reflect.KProperty
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.declaredMemberFunctions
 import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.findParameterByName
 import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.full.memberFunctions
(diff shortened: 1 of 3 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
```diff
@@ -88,8 +89,19 @@ internal fun KClass<*>.getSimpleName(isInputClass: Boolean = false): String {
         ?: this.simpleName
         ?: throw CouldNotGetNameOfKClassException(this)
 
+    fun Annotation.isInputOnlyAnnotation(): Boolean = when (this) {
```
dariuszkuc, line 92: lets make it internal and move it to a top level function

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
```diff
@@ -88,8 +89,19 @@ internal fun KClass<*>.getSimpleName(isInputClass: Boolean = false): String {
...
+        is GraphQLValidObjectLocations -> this.locations.contains(GraphQLValidObjectLocations.Locations.INPUT_OBJECT)
```
dariuszkuc, line 93: this check is not correct -> locations is an array so this check would be true even if type can be applied for both input and output (we want to apply the suffix if it is applicable only on the input types)

dariuszkuc (changes requested):

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
mervyn-mccreight: I made it top-level, private and decided to make it an extension of `KClass` rather than `Annotation`.
What do you think?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/extensions/kClassExtensions.kt
mervyn-mccreight: Ah, of course. Nice catch, thanks 👍 
I added a test case for this and fixed it.

samuelAndalon (approved):

dariuszkuc (approved):

PR #1437, opened 2022-05-03
fix: willAddGraphQLTypeToSchema needs annotations from field

bherrmann2:
### :pencil: Description
When using the `@GraphQLUnion` as a meta annotation, the KType is `Any` when the type for the field is being generated by reflection. The correct type is actually the annotation on the field with the `@GraphQLUnion` meta annotation. In this case, the type passed to the hook needs to be that annotation since that is the class with other annotations on it, which could be needed by the hook.

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateGraphQLType.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateGraphQLType.kt
@@ -70,8 +70,12 @@ private fun objectFromReflection(generator: SchemaGenerator, type: KType, typeIn
          * For a field using the meta union annotation, the `type` is `Any`, but we need to pass the annotation with the meta union annotation as the type
          * since that is really the type generated from reflection and has any potential directives on it needed by the hook
          */
-        val metaUnion: Annotation? = typeInfo.fieldAnnotations.firstOrNull { it.getMetaUnionAnnotation() != null }
-        val resolvedType = if (kClass.isInstance(Any::class) && metaUnion != null) metaUnion.annotationClass.createType() else type
+        val metaUnion = typeInfo.fieldAnnotations.firstOrNull { it.getMetaUnionAnnotation() != null }
+        val resolvedType = if (kClass.isInstance(Any::class) && metaUnion != null) {
+            metaUnion.annotationClass.createType()
+        } else {
+            type
+        }
 
         generator.config.hooks.willAddGraphQLTypeToSchema(resolvedType, graphQLType)
     }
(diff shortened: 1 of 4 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateGraphQLType.kt
```diff
@@ -63,7 +65,15 @@ private fun objectFromReflection(generator: SchemaGenerator, type: KType, typeIn
...
     return generator.cache.buildIfNotUnderConstruction(kClass, typeInfo) {
         val graphQLType = getGraphQLType(generator, kClass, type, typeInfo)
-        generator.config.hooks.willAddGraphQLTypeToSchema(type, graphQLType)
+
+        /*
+         * For a field using the meta union annotation, the `type` is `Any`, but we need to pass the annotation with the meta union annotation as the type
+         * since that is really the type generated from reflection and has any potential directives on it needed by the hook
+         */
+        val metaUnion: Annotation? = typeInfo.fieldAnnotations.firstOrNull { it.getMetaUnionAnnotation() != null }
+        val resolvedType = if (kClass.isInstance(Any::class) && metaUnion != null) metaUnion.annotationClass.createType() else type
```
dariuszkuc, line 74: brackets please

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateGraphQLType.kt
```diff
@@ -63,7 +65,15 @@ private fun objectFromReflection(generator: SchemaGenerator, type: KType, typeIn
 
     return generator.cache.buildIfNotUnderConstruction(kClass, typeInfo) {
         val graphQLType = getGraphQLType(generator, kClass, type, typeInfo)
-        generator.config.hooks.willAddGraphQLTypeToSchema(type, graphQLType)
+
+        /*
+         * For a field using the meta union annotation, the `type` is `Any`, but we need to pass the annotation with the meta union annotation as the type
+         * since that is really the type generated from reflection and has any potential directives on it needed by the hook
+         */
+        val metaUnion: Annotation? = typeInfo.fieldAnnotations.firstOrNull { it.getMetaUnionAnnotation() != null }
```
samuelAndalon, line 73: We can just infer the type and not explicitly specify it

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateGraphQLType.kt
tapaderster: this should have been caught by our lint checker?

samuelAndalon (approved):

PR #1463, opened 2022-06-12
feat: allow graphql java version range until v18

samuelAndalon:
### :pencil: Description
allow minor releases of `graphql-java` from 17.2 until 18.0
[version declaration syntax](https://docs.gradle.org/current/userguide/single_versions.html)

### DIFF
--- a/generator/graphql-kotlin-schema-generator/build.gradle.kts
+++ b/generator/graphql-kotlin-schema-generator/build.gradle.kts
@@ -11,7 +11,7 @@ val slf4jVersion: String by project
 dependencies {
     api("com.graphql-java:graphql-java") {
         version {
-            strictly(graphQLJavaVersion)
+            strictly("[$graphQLJavaVersion, 18.0)")
         }
     }
     api("org.jetbrains.kotlinx:kotlinx-coroutines-reactive:$kotlinCoroutinesVersion")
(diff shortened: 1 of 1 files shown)


generator/graphql-kotlin-schema-generator/build.gradle.kts
```diff
@@ -11,7 +11,7 @@ val slf4jVersion: String by project
 dependencies {
     api("com.graphql-java:graphql-java") {
         version {
-            strictly(graphQLJavaVersion)
+            strictly("[$graphQLJavaVersion, 18.0)")
```
dariuszkuc, line 14: I think it should be 
```suggestion
            strictly("[$graphQLJavaVersion, 18.0[")
```

smyrick: I actually don't think we should use `[`

> An upper bound exclude acts as a prefix exclude. This means that `[1.0, 2.0[` will also exclude all versions starting with 2.0 that are smaller than 2.0. For example versions like 2.0-dev1 or 2.0-SNAPSHOT are no longer included in the range.

https://docs.gradle.org/current/userguide/single_versions.html

dariuszkuc (approved):

Issue #1429, opened 2022-04-22
Allow subgraph contact info during Schema generation


fletchgqc:
Apollo uses the @contact directive and extension to the schema object to show contact information for subgraphs, as described here: https://www.apollographql.com/docs/studio/federated-graphs/#subgraph-contact-info. I think it's a pretty handy thing.

Is there any way to include this data during schema generation? I saw in the doco that directives can be created, but I'm not sure how to modify the `schema` object.

dariuszkuc:
Hello 👋 
We are currently looking into supporting federation v2 spec and most likely will provide support for contracts at that time as well (or as a quick follow up).

PR #1466, opened 2022-06-14
[generator] add support for schema directives

dariuszkuc:
### :pencil: Description

Update `SchemaGenerator` to accept optional schema `TopLevelObject`. Schema object can be annotated with `@GraphQLDescription` and/or various `@GraphQLDirectives` to provide additional metadata about the schema. In order to avoid the confusion, schema `TopLevelObject` **cannot** contain any properties or functions.

Add new `Schema` marker interface to allow `spring-server` to automatically locate schema beans.

### :link: Related Issues

Partially resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1429 (`@contact` directive will be added in subsequent PR)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -72,26 +74,36 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) : Closeab
...
+                if (schemaObject != null) {
+                    directives.clear()
 
-        return config.hooks.willBuildSchema(builder).build()
+                    builder.description(schemaObject.kClass.getGraphQLDescription())
+                        .withSchemaAppliedDirectives(generateSchemaDirectives(this, schemaObject))
+                        .withSchemaDirectives(directives.values)
+                }
+            }
+            .let { builder ->
```
samuelAndalon, line 104: nip: should we move hook invocation to the `.also` scope function ? usually `let` is used to execute lambdas on non-null objects
also let returns lambda result, this suit more as a side effect, and we would be returning result of hook.

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/extensions/generatorExtensions.kt
```diff
@@ -25,6 +25,13 @@ import org.springframework.aop.support.AopUtils
  * the schema generator can use
  */
 internal fun List<Any>.toTopLevelObjects() = this.map {
+    it.toTopLevelObject()
+}
+
+/**
+ * Converts given object to a [TopLevelObject] wrapper.
+ */
+internal fun Any.toTopLevelObject() = this.let {
```
samuelAndalon, line 34: nip: would it be helpful to be explicit about the return type ? the name of the function pretty much says it, but specifying the return type of function wouldn't harm

samuelAndalon (approved):

PR #1481, opened 2022-07-04
sample: upgrade to ktor 2.0 and add end-to-end test

jmfayard:
### :pencil: Description

Hello I wanted to see how end-to-end test works with ktor,
but in order to do that I needed to upgrade the sample to ktor 2


jmfayard:
Closing, because I git messed up...

PR #1484, opened 2022-07-16
small letter case fix

nikitakoselev:
All other particles in headers are in small case, so "to" should be the same.
image

Also - smile, we created this contribution together with Eman Hammad, as part of our https://www.meetup.com/twosseng/
You can see the recording here:  https://tinyurl.com/bdednj3x

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# How To Contribute
+# How to Contribute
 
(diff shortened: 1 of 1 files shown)


tapaderster (approved):

EXPErdelfavero:
Topic title was in title case, as it should be under Expedia's style guide. If you're going to go with the style of the subheaders, which isn't our style guide but would at least be consistent, then "Contribute" should be lower case, too.

PR #1489, opened 2022-07-26
[federation] fix @key fed v2 definition

dariuszkuc:
### :pencil: Description

Missed default value for the `resolvable` argument. Updated definition

```kotlin
directive @key(fields: _FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
```

### :link: Related Issues

N/A


samuelAndalon (approved):

alexhancock:
@dariuszkuc I would like to see this one land / use it. Do you know what version it will be available in? Thank you for doing it!

PR #1491, opened 2022-07-28
Update to graphql-java 19.0

josephlbarnett:
FlowSubscriptionExecutionStrategy needs to be updated to
pass instrumentationState to instrumentations per
https://github.com/graphql-java/graphql-java/pull/2769

ValidationError string changed slightly so adjust assertions.


samuelAndalon:
Thanks for the PR, we might need to also update the the custom data loader instrumentations we  provide to stop using the deprecated methods and start using the `instrumentXXX` methods have been changed to take in InstrumentationState

dariuszkuc:
@samuelAndalon since data loader dep is explicitly specified, IMHO it can be updated in the subsequent PR

samuelAndalon (approved):

PR #1495, opened 2022-07-29
[build] update to latest dependencies

dariuszkuc:
### :pencil: Description

Update to the latest dependencies. 

Fixed duplicate publication from `graphql-kotlin-gradle-plugin` - this was a warning before but latest version of `com.gradle.plugin-publish` treats it as a failure

NOTE: I did not update project to Kotlin v1.7 as I think that as a library we should target older version to allow for largest adoption.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -99,7 +99,7 @@ subprojects {
             sourceCompatibility = JavaVersion.VERSION_1_8
         }
 
-        // published artifacts
+        // published sources and javadoc artifacts
         val jarComponent = currentProject.components.getByName("java")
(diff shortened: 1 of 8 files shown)


dariuszkuc:
Looks like some env issue with actions -> android unit test is triggered based on the existence of `ANDROID_SDK_ROOT` env value but the test fails saying that it cannot find it

```
SDK location not found. Define location with an ANDROID_SDK_ROOT environment variable or by setting the sdk.dir path in your project's local properties file at...
```

samuelAndalon (approved):

Issue #1510, opened 2022-08-08
directive @contact not generated leading to invalid schema when @ContactDirective is used


chriskn:
**Library Version**
6.1.0 with spring-boot 2.7.2 

application.yml

```
graphql:
  packages: 'my.package'
  federation:
    enabled: true
    tracing:
      enabled: true
    opt-in-v2: true
```

**Describe the bug**
When using the @ContactDirective annotation like this

```
@ContactDirective(
    name = "My Team",
    url = "foo",
    description = "bar"
)
@Component
class MySchema : Schema
```

I get the @contact generated for the schema, which is fine

```
schema @contact(description : "bar", name : "My Team", url : "foo"){
  query: Query
}
```

but what's **missing** is the definition of the @contact directive:

```
directive @contact(
    "Contact title of the subgraph owner"
    name: String!
    "URL where the subgraph's owner can be reached"
    url: String
    "Other relevant notes can be included here; supports markdown links"
    description: String
) on SCHEMA
```

**To Reproduce**
Use the annotation and generate sdl

**Expected behavior**
`directive @contact` is generated and part of sdl

PR #1511, opened 2022-08-08
feat: use additionalDirectives for schema directives

samuelAndalon:
### :pencil: Description
use `additionalDirectives` instead of `withSchemaDirectives`, this method should be marked as depreacted, will open a PR in graphql-java

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1510


dariuszkuc (changes requested):
Can we add a test to verify the behavior?

dariuszkuc (approved):

tapaderster:
is it because we are not adding the `Contact` directive here:
https://github.com/ExpediaGroup/graphql-kotlin/blob/master/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/FederatedSchemaGeneratorHooks.kt#L92

dariuszkuc:
@tapaderster no, its because of broken logic in `graphql-java` that does not add the schema directives to the schema, we changed logic to explicitly add those to "additional" directives as a workaround.

Issue #1506, opened 2022-08-03
Using a DataLoader in a FederatedTypeResolver hangs execution 


roookeee:
**Library Version**
6.1.0

**Describe the bug**
When loading data via a `DataLoader` through the `DataFetchingEnvironment` in a `FederatedTypeResolver`, the execution hangs and will never finish. This is probalby related to #986 as `FederatedTypeResolver.resolve` is a `suspend fun` which calls into the `DataLoader` which creates nested asynchronicity: `suspend fun -> CompletableFuture<T>` of the `DataLoader`. This probably breaks the automatic dispatch as manually calling `dispatch` after the `List<CompletableFuture<T>>` has been created by calling the `DataLoader` removes the issue.

**To Reproduce**
- Create a `FederatedTypeResolver`
- Use a `DataLoader` via the given `DataFetchingEnvironment`
- `.await()` on the `CompletableFuture<T>` list created by calling into the `DataLoader` in `FederatedTypeResolver.resolve`

**Expected behavior**
`DataLoaders` should be usable in a `FederatedTypeResolver`

samuelAndalon:
Hello 👋 , as mentioned on our docs

```
Given that graphql-java relies on CompletableFutures for scheduling and asynchronous execution of DataLoader calls, currently we don't provide any native support for DataLoader pattern using coroutines. Instead, return the CompletableFuture directly from your DataLoaders. 
```

if you want to use DataLoaders you need to return a `CompletableFuture<*>` and stop marking the method as `suspend` as the `graphql-java` engine only supports it for async computations.

so for example  in your data fetcher you need to go from this

```kt
suspend fun user(id: Int): User? =
        userService.getUser(id)
```

to this: 

```kt
fun user(id: Int, environment: DataFetchingEnvironment): CompletableFuture<User?> =
  environment
    .getDataLoader<Int, User?>("UserDataLoader")
    .load(id)
```

and in your DataLoader do something like this:  

```kt
DataLoaderFactory.newDataLoader { requests, environment ->
  val coroutineScope =
    environment.getLoadContext<DataFetchingEnvironment>()?.graphQlContext?.get<CoroutineScope>()
    ?: CoroutineScope(EmptyCoroutineContext)

    coroutineScope.async {
      productService.getProducts(requests)
    }.asCompletableFuture()
}
```
[...]

PR #1514, opened 2022-08-12
feat: CompletableFuture in FederatedTypeResolver, standardize entities resolver with other subgraph implementations

samuelAndalon:
### :pencil: Description
With the idea of proving support for returning a `CompletableFuture` in `FederatedTypeResolver` 
_**and match other subgraphs implementations**_ a new sealed interface is added 

```kotlin
sealed interface FederatedTypeResolver {
    val typeName: String
}
```

this will be the basic interface of 2 interfaces

1. The currently supported `FederatedTypeResolver` is renamed to `FederatedTypeSuspendResolver`, with  support of suspendable function to be executed in a coroutine
```kotlin
interface FederatedTypeSuspendResolver<out T> : TypeResolver {
    override val typeName: String
    suspend fun resolve(
      environment: DataFetchingEnvironment, 
      representation: Map<String, Any>
    ): T?
}
```

2. Added a new interface called `FederatedTypePromiseResolver` that will provide support to return a `CompletableFuture` allowing compatibility with a `DataLoader`
```kotlin
interface FederatedTypePromiseResolver<T> : TypeResolver {
    fun resolve(
      environment: DataFetchingEnvironment, 
      representation: Map<String, Any>
    ): CompletableFuture<T?>
}
```

Clients will have the freedom to decided which interface to use for their resolvers and `graphql-kotlin` will be smart enough to decide how to execute them.

The execution logic will be capable of executing suspend resolvers or promise resolvers concurrently, specially useful if you have support for data loaders for only certain entities

example representations for multiple types, request to `_entities` field

`UserResolver` implementing `FederatedTypeResolver`
`PromiseResolver` implementing `FederatedTypePromiseResolver`

```json
[
  { "__typename": "User", "userId": 123 },
  { "__typename": "User", "userId": 456 },
  { "__typename": "Author", "authorId": 1 },
  { "__typename": "Author", "authorId": 2 },
]
```
This is going to be a breaking change as it changes the signature of the current `FederatedTypeResolver` by simplifying the logic, before it was providing a list of representations, and now it will provide a representation (internally graphql-kotlin will ensure the ordering).

All this with the effort of matching other subgraph implementations like Apollo Server
https://www.apollographql.com/docs/federation/api/apollo-subgraph/#__resolvereference

graphql-kotlin
```kotlin
object : FederatedTypeSuspendResolver<User> {
  override val typeName: String = "User"
    suspend fun resolve(
      environment: DataFetchingEnvironment, 
      representation: Map<String, Any>
    ): User? {
      // representation will always have at least the `id` and the `__typename` here
    }
}
object : FederatedTypePromiseResolver<User> {
  override val typeName: String = "User"
    fun resolve(
      environment: DataFetchingEnvironment, 
      representation: Map<String, Any>
    ): CompletableFuture<User?> {
      // representation will always have at least the `id` and the `__typename` here
    }
}
```

apollo-server
```ts
const resolvers = {
  User: {
    __resolveReference(representation, env) {
      // representation will always have at least the `id` and the `__typename` here
    }
  }
};
```

A couple of tests simulating async operations and delays were added to ensure ordering is properly maintained

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1506


generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/data/queries/federated/Product.kt
```diff
@@ -124,6 +124,19 @@ data class User(
     @ExternalDirective val name: String
 )
 
+/*
+type Author {
```
dariuszkuc, line 128: based on the code the type is actually

```graphql
type Author @extends @key(fields: "authorId") {
  authorId: Int! @external
  name: String: @external
```

If both fields are external -> should it be `@key(fields: "authorId", resolvable: false)`? or should there be at least a single local field?

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/EntitiesResultsExtensions.kt
```diff
@@ -0,0 +1,39 @@
...
+internal fun Sequence<Any?>.toDataFetcherResult(): DataFetcherResult<List<Any?>> {
+    val data = mutableListOf<Any?>()
+    val errors = mutableListOf<GraphQLError>()
+
+    forEach { result ->
+        if (result is GraphQLError) {
+            data += null
+            errors += result
+        } else {
+            data += result
```
dariuszkuc, line 31: should else also keep track of errors in the list?

dariuszkuc:
Related https://github.com/ExpediaGroup/graphql-kotlin/issues/1506

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/execution/EntitiesDataFetcher.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        val noResolverErrors: CompletableFuture<List<Map<Int, Any?>>> = CompletableFuture.completedFuture(
+            listOf(
+                representationsWithoutResolver.map(IndexedValue<Map<String, Any>>::index).zip(
+                    representationsWithoutResolver.map(IndexedValue<Map<String, Any>>::value).map { representation ->
+                        InvalidFederatedRequest("Unable to resolve federated type, representation=$representation")
+                    }
+                ).toMap()
+            )
+        )
```
smyrick, line 86: I think you can solve this with [associateWith](https://kotlinlang.org/docs/collection-transformations.html#associate)??

If not maybe a variable or two in this block would make it seem a little less magic or deep kotlin knowledge

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/execution/EntitiesDataFetcher.kt
```diff
@@ -0,0 +1,107 @@
...
+import graphql.schema.DataFetchingEnvironment
+import java.util.concurrent.CompletableFuture
+
+private const val TYPENAME_FIELD = "__typename"
+private const val REPRESENTATIONS = "representations"
+
+/**
+ * Federated _entities field data fetcher.
+ */
+open class EntitiesDataFetcher(
```
smyrick, line 35: This would be a breaking change so we would need another major. Do you need to change the name? It looks like the type if the class is the same

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/execution/EntitiesDataFetcher.kt
samuelAndalon:
```kotlin
representationsWithoutResolver.associateBy(IndexedValue<Map<String, Any>>::index) { (_, representation) ->
                    InvalidFederatedRequest("Unable to resolve federated type, representation=$representation")
                }
```

samuelAndalon:
This is going to be a breaking change as it changes the signature of the current `FederatedTypeResolver` by simplifying the logic, before it was providing a list of representations, and now it will provide a representation (internally graphql-kotlin will ensure the ordering).

All this with the effort of matching other subgraph implementations like Apollo Server
https://www.apollographql.com/docs/federation/api/apollo-subgraph/#__resolvereference

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/execution/EntitiesDataFetcher.kt
samuelAndalon: i think is worth it, we are pursuing a signature similar to other subgraph implementations

samuelAndalon:
a couple of tests simulating async operations and delays were added to ensure ordering is properly mantained

tapaderster (approved):

Issue #1515, opened 2022-08-12
pojo support for query / mutation parameter


cryptoki:
up to version 5.x we could use our Java Pojos

```
public class SomeOldPojo {
  public String value;
}
```

```
  fun pojo(pojo: SomeOldPojo): String? = "hello ${pojo.value}"
```

```
query {
	pojo(pojo: {value: "abc"})
}
```

with the 6.x release the Jackson ObjectMapper is removed and therewith the java support. Is there a best practice to use / integrate old pojo?

little playground https://github.com/cryptoki/graphql-defaults

dariuszkuc:
Hello 👋 
While Jackson is no longer used for object serialization/deserialization, POJOs should still be handled in the same way*.

*by using Java POJOs you are missing nullability information

cryptoki:
@dariuszkuc thank you. what do you mean with `in the same way`?

If I'm using Java POJOs in the same way as I used it with `graphql-kotlin 5.x.x` I've got the following error
```
com.expediagroup.graphql.generator.exceptions.PrimaryConstructorNotFound: Invalid input object SomeOldPojo - missing public primary constructor
```

v5.x.x works well with the inside Jackson mapper.

cryptoki:
add on info. 
we can't migration these Java POJOs to Kotlin. It's important to use the same java classes. We have a lot of services with java/groovy code using these POJOs.

dariuszkuc:
@cryptoki correction, yes it is not in the "same" way as instead of reflections and Jackson we are now using constructor to pass in the parameters. Your exception states your POJO classes do not specify the appropriate primary constructor.

cryptoki:
@dariuszkuc java classes does not have a `primary constructor`. 

how can we integrate / using our existing java pojo classes? It's important. see my little playground project linked in the top post.

```
query {
  pojo(pojo: {value: "abc"})
}
```

dariuszkuc:
Looks like `primary` constructor designation is from Kotlin, so Java classes won't provide this metadata. We could potentially update our logic from

```kotlin
var targetConstructor = targetClass.primaryConstructor ?: throw PrimaryConstructorNotFound(targetClass)
```

to something like
```kotlin
var targetConstructor = targetClass.primaryConstructor
if (targetConstructor == null && targetClass.constructors.size == 1) {
  targetConstructor = targetClass.constructors.first()
}
if (targetConstructor == null) {
[...]

PR #1518, opened 2022-08-15
feat: support argument types without primaryConstructor

samuelAndalon:
### :pencil: Description
there are some use cases where no primary constructor will be defined, like in java classes, for those cases if no primaryConstructor is found use the first element on the constructors list, otherwise throw the error

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1515

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
@@ -16,6 +16,7 @@
 
 package com.expediagroup.graphql.generator.execution
 
+import com.expediagroup.graphql.generator.exceptions.MultipleConstructorsFound
 import com.expediagroup.graphql.generator.exceptions.PrimaryConstructorNotFound
 import com.expediagroup.graphql.generator.internal.extensions.getGraphQLName
 import com.expediagroup.graphql.generator.internal.extensions.getKClass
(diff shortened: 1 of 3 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
```diff
@@ -93,7 +93,10 @@ private fun convertValue(
  * the only thing left to parse is object maps into the nested Kotlin classes
  */
 private fun <T : Any> mapToKotlinObject(input: Map<String, *>, targetClass: KClass<T>): T {
-    val targetConstructor = targetClass.primaryConstructor ?: throw PrimaryConstructorNotFound(targetClass)
+    val targetConstructor = targetClass.primaryConstructor
+        ?: targetClass.constructors.firstOrNull()
```
dariuszkuc, line 97: I think we should throw an error if there are multiple constructors.

dariuszkuc:
Theoretically we could use reflection and attempt to match input map against properties on the object (this is what jackson does) but unsure if we should do it.

samuelAndalon:
i think for ppl using classes without primary constructor, the effort of adding a constructor would be minimal, if we start adding support for matching properties without a constructor we would need to check several use cases ?

dariuszkuc:
Yes, matching properties is non trivial - you would need to most likely account for setters and follow Java Beans naming conventions.

It would allow folks to use Java objects but we never targeted that compatibility so I am unsure whether it is worth the effort.

dariuszkuc (approved):

Issue #1527, opened 2022-08-23
schema generator should fail if input types with nullable fields do not define a default value


samuelAndalon:
### Discussed in https://github.com/ExpediaGroup/graphql-kotlin/discussions/1525

<div type='discussions-op-text'>

<sup>Originally posted by **Anthonyzou** August 22, 2022</sup>
With the changes in version 6.2.0 and above, input types that could be nullable did not require a default value.

This code would work before version 6.2.0

```kotlin
@GraphQLDescription("")
data class Book(
    val title: String?
)
```

Now I get the error `No argument provided for a required parameter: parameter #0...`  unless I change the input type to:

```kotlin
@GraphQLDescription("")
data class Book(
    val title: String? = null
)
```</div>

**Comment by dariuszkuc** [comment]
I don't think this can be fixed without being able to access default values (https://github.com/ExpediaGroup/graphql-kotlin/issues/53). There is no way to programmatically decide whether default value is provided or not (without actually calling the function).

**Comment by xenomachina** [comment]
> There is no way to programmatically decide whether default value is provided or not (without actually calling the function).

You can use [KParameter.isOptional](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-parameter/is-optional.html).

**Comment by xenomachina** [comment]
As I mentioned in the linked discussion, I have already implemented these checks in our `SchemaGeneratorHooks`. The check is something like:

``` kotlin
if (inputParam.type.isMarkedNullable) {
    require(inputParam.isOptional) {
        "$inputClassName.${inputParam.name} is nullable but not optional"
    }
}
```

samuelAndalon:
we can add more custom logic here
https://github.com/ExpediaGroup/graphql-kotlin/blob/master/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/hooks/SchemaGeneratorHooks.kt#L144

adding this logic when unwrapped is `GraphQLInputObjectType`

```kotlin
type.jvmErasure.primaryConstructor?.parameters?.forEach { param ->
  if (param.type.isMarkedNullable && !param.isOptional) {
   // just for demo, we can use a custom exception
    throw RuntimeException("${type.jvmErasure.simpleName}.${param.name} is nullable but not optional")
  }
}
```

xenomachina:
> @xenomachina i would assume you did it in the `didGenerateGraphQLType` hook ?

[...]

PR #1528, opened 2022-08-24
Make nullable parameters always optional

xenomachina:
### :pencil: Description

Treat nullable parameters that have no default value as if their default was set to null.

Nullable parameters that have a default value explicitly set will continue to use that as their default.

### :link: Related Issues

#1527

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
@@ -109,7 +109,11 @@ private fun <T : Any> mapToKotlinObject(input: Map<String, *>, targetClass: KCla
     // filter parameters that are actually in the input in order to rely on parameters default values
     // in target constructor
     val constructorParametersInInput = constructorParameters.filter { parameter ->
-        input.containsKey(parameter.getName()) || parameter.type.isOptionalInputType()
+        input.containsKey(parameter.getName()) ||
+            parameter.type.isOptionalInputType() ||
+
+            // for nullable parameters that have no explicit default, we pass in null if not in input
+            (parameter.type.isMarkedNullable && !parameter.isOptional)
     }
     val constructorArguments = constructorParametersInInput.associateWith { parameter ->
         convertArgumentValue(parameter.getName(), parameter, input)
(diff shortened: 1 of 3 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
```diff
@@ -109,7 +109,11 @@ private fun <T : Any> mapToKotlinObject(input: Map<String, *>, targetClass: KCla
     // filter parameters that are actually in the input in order to rely on parameters default values
     // in target constructor
     val constructorParametersInInput = constructorParameters.filter { parameter ->
-        input.containsKey(parameter.getName()) || parameter.type.isOptionalInputType()
+        input.containsKey(parameter.getName()) ||
+            parameter.type.isOptionalInputType() ||
+
+            // for nullable parameters that have no explicit default, we pass in null if not in input
+            (parameter.type.isMarkedNullable && !parameter.isOptional)
```
samuelAndalon, line 116: probably worth to wrap this expression in an extension function in `kParameterExtensions`, 
`KParameter.isNotOptionalNullable()`

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
```diff
@@ -109,7 +109,11 @@ private fun <T : Any> mapToKotlinObject(input: Map<String, *>, targetClass: KCla
     // filter parameters that are actually in the input in order to rely on parameters default values
     // in target constructor
     val constructorParametersInInput = constructorParameters.filter { parameter ->
```
samuelAndalon, line 111: we will probably need to rename this
we can just call it `constructorParameters`, and do

```kotlin
val constructorParameters = targetConstructor.parameters.filter { ... }
```

samuelAndalon:
Thanks for the PR, i left couple of minor comments

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
xenomachina: Done.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
xenomachina: Done

samuelAndalon (approved):

Issue #1118, opened 2021-04-19
Inconsistent order of kotlin prop declaration and scheme prop order


heyvvwg:
**Library Version**
4.0.0-rc.1

**Describe the bug**
Inconsistent order of kotlin prop declaration and scheme prop order

**To Reproduce**
[Data Class]
@GraphQLName("BbsMdBoard")
data class BbsMdBoardType(
    override val id: ID,
    val name: String,
    val dtCreate: String
) : GqlNode

[Scheme]
type BbsMdBoard implements Node {
  dtCreate: String!
  id: ID!
  name: String!
}

**Expected behavior**
[Scheme]
type BbsMdBoard implements Node {
  id: ID!
  name: String!
  dtCreate: String!
}

dariuszkuc:
I believe the schema printer prints the properties in alphabetical order to ensure consistent order. Can you elaborate why this would be an issue?

icon1123:
I think the order in which it is declared has meaning. For example, these are the following cases:

[declared] 
id,         (id, i want first)
width,   (size)
height,  (size)
left,       (position)
top,       (position)
dtCreate

[alphabet]
dtCreate
height
id
left
top
width

icon1123:
> I believe the schema printer prints the properties in alphabetical order to ensure consistent order. Can you elaborate why this would be an issue?

I think the order in which it is declared has meaning. For example, these are the following cases:
[declared] 
id,         (id, i want first)
width,   (size)
height,  (size)
left,       (position)
top,       (position)
dtCreate

[alphabet]
dtCreate
height
id
left
top
width

smyrick:
We also don't control the schema printer and I wouldn't want to change what we get from graphql-java [SchemaPrinter](https://javadoc.io/doc/com.graphql-java/graphql-java/16.2/graphql/schema/idl/SchemaPrinter.html)

While I also agree that GraphQL services should not be concerned with field order, if you wanted to start a discussion, I would open one on the [graphql-java repo](https://github.com/graphql-java/graphql-java)

Issue #689, opened 2020-05-04
Properties sorting on generated schem


maespinoza:
It looks like the current implementation reads kotlin files and creates the corresponding properties in the graphql schema correctly. However, the order of the properties is in alphabetical order and not in the order of the defined kotlin files. If we then want to generate code from the generated SDL (lets say clients that want to use your service) it will be also in alphabetical order which might not be the best way from the business perspective. Is it possible to configure the way the schema is generated? Currently we use the toSchema function and then call .print() is there another way to generate the schema without passing through GraphQLSchema object so that we can preserve the order on the original kotlin files?

smyrick:
A GraphQL schema is not sorted in anyway and exposing this order doesn't seem like something that truly provides business value, but is instead leaking implementation details. If there is some reason for an order of fields you should probably return something in a list of objects so you can order them server side or use directives or descriptions to mark any additional metadata about the schema.

smyrick:
However, the schema generator just returns a `GraphQLSchema` object so once you have that you could create a custom implementation of the [SchemaPrinter](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/idl/SchemaPrinter.java) and print the output in any way you want.

smyrick:
I am going to close this issue but feel free to continue the discussion here or in our Slack channel: https://github.com/ExpediaGroup/graphql-kotlin#-contact

PR #1530, opened 2022-08-27
[server] add injected SchemaPrinter to allow SDL endpoint customization

jdsatlin-exp:
✅ [Contributing.md](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/CONTRIBUTING.md) has followed for this change.

### :pencil: Description
This change adds in an automatically configured SchemaPrinter bean to enable users of the spring server to provide their own SchemaPrinter (either one configured with different options, or an overriding implementation) to be used with the SDL endpoint.

I encountered this issue #689 while seeking to do the same thing (change the sort order of properties in the generated schema). The suggested resolution there was:

>However, the schema generator just returns a GraphQLSchema object so once you have that you could create a custom implementation of the [SchemaPrinter](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/idl/SchemaPrinter.java) and print the output in any way you want.

However in a slight variation of the problem, we run integration tests checking our expected schema against our SDL endpoint to guard against unexpected schema changes. Previous to these changes the SDL endpoint just relies on a hardcoded default Schema Printer implementation and set of options. Which means even a custom schema printing implementation does not get reflected in the SDL endpoint

I believe the only documentation changes that would be required would be on the [spring beans page](https://opensource.expediagroup.com/graphql-kotlin/docs/server/spring-server/spring-beans) to indicate the new default bean and its interaction with the SDL endpoint.

I also considered related changes in the graphql-kotlin-sdl-generator's [GenerateSDL.kt](https://github.com/ExpediaGroup/graphql-kotlin/blob/095b056b74d52237b73cef872fdbd851c019039c/plugins/schema/graphql-kotlin-sdl-generator/src/main/kotlin/com/expediagroup/graphql/plugin/schema/GenerateSDL.kt#L107) but did not see as natural of extension point, as [SchemaGeneratorHooks](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/hooks/SchemaGeneratorHooks.kt) did not necessarily feel like a completely natural extension point. I'd definitely be open to hearing any suggestions for how to contribute that corresponding change.

### :link: Related Issues
In addition to the above mentioned #689, I believe this change would also fulfill part but not all of the ask in #1118.

### DIFF
--- a/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
+++ b/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
@@ -39,6 +39,7 @@ import graphql.execution.instrumentation.ChainedInstrumentation
 import graphql.execution.instrumentation.Instrumentation
 import graphql.execution.preparsed.PreparsedDocumentProvider
 import graphql.schema.GraphQLSchema
+import graphql.schema.idl.SchemaPrinter
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.annotation.Bean
@@ -159,4 +160,18 @@ class GraphQLSchemaConfiguration {
         contextFactory,
         requestHandler
     )
+
+    @Bean
+    @ConditionalOnMissingBean
+    fun schemaPrinter(): SchemaPrinter {
+        return SchemaPrinter(
+            SchemaPrinter.Options.defaultOptions()
+                .includeIntrospectionTypes(false)
+                .includeScalarTypes(true)
+                .includeSchemaDefinition(true)
+                .includeDirectives(true)
+                .includeDirectives { true }
+                .includeDirectiveDefinitions(true)
+        )
+    }
 }
--- a/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
+++ b/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
@@ -163,7 +163,16 @@ class GraphQLSchemaConfiguration {
 
     @Bean
     @ConditionalOnMissingBean
-    fun schemaPrinter(): SchemaPrinter {
+    fun schemaPrinter(): SchemaPrinter  = 
+      SchemaPrinter(
+            SchemaPrinter.Options.defaultOptions()
+                .includeIntrospectionTypes(false)
+                .includeScalarTypes(true)
+                .includeSchemaDefinition(true)
+                .includeDirectives(true)
+                .includeDirectives { true }
+                .includeDirectiveDefinitions(true)
+        )
         return SchemaPrinter(
             SchemaPrinter.Options.defaultOptions()
                 .includeIntrospectionTypes(false)
--- a/servers/graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/server/spring/SchemaConfigurationTest.kt
+++ b/servers/graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/server/spring/SchemaConfigurationTest.kt
@@ -36,6 +36,7 @@ import graphql.execution.instrumentation.tracing.TracingInstrumentation
 import graphql.schema.DataFetchingEnvironment
 import graphql.schema.GraphQLSchema
 import graphql.schema.GraphQLTypeUtil
+import graphql.schema.idl.SchemaPrinter
 import io.mockk.mockk
 import org.assertj.core.api.AssertionsForInterfaceTypes.assertThat
 import org.dataloader.DataLoader
@@ -77,6 +78,8 @@ class SchemaConfigurationTest {
                 assertNotNull(helloWorldQuery)
                 assertEquals("String", GraphQLTypeUtil.unwrapAll(helloWorldQuery.type).name)
 
+                assertThat(ctx).hasSingleBean(SchemaPrinter::class.java)
+
                 assertThat(ctx).hasSingleBean(GraphQL::class.java)
                 val graphQL = ctx.getBean(GraphQL::class.java)
                 val result = graphQL.execute("query { hello }").toSpecification()
@@ -111,6 +114,10 @@ class SchemaConfigurationTest {
                 assertThat(ctx).hasSingleBean(GraphQLSchema::class.java)
                 assertThat(ctx).getBean(GraphQLSchema::class.java)
                     .isSameAs(customConfiguration.mySchema())
+
+                assertThat(ctx).hasSingleBean(SchemaPrinter::class.java)
+                assertThat(ctx).getBean(SchemaPrinter::class.java)
+                    .isSameAs(customConfiguration.myCustomSchemaPrinter())
 
                 assertThat(ctx).hasSingleBean(GraphQL::class.java)
                 assertThat(ctx).getBean(GraphQL::class.java)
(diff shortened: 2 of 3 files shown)


servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -159,4 +160,18 @@ class GraphQLSchemaConfiguration {
...
+    fun schemaPrinter(): SchemaPrinter {
+        return SchemaPrinter(
+            SchemaPrinter.Options.defaultOptions()
+                .includeIntrospectionTypes(false)
+                .includeScalarTypes(true)
+                .includeSchemaDefinition(true)
+                .includeDirectives(true)
+                .includeDirectives { true }
+                .includeDirectiveDefinitions(true)
+        )
```
jdsatlin-exp, line 175: The values here all match the defaults from [GraphQLSchema.print()](https://github.com/ExpediaGroup/graphql-kotlin/blob/095b056b74d52237b73cef872fdbd851c019039c/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/extensions/GraphQLSchemaExtensions.kt#L34-L51) 

I'd be open to suggestions for a good place to codify the shared set of defaults, and use that instead of defining a duplicate set of values here.

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -159,4 +160,18 @@ class GraphQLSchemaConfiguration {
         contextFactory,
         requestHandler
     )
+
+    @Bean
```
jdsatlin-exp, line 164: Due to the SchemaPrinter's relationship with the Schema I thought this was the most natural place to put this bean, especially given the `SdlRouteConfiguration` also uses the GraphQLSchema, relies on both being injected there. Open to hearing if there's another place the maintainers think is more natural, including either another existing class or a class of its own.

servers/graphql-kotlin-spring-server/src/test/kotlin/com/expediagroup/graphql/server/spring/SchemaConfigurationTest.kt
```diff
@@ -173,6 +180,9 @@ class SchemaConfigurationTest {
 
         @Bean
         fun myDataLoaderRegistryFactory(): KotlinDataLoaderRegistryFactory = mockk()
+
+        @Bean
+        fun myCustomSchemaPrinter(): SchemaPrinter = mockk(relaxed = true)
```
jdsatlin-exp, line 185: This requires a relaxed mock because the `SdlRouteConfiguration` calls this class in its constructor, so a strict mock throws a bean instantiation error before the relevant tests here can even begin.

jdsatlin-exp (commented):
Notes for the reviewer

samuelAndalon:
Im not totally oppose against the idea of this PR, but by allowing having a custom printer we might open the door for issues when trying to consume the SDL from an external system, let's say a federated gateway.

I would be very careful about that and explicitly mention this in the documentation, having said that -- can you update the documentation to mention this, along with the information about how to provide a custom SchemaPrinter ?

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt
```diff
@@ -159,4 +160,18 @@ class GraphQLSchemaConfiguration {
...
+    @ConditionalOnMissingBean
+    fun schemaPrinter(): SchemaPrinter {
```
samuelAndalon, line 166:
```suggestion
    fun schemaPrinter(): SchemaPrinter  = 
      SchemaPrinter(
            SchemaPrinter.Options.defaultOptions()
                .includeIntrospectionTypes(false)
                .includeScalarTypes(true)
                .includeSchemaDefinition(true)
                .includeDirectives(true)
                .includeDirectives { true }
                .includeDirectiveDefinitions(true)
        )
```

samuelAndalon: use function expression body

samuelAndalon (changes requested):
update documentation

dariuszkuc:
SDL endpoint is provided for convenience only and AFAIK introspection should be generally used instead. As long as type definitions are the same then definition order in SDL shouldn't matter*. IMHO this change is unnecessary.

*I believe only recently `graphql-java` prints schema out in a consistent order

jdsatlin-exp:
> SDL endpoint is provided for convenience only and AFAIK introspection should be generally used instead. As long as type definitions are the same then definition order in SDL shouldn't matter*. IMHO this change is unnecessary.
> 
> *I believe only recently `graphql-java` prints schema out in a consistent order

I guess I didn't make the full case here, sorry about that. In my teams recent discussions we've seen two places where this might be a change that would enable valuable workloads on our side. The first one was what I stated above integration tests checking expected schema vs. the SDL endpoint (if there's a reasonably easy way to do that using introspection I'd be happy to hear any suggestions for it). Which would also allow us to order those schemas with a focus on readability for creating documentation automatically (or with a minimum of work). GraphQL Java added support for sorting on SchemaPrinters in https://github.com/graphql-java/graphql-java/pull/1546 via configuration to support similar cases to this, which makes this seem like a relevant spot to delegate configuration to clients while preserving current defaults. 

In terms of this being a Server SDL endpoint only change, I'd definitely love any suggestions for how to implement this in graphql-kotlin-sdl-generator as well, as that seems like a natural companion to me. I'd have no trouble contributing that as a companion change (or updating this PR to include that work as well). As mentioned above, I just wasn't very confident about adding the SchemaPrinter as a hook.

The other use case that has come up is keeping some non-production-ready portions of the graph private but allowing active incremental development on them, by not generating schema for 'experimental' features before the subgraph is fed into the federated supergraph. This seems preferable to disabling introspection on the federated server as a method of security through security. This change would also benefit from graphql-kotlin-sdl-generator I mentioned, but the SDL endpoint was the easier first reach into contributing here.

Thanks for taking the time to look this over and talk it though with me.

dariuszkuc:
IMHO doing custom printer will add tons of complexity (to your workflow) and is generally asking for a trouble. I'd advise against it. Changing printing order is one thing but with custom printer you can also filter out types etc. Introspection is generally* used as a source of truth by various tooling, so you may end up with inconsistent schema between SDL and introspection.

*sadly introspection does not include applied directive information which some tooling (e.g. Apollo Federation) relies on, see https://github.com/graphql/graphql-wg/discussions/1096

> The other use case that has come up is keeping some non-production-ready portions of the graph private but allowing active incremental development on them, by not generating schema for 'experimental' features before the subgraph is fed into the federated supergraph. 

Have you seen [`@inaccessible`](https://opensource.expediagroup.com/graphql-kotlin/docs/schema-generator/federation/federated-directives#inaccessible-directive)?

jdsatlin-exp:
@dariuszkuc, the `@inaccessible` directive seems like a good solution for non-production-ready endpoints, thank you pointing me in that direction.

In terms of PR, would it be a helpful contribution at this point to replace the injected schema writer proposed here with injected SchemaPrinter.Options instead, that would then be picked up by the SDL endpoint? At least speaking in our repo that would allow us to use the SDL endpoint rather than making a nearly-identical endpoint just with customized options.

samuelAndalon:
The problem can be solved by either using introspection query or using the `@inaccessible` directive, the `/sdl` endpoint was added for convenience, and should not be used for other purposes.

closing this PR as won't fix.

PR #1520, opened 2022-08-16
add custom java mapping support to FunctionDataFetcher

cryptoki:
### :pencil: Description
@dariuszkuc @samuelAndalon
it's only a *draft* to evaluate if this way is an option as well. I added a function for java object mapping. The default implementation results `null` but it can overwritten in custom `FunctionDataFetcher`. I updated my little playground project as well.

I'm going to add tests as well. actually it is only a draft.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1515

PR #1544, opened 2022-09-12
Add capability to do Argument conversion using ServiceLoader

WassimAkachi:
…ObjectConverter

### :pencil: Description
`ArgumentObjectConverter` allows application and library developers to provide their own implementation
and logic for how to convert the input argument. The `ArgumentObjectConverter` provide a way to extend
this library easily. It is using the service provider spec from `java.util.ServiceLoader`.
By default, `graphql-kotlin-schema-generator` provides `DefaultArgumentObjectConverter` to convert the input argument
value to the corresponding JVM object.

### :link: Related Links

[ServiceLoader JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)
[Java Service Provider Interface @ baeldung](https://www.baeldung.com/java-spi)

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/convertArgumentValue.kt
@@ -16,15 +16,14 @@
 
 package com.expediagroup.graphql.generator.execution
 
-import com.expediagroup.graphql.generator.exceptions.MultipleConstructorsFound
 import com.expediagroup.graphql.generator.exceptions.PrimaryConstructorNotFound
+import com.expediagroup.graphql.generator.execution.spi.ArgumentObjectConverter
 import com.expediagroup.graphql.generator.internal.extensions.getGraphQLName
 import com.expediagroup.graphql.generator.internal.extensions.getKClass
-import com.expediagroup.graphql.generator.internal.extensions.getName
 import com.expediagroup.graphql.generator.internal.extensions.getTypeOfFirstArgument
-import com.expediagroup.graphql.generator.internal.extensions.isNotOptionalNullable
 import com.expediagroup.graphql.generator.internal.extensions.isOptionalInputType
 import com.expediagroup.graphql.generator.internal.extensions.isSubclassOf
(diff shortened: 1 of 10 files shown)


samuelAndalon:
is there any reason you need to use `ObjectNode` ? why not just create a specify type for `settings` field, which i suspect anyways need a type to do the schema generation ?

as a last resource i would advice to extend the [FunctionDataFetcher](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt#L45) and write your own [dataFetcherFactoryProvider](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLExecutionConfiguration.kt#L44) bean

WassimAkachi:
`ObjectNode` is one of the classes we expect. The settings field is a very complex and dynamic field. This can not be easily changed. And I can't change it and use specific classes instead. Using the feature implemented in this pull request would be very helpful. It would solve such a problem with minimal code change in the application. The implementation and extension of the library is a much easier, stable and faster solution in comparison to changing `ObjectNode` to something else and to change other argument fields which are coming from 3rd party libraries…

How can I succeed with this feature (pull request)? What should I do further to let you accept this PR?

dariuszkuc:
Hello 👋 
Thanks for submitting this PR. While this PR would add extra flexibility, it would also add (some) performance overhead of locating appropriate converter for everyone even if they don't need it (which I'd assume would be vast majority of cases).

Our suggestion [before](https://github.com/ExpediaGroup/graphql-kotlin/pull/1520) was to create custom data fetcher/resolver to handle your specific use case. Creating custom resolver would require re-writing the conversion logic which might be mostly just copy+paste of existing code. Currently argument converters are functions -> @samuelAndalon maybe we should encapsulate it in some (open) class so it would be easier to extend?

samuelAndalon:
>@samuelAndalon maybe we should encapsulate it in some (open) class so it would be easier to extend?

there is definitely room for improvement, the migration from jackson to constructor invocation with reflections make some use cases to break,

we can look into the possibility to encapsulate in a more granular interface the logic that does the argument conversion.

PR #1552, opened 2022-09-14
feat: DataFetchingEnvironment extensions to access to entries in GraphQLContext

samuelAndalon:


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/exceptions/KeyNotFoundInGraphQLContextException.kt
```diff
@@ -0,0 +1,9 @@
+package com.expediagroup.graphql.generator.exceptions
```
dariuszkuc, line 1: missing copyright

PR #1562, opened 2022-09-23
[federation] deprecate @extends directive

dariuszkuc:
### :pencil: Description

Federation v2 no longer requires `@extends` directive due to the smart entity type merging. All usage of `@extends` directive should be removed from Federation v2 schemas.


generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/validation/integration/FederatedKeyDirectiveIT.kt
```diff
@@ -104,15 +104,6 @@ class FederatedKeyDirectiveIT {
         assertTrue(validatedType.hasAppliedDirective("key"))
     }
 
-    @Test
-    fun `@extended type should specify @key directive`() {
```
dariuszkuc, line 108: this test is no longer valid as we are not scanning for `@extends` directive so the underlying type won't be picked up anymore

samuelAndalon (approved):

Issue #1549, opened 2022-09-13
update Kotlin version to 1.7


dariuszkuc:
In the next major release we should update Kotlin to version 1.7. Many of our dependent libraries already updated to the new 1.7 baseline.

dariuszkuc:
resolved in #1572

PR #1572, opened 2022-10-07
[build] update to kotlin 1.7.10

samuelAndalon:
https://github.com/ExpediaGroup/graphql-kotlin/issues/1549


dariuszkuc (approved):
Looks like there were some unrelated changes (in instrumentation tests and context extension) but otherwise looks fine.

Issue #1578, opened 2022-10-15
`GraphQLClientJacksonSerializer` silently skips some `is`-prefixed input fields


myronmarston:
**Library Version**
6.2.5 (and the bug exists as of the current commit in the `master` branch, too).

**Describe the bug**

When an input type has a field named `is<Whatever>` that is _not_ a `Boolean`, then `GraphQLClientJacksonSerializer` omits the field when serializing a request.  Note that this bug is not present when `GraphQLClientKotlinxSerializer` is used as the serializer.

**To Reproduce**

See the tests I added in https://github.com/myronmarston/graphql-kotlin/pull/1--the new test in `GraphQLClientJacksonSerializerTest` fails in a way that reproduces the bug.  Here's the test failure output:

```
expected: <{
  "variables" : {
    "requiredInput" : 1,
    "nullableElementList" : [ ],
    "nonNullableElementList" : [ ],
    "isPrefixedBooleanField" : true,
    "isPrefixedNonBooleanField" : {
      "equalTo" : false
    }
  },
  "query" : "INPUT_QUERY",
  "operationName" : "InputQuery"
}> but was: <{
  "variables" : {
    "requiredInput" : 1,
    "nullableElementList" : [ ],
    "nonNullableElementList" : [ ],
    "isPrefixedBooleanField" : true
  },
  "query" : "INPUT_QUERY",
  "operationName" : "InputQuery"
}>
```

(Note: I have no idea how to fix the bug so I didn't bother sending it as a a pull request here.)

**Expected behavior**

I expect all input fields to be serialized, regardless of the field names used.

myronmarston:
It looks like #1569 may have the same root cause as this bug.

dariuszkuc:
Hello 👋 
It sounds like the issue is with Jackson reliance on JavaBean naming conventions. I think code will need to explicitly annotate those fields with `@JsonProperty` to force Jackson to use appropriate fields (kotlinx doesn't use reflections so it does not have this problem).

myronmarston:
> Hello 👋 It sounds like the issue is with Jackson reliance on JavaBean naming conventions.

That's what I was thinking, too.

> I think code will need to explicitly annotate those fields with `@JsonProperty` to force Jackson to use appropriate fields 

Thanks.  My initial attempt at using that didn't actually work:

``` diff
[...]

Issue #1569, opened 2022-10-03
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "isCustomerEligible" 


viralpatel-phx:
**Library Version**
GraphQL Kotlin version: 6.2.5
Kotlin version: 1.7.0
Spring Boot: 2.7.0

**Describe the bug**
Getting `com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException` exception when serialising the JSON data `{"data":{"isCustomerEligible":{"eligibility":false}}}` from response. The issue only occurs for the below query. 

Complete exception:
```

com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "isCustomerEligible" (class com.example.graphql.client.iscustomereligible.CustomerEligibilityResponse), not marked as ignorable (one known property: "eligibility"])
 at [Source: (String)"{"data":{"isCustomerEligible":{"eligibility":false}}}
"; line: 1, column: 50] (through reference chain: com.expediagroup.graphql.client.jackson.types.JacksonGraphQLResponse["data"]->com.example.graphql.client.iscustomereligible.CustomerEligibilityResponse["isCustomerEligible"])
        at com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:61)
        at com.fasterxml.jackson.databind.DeserializationContext.handleUnknownProperty(DeserializationContext.java:1127)
        at com.fasterxml.jackson.databind.deser.std.StdDeserializer.handleUnknownProperty(StdDeserializer.java:2023)
        at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.handleUnknownProperty(BeanDeserializerBase.java:1700)
        at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.handleUnknownProperties(BeanDeserializerBase.java:1650)
        at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeUsingPropertyBased(BeanDeserializer.java:539)
        at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromObjectUsingNonDefault(BeanDeserializerBase.java:1405)
        at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:351)
[...]

Issue #1486, opened 2022-07-21
Generated code for input with non-standard naming fails


hameno:
**Library Version**
5.5.0

**Describe the bug**
Considering the following schema:

```
input Filter {
    AND: [Filter!]
    OR: [Filter!]

    id: StringFilter
    name: StringFilter
}
```

The maven plugin generates the following class for Jackson:

```
@Generated
public data class Filter(
  public val AND: List<Filter >? = null,
  public val OR: List<Filter >? = null,
  public val id: StringFilter? = null,
  public val name: StringFilter? = null
)
```

This however will fail to deserialize on the server side, as the request will not send "AND" or "OR" as the name of the field in the JSON.
You need to use `@JsonProperty` for this situation.
Is there any way to force this?

**To Reproduce**
see above

**Expected behavior**
Serialized json looks correct

dariuszkuc:
Hello 👋 
This is indeed an issue with Jackson (kotlinx-serialization does not have this issue) that does not honor the original field names and instead uses getters to derive the name based on JavaBean conventions. Unsure if this will be addressed within this lib - options would be to always annotate the fields with `@JsonProperty` (which seems like it would redundant in vast majority of cases) OR try to re-implement jackson name logic to determine whether to apply extra annotation (which might be error prone).

PR #1579, opened 2022-10-19
[client] update Jackson client to always annotate input fields

dariuszkuc:
### :pencil: Description

Update Jackson client generation logic to always annotate input fields with `@get:JsonProperty`. This is a workaround to Jackson limitations due to its reliance on reflections to find getters/setters following JavaBean naming conventions.

Given a simple data class:

```kotlin
data class Foo(
  val isBoolean: Boolean = true,
  val isNotBoolean: String = "yes",
  val NOT: Boolean = true,
  val pID: Int = 1
)
```

It would be serialized as (before this change):

```json
{
  "isBoolean" : true,
  "not" : true,
  "pid" : 1
}
```

After this change:

```json
{
  "isBoolean" : true,
  "isNotBoolean" : "yes",
  "NOT" : true,
  "pID" : 1
}
```

### :link: Related Issues

* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1486
* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1569
* resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1578


samuelAndalon (approved):

myronmarston:
@dariuszkuc thanks for tackling this so quickly!

Issue #1548, opened 2022-09-13
remove deprecated GraphQL context object logic in favor of context map


dariuszkuc:
Generic `GraphQLContext` object map was deprecated in `graphql-kotlin` v5 (and `graphql-java` v17). We should clean up our logic and drop the support for the generic object in the next major release. GraphQL context map is the official approved context mechanism.

dariuszkuc:
Resolved in #1580

PR #1580, opened 2022-10-21
feat: remove usage of GraphQLContext interface

samuelAndalon:
### :pencil: Description
Removing all references of `GraphQLContext` in favor of context map. 
GraphQL context map is the official approved context mechanism.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1548


servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLContextFactory.kt
```diff
@@ -16,20 +16,10 @@
...
-interface GraphQLContextFactory<out Context : GraphQLContext, Request> {
-
-    /**
-     * Generate GraphQL context based on the incoming request and the corresponding response.
-     * If no context should be generated and used in the request, return null.
-     */
-    @Deprecated(message = "The generic context object is deprecated in favor of the context map", replaceWith = ReplaceWith("generateContextMap(request)"))
-    suspend fun generateContext(request: Request): Context? = null
-
+interface GraphQLContextFactory<Request> {
```
smyrick, line 22: Maybe we can default to `GraphQLContext.of()` to be extra clear what the object here is

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/dataFetchingEnvironmentExtensions.kt
```diff
@@ -31,15 +31,15 @@ import java.util.concurrent.CompletableFuture
...
 /**
 * Helper method to get values from a registered DataLoader.
 */
 fun <K, V> DataFetchingEnvironment.getValuesFromDataLoader(dataLoaderName: String, keys: List<K>): CompletableFuture<List<V>> {
     val loader = getDataLoader<K, V>(dataLoaderName) ?: throw MissingDataLoaderException(dataLoaderName)
-    return loader.loadMany(keys, listOf(this.getContext()))
+    return loader.loadMany(keys, listOf(this.graphQlContext))
 }
 
 /**
```
smyrick, line 45: Maybe we can deprecate the methods below too. We do have the useful extension methods on `GraphQLContext` but these just serve as one function call wrappers and we don't even use them in the lib either

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLContextFactory.kt
smyrick: Actually, we could make a MUCH more significant change and instead start using `GraphQLContext` as the type everywhere instead of `Map<*, Any>?`. 

We end up mapping back to a builder anyway in `GraphQLRequest.toExecutionInput()` so if we let everyone just call the builder themselves to modify that might be eaiser

smyrick: That would require everyone to migrate though, and almost certainly break existing code.Tthis is at least a breaking change but will possibly only require removal as someone could have migrated to the new `Map` methods

smyrick:
Make sure you add the proper PR labels

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLContextFactory.kt
samuelAndalon: i think for simplicity it was decided to just return a `Map` and let the `GraphQLContextFactory` build the `GraphQLContext` using the map. 
i would keep it like that for simplicity and to not add another breaking change bc of the potential change of the return type

dariuszkuc: since it is a breaking change anyway -> lets approach it from a new user perspective, what would be a better user experience?

smyrick: So `DataFetchingEnvironment.getGraphQlContext` returns `GraphQLContext` not `Map<>` so I think it would sense that the factory sets an object and that is the one you get back out.

samuelAndalon: applied the updates to return a `GraphQLContext` object instead of a Map

servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/dataFetchingEnvironmentExtensions.kt
```diff
@@ -29,17 +29,25 @@ import java.util.concurrent.CompletableFuture
...
 fun <K, V> DataFetchingEnvironment.getValueFromDataLoader(dataLoaderName: String, key: K): CompletableFuture<V> {
     val loader = getDataLoader<K, V>(dataLoaderName) ?: throw MissingDataLoaderException(dataLoaderName)
-    return loader.load(key, this.getContext())
+    return loader.load(key, this.graphQlContext)
 }
 
 /**
 * Helper method to get values from a registered DataLoader.
 */
+@Deprecated(
```
smyrick, line 44: These methods are useful. I don't think we need to deprecate these. I was referring to the methods to get values from context directly on `DataFetchingEnvironment`, ie lines 57+ in this file

samuelAndalon: ahh ok, well we added those extensions bc of a feature request, where ppl stores in context using KClass, so is definetely more idiomatic do to

```kotlin
dataFetchingEnvironment.getFromContext<CoroutineScope>()
```

than

```kotlin
dataFetchingEnvironment.graphQLContext.get(CoroutineScope::class)
```

smyrick: But with the extension methods [here](https://github.com/ExpediaGroup/graphql-kotlin/blob/415a745254262c8757b0805c2e6a85e6fba9a485/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/extensions/GraphQLContextExtensions.kt#L26-L27) you can do this

```kotlin
dataFetchingEnvironment.graphQLContext.get<CoroutineScope>()
```

samuelAndalon: i still find it useful, but yes, it looks like an unneeded indirection, i can remove these in a separate PR.

smyrick: I don't think they should be deleted just yet, maybe just deprecated, but yes that can be a separate PR

smyrick (approved):

PR #1581, opened 2022-10-27
[federation] cleanup validation rules

dariuszkuc:
### :pencil: Description

Cleanup field set validation rules based on Federation v2 functionality.

Validator now verifies:

- @key, @provides and @requires field sets reference existing fields
- @requires references @external fields
- @provides references an object
- field sets cannot reference unions
- list and interfaces can only be referenced from `@requires` and `@provides`


dariuszkuc:
Apologies for the large PR -> its mostly due to the complete rewrite of the integration tests.

Actual code is just in the `com/expediagroup/graphql/generator/federation/validation` package

PR #1583, opened 2022-10-28
update kotlin patch version

tapaderster:
https://github.com/JetBrains/kotlin/releases/tag/v1.7.20


dariuszkuc:
I believe we need to use the same version of kotlin as used by the kotlin compile testing (as otherwise tests will blow)

PR #1586, opened 2022-11-01
Configure client generate parser options

### LINKED ISSUE #1526: Build plugins should allow configuring max parse tokens
Generate SDL task/mojo currently do not allow configuring this option and instead rely on the defaults provided by `graphql-java`. Since it is possible to configure this option when build a GraphQL server, we should also allow configuring it in build plugins.

> @dariuszkuc I just saw this coming back with `graphql-kotlin` version 6.2.1. That version depends on `graphql-java` 19.1 which *does* allow to configure the `maxTokens` in `ParserOptions`, but I do not see that option being exposed by the Gradle plugin. Am I missing something?

_Originally posted by @sschuberth in https://github.com/ExpediaGroup/graphql-kotlin/issues/1303#issuecomment-1222215513_

mgilbey:
### :pencil: Description

Allows a user of the client generator to set the parser options for graphql-java. The base `generateClient()` function accepts a function type with receiver to customise the builder object, whilst the gradle and maven plugins provide an explicit list of (current) options permitted by graphql-java, with the exception of the [parsingListener](https://github.com/graphql-java/graphql-java/blob/13ab0009bd42b591bb4a542cd9c780a53515321d/src/main/java/graphql/parser/ParserOptions.java#L289).

### :link: Related Issues

Resolves #1526 

### Notes

- I've picked fairly simple naming, but if we suspect any ambiguity of uses, or future collisions, it would be good to call that out so that it's clear this is configuration for the client generation only.
- Rather than abstract over `ParserOptions.Builder` the client module now exposes this in it's public API, and would therefore inherit breaking changes made to this class.

### DIFF
--- a/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
+++ b/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
@@ -1,0 +1,20 @@
+package com.expediagroup.graphql.plugin.gradle.config
+
+import java.io.Serializable
+
+/**
+ * Configure options for parsing GraphQL queries and schema definition language documents. Settings
+ * here override the defaults set by GraphQL Java.
+ */
+data class GraphQLParserOptions(
+    /** Modify the maximum number of tokens read to prevent processing extremely large queries */
+    var maxTokens: Int? = null,
+    /** Modify the maximum number of whitespace tokens read to prevent processing extremely large queries */
+    var maxWhitespaceTokens: Int? = null,
+    /** Memory usage is significantly reduced by not capturing ignored characters, especially in SDL parsing. */
+    var captureIgnoredChars: Boolean? = null,
+    /** Single-line comments do not have any semantic meaning in GraphQL source documents and can be ignored */
+    var captureLineComments: Boolean? = null,
+    /** Memory usage is reduced by not setting SourceLocations on AST nodes, especially in SDL parsing. */
+    var captureSourceLocation: Boolean? = null
+) : Serializable
--- a/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
+++ b/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
@@ -8,13 +8,13 @@ import java.io.Serializable
  */
 data class GraphQLParserOptions(
     /** Modify the maximum number of tokens read to prevent processing extremely large queries */
-    var maxTokens: Int? = 15000,
+    var maxTokens: Int? = null,
     /** Modify the maximum number of whitespace tokens read to prevent processing extremely large queries */
-    var maxWhitespaceTokens: Int? = 200000,
+    var maxWhitespaceTokens: Int? = null,
     /** Memory usage is significantly reduced by not capturing ignored characters, especially in SDL parsing. */
-    var captureIgnoredChars: Boolean? = false,
+    var captureIgnoredChars: Boolean? = null,
     /** Single-line comments do not have any semantic meaning in GraphQL source documents and can be ignored */
-    var captureLineComments: Boolean? = true,
+    var captureLineComments: Boolean? = null,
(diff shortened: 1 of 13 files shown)


website/docs/plugins/gradle-plugin-tasks.mdx
```diff
@@ -263,6 +289,7 @@ resulting generated code will be automatically added to the project main source
 | `serializer` | GraphQLSerializer | | JSON serializer that will be used to generate the data classes.<br/>**Default value is:** `GraphQLSerializer.JACKSON`. |
 | `schemaFile` | File | yes | GraphQL schema file that will be used to generate client code. |
 | `useOptionalInputWrapper` | Boolean | | Boolean opt-in flag to wrap nullable arguments in `OptionalInput` that distinguish between `null` and undefined/omitted value.<br/>**Default value is:** `false`.<br/>**Command line property is**: `useOptionalInputWrapper` |
+| `parserOptions` | GraphQLParserOptions | | Configures the settings used when parsing GraphQL queries and schema definition language documents. |
```
dariuszkuc, line 292: let's keep the alphabetical order -> can you also fix `serializer` <-> `schemaFile` order?

*this comment applies to all doc changes

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
```diff
@@ -0,0 +1,20 @@
+package com.expediagroup.graphql.plugin.gradle.config
+
+import java.io.Serializable
+
+/**
+ * Configure options for parsing GraphQL queries and schema definition language documents. Settings
+ * here override the defaults set by GraphQL Java.
+ */
+data class GraphQLParserOptions(
```
dariuszkuc, line 9: instead of null, should those values have some defaults? i.e. when configuring the plugin I have no idea what would be the default, I have to look up the `graphql-java` code to find out the defaults

website/docs/plugins/gradle-plugin-tasks.mdx
dariuszkuc: should we also specify somewhere in the docs what are the available options?

dariuszkuc (commented):
Looks good, just few nitpicks about the docs and the defaults.

dariuszkuc:
> Rather than abstract over ParserOptions.Builder the client module now exposes this in it's public API, and would therefore inherit breaking changes made to this class.

Client generator module is not intended to be called outside of the maven/gradle plugin which abstract it away so I think it is fine. Unsure if we need to be able to configure all options but why not.

website/docs/plugins/gradle-plugin-tasks.mdx
mgilbey: Fixed the ordering and added the specific options and defaults in the docs.

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
mgilbey: I've set default values that match current graphql-java defaults. They are not particularly accessible in the plugin sources, and if graphql-java changes these we will no longer inherit their defaults. This is easy to reverse in the future if we need to.

plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/config/GraphQLParserOptions.kt
```diff
@@ -0,0 +1,20 @@
...
+    /** Modify the maximum number of tokens read to prevent processing extremely large queries */
+    var maxTokens: Int? = 15000,
```
dariuszkuc, line 11: if those have default values set -> does it make sense to keep those fields as nullable?

mgilbey: If you wish to go back to the default value (whatever it is) then you can explicitly set it as null.

dariuszkuc: seems odd experience for the end user - if the intent is to keep null as not-set and use defaults from `graphql-java` lets revert back to keep those values null and document the behavior in the docs

dariuszkuc (approved):

PR #1591, opened 2022-11-04
[federation] add support for Fed 2.1

dariuszkuc:
### :pencil: Description

By default, Supergraph schema excludes all custom directives. Federation 2.1 adds new `@composeDirective` that can be used to instruct composition logic to preserve custom directives in the Supergraph schema.

```graphql
directive @composeDirective(name: String!) repeatable on SCHEMA
```

### :link: Related Issues
N/A


dariuszkuc:
IMHO in a code first library it doesn't make sense to support multiple versions of Federation v2 (i.e. `2.0`, `2.1`, `2.2` etc) and we should always default to the latest supported version.

If needed we can introduce this functionality behind some opt-into-version mechanism but it will greatly increase complexity and support burden of maintaining multiple versions.

dariuszkuc:
Should be merged after https://github.com/ExpediaGroup/graphql-kotlin/pull/1595

samuelAndalon (approved):

PR #1598, opened 2022-11-08
[build] use reusable workflows to simplify actions

dariuszkuc:
### :pencil: Description

CI and PR builds are very similar so we use reusable workflows to keep the build logic in single place. By separating regular build and examples build into separate jobs, we can now re-run those individually as needed.

See: https://docs.github.com/en/actions/using-workflows/reusing-workflows

### DIFF
--- a/.github/workflows/build-examples.yml
+++ b/.github/workflows/build-examples.yml
@@ -1,0 +1,45 @@
+name: Build and Test Examples
+
+on:
+  workflow_call:
+
+jobs:
+  build:
+    timeout-minutes: 30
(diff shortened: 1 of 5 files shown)


dariuszkuc:
Example of how the workflow could look like (with composition tests) -> https://github.com/dariuszkuc/graphql-kotlin/actions/runs/3423004460

dariuszkuc:
Example composition test results -> https://github.com/dariuszkuc/graphql-kotlin/actions/runs/3423004460

samuelAndalon (approved):

PR #1619, opened 2022-12-15
chore: label 'pre-release' latest unreleased docs

samuelAndalon:


website/docusaurus.config.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
           sidebarPath: require.resolve('./sidebars.js'),
           versions: {
             current: {
-              label: '7.x.x',
+              label: 'pre-release',
               path: '7.x.x'
```
dariuszkuc, line 32: is the path correct? I think latest should be just `""`?

Issue #1589, opened 2022-11-04
[spring-server] replace playground with graphiql


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
Originally we chose Playground as it was much more feature rich than [GraphiQL](https://github.com/graphql/graphiql). Since then project was deprecated in favor of GraphiQL v2 which has full feature parity (and then recently un-deprecated with potential rewrite to reuse GraphiQL internals, still TBD).

**Describe the solution you'd like**
Update `spring-server` to bundle GraphiQL instead of Playground.

**Describe alternatives you've considered**
Since it is just including an extra [HTML file](https://github.com/graphql/graphiql/tree/main/examples/graphiql-cdn) we could potentially ship with both.

**Additional context**
[Playground status](https://github.com/graphql/graphql-playground/issues/1366#issuecomment-1062088978)

PR #1623, opened 2022-12-22
feat: add and use graphiql as default for browser IDE

samuelAndalon:
### :pencil: Description
[Prisma Labs Playground is deprecated](https://github.com/graphql/graphql-playground/issues/1366#issuecomment-1062088978) and we should use the official GraphQL IDE under GraphQL foundation

GraphiQL will now be the default IDE under the `/graphiql` endpoint, while Playground will still be available with explicit opt-in through configuration properties

```yml
graphql:
  graphiql:
    enabled: true
  playground:
    enabled: false
```

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1589

### DIFF
--- a/servers/graphql-kotlin-spring-server/src/main/resources/graphql-graphiql.html
+++ b/servers/graphql-kotlin-spring-server/src/main/resources/graphql-graphiql.html
@@ -1,0 +1,70 @@
+<!--
+ *  Copyright (c) 2021 GraphQL Contributors
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE file in the root directory of this source tree.
+-->
+<!DOCTYPE html>
+<html>
+<head>
+    <title>GraphiQL</title>
+    <style>
+        body {
+            height: 100%;
+            margin: 0;
+            width: 100%;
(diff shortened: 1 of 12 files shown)


servers/graphql-kotlin-spring-server/src/main/resources/graphql-graphiql.html
```diff
@@ -0,0 +1,70 @@
+<!--
```
samuelAndalon, line 1: retrieved it from
https://github.com/graphql/graphiql/tree/main/examples/graphiql-cdn

dariuszkuc:
Instead of

```yaml
graphql:
  browserIDE:
    ide: "PLAYGROUND"
```

I think it might be simpler to keep those configs separate:
 * easier to drop support for tooling in future
 * easier to add additional tooling (if ever needed)
 * ability to use multiple tools

```yaml
graphql:
  playground:
    enabled: true
    endpoint: "graphiql"
  playground:
    enabled: false
    endpoint: "playground"
  # altair|studio|voyager|etc
```

dariuszkuc (approved):

PR #1626, opened 2023-01-05
feat: add explorer and exporter plugins

samuelAndalon:
image

image


Samjin:
Once this goes in, should probably add it to excluded client list in Studio. https://studio.apollographql.com/graph/ExpediaInc-8789/checks/graph/config?variant=test

tapaderster (approved):

PR #1633, opened 2023-01-11
build: migrate to use buildSrc and convention plugins

dariuszkuc:
### :pencil: Description

Additional information:
* [build sources](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources)
* [convention plugins](https://docs.gradle.org/current/samples/sample_convention_plugins.html)


dariuszkuc:
TLDR this moves subproject build logic from root `build.gradle.kts` files to `buildSrc` and makes it a convention plugin that is applied to individual projects.

There are few outstanding Gradle issues
* https://github.com/gradle/gradle/issues/22797 - we need to suppress `DSL_SCOPE_VIOLATION` when referencing version catalog in plugins block
* https://github.com/gradle/gradle/issues/15383 - we need a workaround to access version catalog in convention plugin

samuelAndalon (approved):

Issue #1609, opened 2022-12-06
use new Spring Boot auto-configuration


snv:
**Library Version**
com.expediagroup:graphql-kotlin-spring-server:7.0.0-alpha.0

**Describe the bug**
When switchen to Spring Boot 3 `com.expediagroup.graphql.server.spring.GraphQLAutoConfiguration` does not get executed, because the old way of registering them via `spring.factories` has been removed.
The new way needs a `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports`-File.

**Additional Info**
Details about [the introduction of the new mechanism](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes#auto-configuration-registration), and [the deactivation of the old one](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#auto-configuration-files).

As a workaround i can get my project to work, by registering `com.expediagroup.graphql.server.spring.GraphQLAutoConfiguration` myself with a  `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` in my own application.

hantsy:
>As a workaround i can get my project to work, by registering com.expediagroup.graphql.server.spring.GraphQLAutoConfiguration myself with a META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports in my own application.

Alternatively, add `@ImportAutoConfiguration` on your Spring application class to import it manually.

PR #1638, opened 2023-01-14
update to SpringBoot 3 and Java 17

dariuszkuc:
### :pencil: Description

SpringBoot v3 requires Java 17. This PR updates target bytecode of ALL `graphql-kotlin` modules to Java 17 and as a result ibrary users will have to use Java 17+ as well.

### :link: Related Issues
* Resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1609


gradle/libs.versions.toml
```diff
@@ -1,10 +1,10 @@
 [versions]
-android-plugin = "7.1.2"
+android-plugin = "8.0.0-alpha11"
```
dariuszkuc, line 2: AGP version 7.3+ is required for Java 17 support

* v7.3 was failing complaining about incorrect path to generated sources (which were correct)
* v7.4 was failing due to https://issuetracker.google.com/issues/261329823

tapaderster (approved):

PR #1654, opened 2023-01-23
build: move gradle plugin integration tests to composite build

dariuszkuc:
### :pencil: Description

Moves Gradle plugin integration tests to a composite builds.

Benefits:
* faster build - integration tests are executed as composite build which can be built in parallel and utilize Gradle build cache to speed up process
* easier env variable sharing - we no longer need to pass system properties to specify which version of library to use, `graphql-kotlin` libraries can be referenced without version as they are part of the build and we can rely on same version catalog as main build
* easier to debug build problems - each integration test is now a valid Gradle project that can be easily loaded in IntelliJ (or build from cmd line)


dariuszkuc:
NOTE: IntelliJ auto creates those gradle wrapper files for each composite build, will look whether we can easily drop them once I move all the integration tests.

dariuszkuc:
Since the PR is already pretty big -> I'll follow up with separate ones for migrating remaining integration tests.

samuelAndalon (approved):

PR #1656, opened 2023-01-24
build: gradle plugin download sdl / introspection integration tests

dariuszkuc:
### :pencil: Description

Migrate download sdl task and introspection task integration tests to a composite build. Failure scenarios (i.e. timeout) is still run through `GradleRunner` in the plugin project.


plugins/graphql-kotlin-gradle-plugin/src/integration/download-sdl-timeout/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
+import com.expediagroup.graphql.plugin.gradle.config.TimeoutConfiguration
+import com.expediagroup.graphql.plugin.gradle.tasks.GraphQLDownloadSDLTask
+
+plugins {
+  kotlin("jvm") version "1.7.21"
```
dariuszkuc, line 5: we could potentially load this from version catalog (but would have to update the test logic to copy the toml file - will look into this later)

dariuszkuc:
NOTE: big size of the PR (lines changed) is due to introspection/sdl results copied over for wiremock in the composite build.

samuelAndalon (approved):

PR #1659, opened 2023-01-26
build: update gradle plugin failure tests to use version catalog

dariuszkuc:


plugins/graphql-kotlin-gradle-plugin/src/test/kotlin/com/expediagroup/graphql/plugin/gradle/tasks/GraphQLGenerateClientTaskIT.kt
```diff
@@ -37,6 +37,9 @@ class GraphQLGenerateClientTaskIT : WireMockAbstractIT() {
         val sourceDirectory = File("src/integration/download-sdl-timeout")
         sourceDirectory.copyRecursively(testProjectDirectory)
 
+        // version catalog setup
+        File("../../gradle/libs.versions.toml").copyTo(File(testProjectDirectory, "gradle/libs.versions.toml"))
```
dariuszkuc, line 41: Alternative would be to create `settings.gradle.kts` files in each integration test directory that would load version catalog from path that is specified as env variable. I think this approach is simpler

samuelAndalon (approved):

PR #1661, opened 2023-01-27
federation: add support for Apollo Federation subgraph spec v2.3

dariuszkuc:
### :pencil: Description

Updates `federation` module to support Apollo Federation subgraph spec v2.3

Changes:
* v2.2 - update `@shareable` definition to be repeatable to allow annotating both types and their extensions (NOTE: this functionality is not applicable to `graphql-kotlin`)
* v2.3 - adds new `@interfaceObject` directive that allows you to extend interface entity functionality in subgraphs, i.e. by applying `@interfaceObject` directive on a type we provide meta information to the composition logic that this entity type is actually an interface in the supergraph. This allows us to extend interface functionality without knowing any of its implementing types.


dariuszkuc:
note: `pr-check-federation.yaml` workflow had a bug in it as it used `pull_request_target` event (so the GH PAT secret could be passed over and we get PR comments) but the invoked workflows were not checking out the correct code, i.e. federation tests were run against origin codebase vs actual PR.

I fixed the federation workflows in the PR but we won't get updated results (for `@composeDirective`/`@interfaceObject`) in this PR until workflows are fixed, i.e. merge https://github.com/ExpediaGroup/graphql-kotlin/pull/1671 and rebase

samuelAndalon (approved):

Issue #1471, opened 2022-06-16
Create a ktor-plugin for easy setup


jmfayard:
**Is your feature request related to a problem? Please describe.**

We use graphql-kotlin with ktor at https://tignum.com/ and we really like it.

The initial setup OTOH is not so nice. 
It made my head explode when I first set-it up, 
and then it made the head of my colleagues explode when I explained it to them.

The ktor way to install new feature is :
- add a dependency to a plugin
- `install(ThePlugin)` 
- with a declarative configuration lambda
- abstract away implementation details

**Describe the solution you'd like**

I had a shot at what such a plugin could look like. 

Have a look at https://github.com/ExpediaGroup/graphql-kotlin/compare/master...jmfayard:ktor-plugin?expand=1

The simplest setup for `graphql-kotlin/examples/server/ktor-server` would be:

```kotlin
// build.gradle.kts
implementation("com.expediagroup", "graphql-kotlin-ktor-server", latestVersion)
```

and your main file:

```kotlin
fun Application.graphQLModule() {
    install(Routing)
    install(ContentNegotiation) {
        jackson()
    }
    install(GraphQLKotlin) {
        queries = listOf(
            HelloQueryService(),
            BookQueryService(),
            CourseQueryService(),
            UniversityQueryService(),
        )
        mutations = listOf(
            LoginMutationService()
        )
        schemaGeneratorConfig = SchemaGeneratorConfig(
            supportedPackages = listOf(
                "com.expediagroup.graphql.examples.server.ktor"
            ),
        )
     }   
}        
```

For a more advanced setup you could use those optional parameters

```kotlin
fun Application.graphQLModule() {
    // ...
    install(GraphQLKotlin) {
       // Required settings: 
       // see above
       
       // OPTIONAL SETTINGS
       generateContextMap { request: ApplicationRequest ->
            mapOf(
                User::class to generateUser(request),
                MyHeaders::class to generateMyHeaders(request)
            )
        }
        configureGraphQL {
            valueUnboxer(IDValueUnboxer())
        }
        endpoints {
            graphql = "graphql"
            sdl = "sdl"
            playground = "playground"

            enableSdl = true
            enablePlayground = true
        }
    }    
}
```

**Describe alternatives you've considered**

[...]

PR #1667, opened 2023-02-05
add ktor server plugin module

dariuszkuc:
### :pencil: Description

Creates Ktor GraphQL Plugin with similar functionality as the existing SpringBoot auto-configuration module.

```kotlin
fun Application.graphQLModule() {
    install(GraphQL) {
        schema {
            packages = listOf("com.example")
            queries = listOf(HelloWorldQuery())
        }
    }
}
```

### :link: Related Issues

Resolves: https://github.com/ExpediaGroup/graphql-kotlin/issues/1471


servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/GraphQLConfiguration.kt
```diff
@@ -0,0 +1,206 @@
...
+import graphql.execution.instrumentation.Instrumentation
+import graphql.execution.preparsed.PreparsedDocumentProvider
+import io.ktor.server.config.ApplicationConfig
+import io.ktor.server.config.tryGetString
+import io.ktor.server.config.tryGetStringList
+
+/**
+ * Configuration properties that define supported GraphQL configuration options.
+ */
+class GraphQLConfiguration(config: ApplicationConfig)
```
dariuszkuc, line 43: I started with the same configuration properties as used by Spring. This config contains some properties that can be set through configuration files (i.e. same as spring) and others that can only be set programmatically.

I'm not a big fan of how this is all structured at the moment so most likely structure of this file will change.

In general there are 3 areas of configuration:
- schema - everything related to schema generation (e.g. which queries to include)
- engine - everything related to GraphQL Java engine configuration
- server - Ktor server custom config (e.g. context factory)

I'm thinking I will move the existing properties (which were matching spring config) under those three sections, e.g. when configuring schema I think it would make sense to specify target packages in the same place

dariuszkuc: One drawback of reorganizing the properties would be the change in property file mapping - currently properties match the path from the root `graphql` prefix, e.g. `graphql.packages` is a top level `packages` property. If we move those around then we either have to use different properties than in Spring (not ideal but I guess it should be fine?) or have a mismatch between paths and variables, e.g. 

* `graphql.packages` would become `graphql.schema.packages`
* `graphql.introspection.enabled` would become `graphql.engine.introspection.enabled`

samuelAndalon (approved):

Issue #1672, opened 2023-02-14
Ktor server integration receiveText can use wrong encoding


rocketraman:
In Ktor, the `receiveText` function follows the RFC rules for reading plain text, which defaults to an encoding of ISO-8859-1. See the following references:

https://www.rfc-editor.org/rfc/rfc7231#appendix-B
https://www.rfc-editor.org/rfc/rfc8259.html#section-8.1
https://stackoverflow.com/a/49552784/430128
https://youtrack.jetbrains.com/issue/KTOR-789

However, the GraphQL specification states that in the absence of explicit encoding information from the client, UTF-8 must be assumed:

> If the media type in a `Content-Type` or `Accept` header includes encoding information, then the encoding MUST be `utf-8` (e.g. `Content-Type: application/graphql-response+json; charset=utf-8`). If encoding information is not included then `utf-8` MUST be assumed.

(from https://github.com/graphql/graphql-over-http/blob/dd8817065728ff1db78c1ac77ff22eb516e5d3b2/spec/GraphQLOverHTTP.md?plain=1#L196-L199)

The current version of the request parser uses `receiveText` without specifying UTF-8 as the default, which will likely result in incorrect parsing if the client does not specify an encoding.

https://github.com/ExpediaGroup/graphql-kotlin/blob/5149be9dc46a512ff6ff11119470e7d116d77655/servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt#L64

The correct code should do something like this:

```
val reader = request.call.receiveStream().reader(request.contentCharset() ?: Charsets.UTF_8)
```

rocketraman:
Ktor 2.x now apparently no longer has the same behavior -- receiveText does now default to UTF-8 as of:

https://github.com/ktorio/ktor/commit/5121d10292a3dbe256fcc050fb155b46c965ae91
https://youtrack.jetbrains.com/issue/KTOR-4426

PR #1673, opened 2023-02-14
[ktor-server] fix: encoding with default content-type

rocketraman:
### :pencil: Description

For the ktor server, fixes the default encoding used to read the request body when the client does not specify an encoding in the content type explicitly.

### :link: Related Issues

https://github.com/ExpediaGroup/graphql-kotlin/issues/1672

### DIFF
--- a/servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
+++ b/servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
@@ -24,7 +24,10 @@ import com.fasterxml.jackson.databind.type.MapType
 import com.fasterxml.jackson.databind.type.TypeFactory
 import io.ktor.http.HttpMethod
 import io.ktor.server.request.ApplicationRequest
-import io.ktor.server.request.receiveText
+import io.ktor.server.request.receiveStream
+import io.ktor.server.request.contentCharset
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import java.io.IOException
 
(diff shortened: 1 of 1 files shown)


servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
```diff
@@ -23,8 +23,9 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.type.MapType
 import com.fasterxml.jackson.databind.type.TypeFactory
 import io.ktor.http.HttpMethod
-import io.ktor.server.request.ApplicationRequest
-import io.ktor.server.request.receiveText
+import io.ktor.server.request.*
```
dariuszkuc, line 26: ktlint complains about the wildcard imports -> can you fix it?

rocketraman: Yep, done in https://github.com/ExpediaGroup/graphql-kotlin/pull/1673/commits/a5f75e17d4a270b947aa52ddbee692ad5f66c05c

rocketraman:
@dariuszkuc Ok, a couple of tests failed. While investigating their resolution, I noticed that ktor 2.x now apparently no longer has the same behavior -- `receiveText` does now default to UTF-8 as of:

https://github.com/ktorio/ktor/commit/5121d10292a3dbe256fcc050fb155b46c965ae91
https://youtrack.jetbrains.com/issue/KTOR-4426

Therefore I would say this PR is no longer required.

dariuszkuc:
Thanks for the update!

Issue #1139, opened 2021-05-06
Kotlin Multiplatform with gradle plugin - Task with name 'compileKotlin' not found


Saturas89:
Hello all,
I am trying to use the graphql-kotlin-client in my KMM project on server side (jvmMain) and
want to create the query objects with the latest gradle plugin com.expediagroup.graphql.
I cannot run any graphql gradle task.

`* What went wrong:
A problem occurred configuring root project 'my-project'.
> Could not create task ':graphqlGenerateClient'.
   > Task with name 'compileKotlin' not found in root project 'my-project'.`

I attached my build.gradle.kts
[build.txt](https://github.com/ExpediaGroup/graphql-kotlin/files/6433860/build.txt)

smyrick:
Duplicate of https://github.com/ExpediaGroup/graphql-kotlin/issues/876

smyrick:
@Saturas89 We currently do not publish Kotlin Multiplatform artifacts. Your build file also indicates you are trying to use Kotlin 1.5 and we have not updated the library yet from 1.4

PR #1681, opened 2023-02-22
fix(gradle-plugin): compileKotlin does not exist in Multiplatform

nothendev:
Instead the gradle plugin would use compileKotlinJvm.
This is a temporary fix, but in the short-term, it would make the gradle plugin compatible with Multiplatform.

### :pencil: Description
Fix of a Gradle error, that when running with Multiplatform, the Gradle plugin cannot find a task named "compileKotlin", which does not exist in a Multiplatform context.

### :link: Related Issues
#1139
#876

### DIFF
--- a/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt
+++ b/plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt
@@ -170,7 +170,12 @@ class GraphQLGradlePlugin : Plugin<Project> {
 
             val configuration = project.configurations.getAt(GENERATE_SDL_CONFIGURATION)
             generateSDLTask.pluginClasspath.setFrom(configuration)
-            generateSDLTask.dependsOn(project.tasks.findByName("compileKotlin") ?: project.tasks.named("compileKotlinJvm"))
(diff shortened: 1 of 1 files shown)


plugins/graphql-kotlin-gradle-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/gradle/GraphQLGradlePlugin.kt
```diff
@@ -170,7 +170,7 @@ class GraphQLGradlePlugin : Plugin<Project> {
 
             val configuration = project.configurations.getAt(GENERATE_SDL_CONFIGURATION)
             generateSDLTask.pluginClasspath.setFrom(configuration)
-            generateSDLTask.dependsOn(project.tasks.named("compileKotlin"))
+            generateSDLTask.dependsOn(project.tasks.findByName("compileKotlin") ?: project.tasks.named("compileKotlinJvm"))
```
dariuszkuc, line 173: while having explicit dependency on `compileKotlin` task is generally good, I'm thinking whether we should only configure it if it exists, e.g. something along the lines
```suggestion
            val compileKotlinTask = project.tasks.findByName("compileKotlin") ?: project.tasks.findByName("compileKotlinJvm")
            if (compileKotlinTask != null) {
                generateSDLTask.dependsOn(compileKotlinTask)
            } else {
                project.logger.warn("compileKotlin/compileKotlinJvm tasks not found. Unable to auto-configure the generateSDLTask dependency on compile task.")
            }
```

samuelAndalon (approved):

dariuszkuc:
Thanks @nothendev !

PR #1687, opened 2023-02-24
fix federation integration workflows

dariuszkuc:
### :pencil: Description

Conditional steps do not work in workflows that are invoked from other workflows :(


dariuszkuc:
We need `pull_request_target` event to be able to pass secrets (needed to comment on PRs) to PRs made from forks. One way to work around this limitation is to have separate workflow that triggers based on completion of previous workflow but that is somewhat PITA to setup (e.g. see [example](https://github.com/apollographql/apollo-federation-subgraph-compatibility/blob/main/.github/workflows/comment.yaml))

.github/workflows/pr-check-federation.yml
```diff
@@ -9,9 +9,103 @@ on:
...
+      - name: Stop Supergraph
+        if: ${{ always() }}
+        run: docker compose down --remove-orphans
+
+  federation-compatibility:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: integration/federation-compatibility
```
dariuszkuc, line 80: sadly GH action defaults/step config for `working-directory` only work for `run` steps... when using custom actions they have to implement their own working directory logic (hence the duplicate param below).

dariuszkuc:
fed compatibility failure is due to invalid workflow -> it runs workflow from `master` against the code from the fork (so it cannot find the referenced file).

samuelAndalon (approved):

PR #1704, opened 2023-03-07
[ktor server] simplify KtorGraphQLRequestParser

dariuszkuc:
### :pencil: Description

We can rely on `ContentNegotiation` plugin to automatically parse incoming requests for us. We don't have to read incoming requests as plain text and then attempt to manually map it.

### DIFF
--- a/servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
+++ b/servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
@@ -24,7 +24,7 @@ import com.fasterxml.jackson.databind.type.MapType
 import com.fasterxml.jackson.databind.type.TypeFactory
 import io.ktor.http.HttpMethod
 import io.ktor.server.request.ApplicationRequest
-import io.ktor.server.request.receiveText
+import io.ktor.server.request.receive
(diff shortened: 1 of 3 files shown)


servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
```diff
@@ -61,9 +60,8 @@ class KtorGraphQLRequestParser(
     }
 
     private suspend fun parsePostRequest(request: ApplicationRequest): GraphQLServerRequest? = try {
-        val rawRequest = request.call.receiveText()
-        mapper.readValue(rawRequest, GraphQLServerRequest::class.java)
+        request.call.receive()
```
samuelAndalon, line 63: how does the request know to which type the request payload needs to be deserialized to ? is it an inlineClass ?

```kotlin
request.call.receive<GraphQLServerRequest> 
```

samuelAndalon (approved):

servers/graphql-kotlin-ktor-server/src/main/kotlin/com/expediagroup/graphql/server/ktor/KtorGraphQLRequestParser.kt
dariuszkuc: and yes `receive` is inline reified method

PR #1721, opened 2023-03-13
chore: remove gradle duplicates from composite builds

dariuszkuc:
### :pencil: Description

In order to keep intellij "happy" we need to have `gradle/wrapper/gradle.properties` (as otherwise it will attempt to recreate it). In order to avoid duplication of having multiple wrappers (oneper each composite build) potentially running different versions of Gradle, I've updated them to follow symbolic links instead.

NOTE: this change introduces linux symbolic links which means they will not work for Windows machines. It is inconvenient but it is the simplest way to dedup those scripts. Windows users will have to execute `gradlew.bat` from root dir OR invoke Gradle directly.

### :link: Related Issues
N/A


dariuszkuc:
looks like bump to gradle v8 requires some additional changes -> will use separate PR for that

samuelAndalon (approved):

PR #1734, opened 2023-03-27
Update `graphiql-plugin-code-exporter.umd.js` checksum

martinbonnin:
Opening this in case other bumps into this. GraphiQL started to show errors this morning:

```
Failed to find a valid digest in the 'integrity' attribute for resource 'https://unpkg.com/@graphiql/plugin-code-exporter/dist/graphiql-plugin-code-exporter.umd.js' 
with computed SHA-512 integrity 'bZesuT5p2QZ6cSlpgPxI83Db5G3Bw35RTKz+Z+kc6RUu4/PtPkNddzax0VWA6VoXvIwT8s4i5nQklZ+AGJQD2Q=='.
 The resource has been blocked.
```

I'm not exactly sure what's happening, opening this as draft until we have more information.

Edit: looks like there was a [release 13h ago](https://www.npmjs.com/package/@graphiql/plugin-code-exporter) so that might explain things. Might be worth pinning the version?

PR #1735, opened 2023-03-28
fix(server): pin GraphiQL versions

dariuszkuc:
### :pencil: Description

* pin `@graphiql/plugin-code-exporter` version to ensure GraphiQL won't break when new version is published
* move `graphql-graphiql.html` resource to common `graphql-kotlin-server` module so the same file is used by spring and ktor servers

### :link: Related Issues

Supersedes: https://github.com/ExpediaGroup/graphql-kotlin/pull/1734


dariuszkuc:
NOTE: I did not update `graphiql` to latest (2.4.0) as it was blowing up with missing some polyfill function

samuelAndalon:
should we just skip integrity checks ?

dariuszkuc:
> should we just skip integrity checks ?

I'd keep those as it ensures that no-one tampers with the referenced resources.

samuelAndalon (approved):

PR #1739, opened 2023-04-06
feat: new GraalVM reflect metadata generator

dariuszkuc:
### :pencil: Description

Creates new `graphql-kotlin-graalvm-metadata-generator` project that will be used by Gradle and Maven plugins to auto-generate GraalVM reachability metadata. In order to keep the size of PRs manageable, plugin PRs will be created separately.

[GraalVM reachability metadata](https://www.graalvm.org/22.2/reference-manual/native-image/metadata/) is required by GraalVM to support reflections and resource handling. This feature will be used by plugins as way to simplify running `graphql-kotlin` servers as GraalVM native images. Without this functionality, end users have to start their applications with GraalVM agent and then manually run **ALL** possible execution paths (i.e. queries, loading graphiql, etc).

### DIFF
--- a/plugins/server/graphql-kotlin-graalvm-metadata-generator/src/main/kotlin/com/expediagroup/graphql/plugin/graalvm/DefaultMetadataLoader.kt
+++ b/plugins/server/graphql-kotlin-graalvm-metadata-generator/src/main/kotlin/com/expediagroup/graphql/plugin/graalvm/DefaultMetadataLoader.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023 Expedia, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.plugin.graalvm
+
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import java.io.InputStream
+
+private const val DEFAULT_REFLECT_CONFIG = "default-reflect-config.json"
+private const val DEFAULT_RESOURCE_CONFIG = "default-resource-config.json"
+
+object DefaultMetadataLoader {
+    /**
+     * Load default GraalVM reflect metadata required to run native graphql-kotlin servers
+     */
+    fun loadDefaultReflectMetadata(): List<ClassMetadata> {
+        val defaultResources = DefaultMetadataLoader.javaClass.classLoader.getResourceAsStream(DEFAULT_REFLECT_CONFIG)
+            ?: throw IllegalStateException("Unable to load graphql-kotlin GraalVM reflect metadata")
+        val mapper = jacksonObjectMapper()
+        return mapper.readValue(defaultResources, object : TypeReference<List<ClassMetadata>>() {})
+    }
+
+    /**
+     * Open up InputStream to default GraalVM resource config file to be used with graphql-kotlin servers.
+     */
+    fun defaultResourceMetadataStream(): InputStream =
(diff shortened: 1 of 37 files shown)


settings.gradle.kts
```diff
@@ -53,6 +54,7 @@ project(":graphql-kotlin-client-generator").projectDir = file("plugins/client/gr
 project(":graphql-kotlin-sdl-generator").projectDir = file("plugins/schema/graphql-kotlin-sdl-generator")
 project(":graphql-kotlin-hooks-provider").projectDir = file("plugins/schema/graphql-kotlin-hooks-provider")
 project(":graphql-kotlin-federated-hooks-provider").projectDir = file("plugins/schema/graphql-kotlin-federated-hooks-provider")
+project(":graphql-kotlin-graalvm-metadata-generator").projectDir = file("plugins/server/graphql-kotlin-graalvm-metadata-generator")
```
dariuszkuc, line 57: re: new `server` directory - this logic didn't make much sense to put under existing `schema` directory, directory is just used for grouping related modules 

maybe we should move schema projects under `server` functionality? or just keep those separate

plugins/server/graphql-kotlin-graalvm-metadata-generator/src/main/kotlin/com/expediagroup/graphql/plugin/graalvm/GenerateGraalVmMetadata.kt
```diff
@@ -0,0 +1,134 @@
...
+fun generateGraalVmMetadata(supportedPackages: List<String>): List<ClassMetadata> {
+    val defaultMetadata: List<ClassMetadata> = loadDefaultReflectMetadata()
+    val reflectMetadata: List<ClassMetadata> = generateGraalVmReflectMetadata(supportedPackages)
+    return reflectMetadata + defaultMetadata
+}
+
+/**
+ * Generate GraalVM reflect metadata for the underlying GraphQL schema.
+ */
+fun generateGraalVmReflectMetadata(supportedPackages: List<String>): List<ClassMetadata> {
```
dariuszkuc, line 52: this should be `internal` function  but due to some Gradle (fixable) and IDE (open issue) issues it makes it simpler to just keep it public for now (module should generally be used from plugins anyway).

plugins/server/graphql-kotlin-graalvm-metadata-generator/src/main/kotlin/com/expediagroup/graphql/plugin/graalvm/DefaultMetadataLoader.kt
```diff
@@ -0,0 +1,43 @@
...
+        val defaultResources = DefaultMetadataLoader.javaClass.classLoader.getResourceAsStream(DEFAULT_REFLECT_CONFIG)
+            ?: throw IllegalStateException("Unable to load graphql-kotlin GraalVM reflect metadata")
+        val mapper = jacksonObjectMapper()
+        return mapper.readValue(defaultResources, object : TypeReference<List<ClassMetadata>>() {})
+    }
+
+    /**
+     * Open up InputStream to default GraalVM resource config file to be used with graphql-kotlin servers.
+     */
+    fun defaultResourceMetadataStream(): InputStream =
```
dariuszkuc, line 40: not a big fan of just returning an `InputStream` but since this function would be invoked by plugins to just copy the contents to a local project I don't think there is a need to parse it to a String/some objects. Alternatively we could accept a target file and this method would be responsible for copying it over, i.e.

```kotlin
fun copyResourceMetadata(target: File): Boolean {
    val source = DefaultMetadataLoader.javaClass.classLoader.getResourceAsStream(DEFAULT_RESOURCE_CONFIG)
        ?: throw IllegalStateException("Unable to load graphql-kotlin GraalVM resources metadata")
    val result = Files.copy(source, target.toPath())
    return result > 0;
}
```

WDYT?

dariuszkuc: I'll update with gradle plugin PR

plugins/server/graphql-kotlin-graalvm-metadata-generator/src/main/kotlin/com/expediagroup/graphql/plugin/graalvm/GenerateGraalVmMetadata.kt
dariuszkuc: updated tests so it can be internal

samuelAndalon (approved):

PR #1742, opened 2023-04-10
graalvm gradle plugin task

dariuszkuc:


dariuszkuc:
Few notes:
* cannot use `nativeTest` task directly as it is unaware of the generated metadata files - we should revisit the implementation later so we could just run the same unit tests as for regular build
* I am not using `nativeRun` task directly because it takes a long time to build native image and we would need to add some custom polling to ensure app is up before running the integration (we need to start the server in the background to be able to run some tests against it)

samuelAndalon (approved):

PR #1749, opened 2023-04-15
feat: update graphql-java version

samuelAndalon:
### :pencil: Description
update `graphql-java` from 19.2 to 20.2

https://github.com/graphql-java/graphql-java/releases/tag/v20.0
https://github.com/graphql-java/graphql-java/releases/tag/v20.1
https://github.com/graphql-java/graphql-java/releases/tag/v20.2

graphql-java 20.0 introduces [Lightweight Data Fetchers](https://github.com/graphql-java/graphql-java/pull/2953) under the [LightDataFetcher](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/LightDataFetcher.java#L14) interface, which willa void creating a `DataFetchingEnvironment` instance each time the execution engine attempt to resolve a property using the `PropertyDataFetcher`.

updates **federation-jvm** to [3.0.0](https://github.com/apollographql/federation-jvm/releases/tag/v3.0.0)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/PropertyDataFetcher.kt
```diff
@@ -16,22 +16,36 @@
...
 import kotlin.reflect.KProperty
 
 /**
  * Property [DataFetcher] that directly invokes underlying property getter.
  *
- * @param propertyGetter Kotlin property getter that will be invoked to resolve a field
+ * @param propertyGetter Kotlin's property getter that will be invoked to resolve a field
  */
-class PropertyDataFetcher(private val propertyGetter: KProperty.Getter<*>) : TrivialDataFetcher<Any?> {
+class PropertyDataFetcher(private val propertyGetter: KProperty.Getter<*>) : LightDataFetcher<Any?> {
```
samuelAndalon, line 31: `PropertyDataFetcher` will now implement the  [LightDataFetcher](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/LightDataFetcher.java#L14), which will make the execution engine to create the **dataFetchingEnviroment** on demand, most (if not all), PropertyDataFetchers dont require this field, so this will represent a memory improvement

examples/server/spring-server/src/main/kotlin/com/expediagroup/graphql/examples/server/spring/instrumentation/TrackTimesInvokedInstrumentation.kt
```diff
@@ -33,22 +33,22 @@ import java.util.concurrent.ConcurrentHashMap
  * Adds field count tracking in the instrumentation layer if [com.expediagroup.graphql.examples.directives.TrackTimesInvoked] is present.
  */
 @Component
-class TrackTimesInvokedInstrumentation : SimpleInstrumentation() {
+class TrackTimesInvokedInstrumentation : Instrumentation {
```
dariuszkuc, line 36: per deprecation notice - should we use `SimplePerformantInstrumentation` instead?

dariuszkuc: the diff is the interface defaults new method to call deprecated methods, from `SimplePerformantInstrumentation` javadoc

> An implementation of Instrumentation that does nothing. It can be used as a base for derived classes where you only implement the methods you want to. The reason this class is designated as more performant is that it does not delegate back to the deprecated methods and allocate a new state object per call.
> 
> This behavior was left in place for backwards compatibility reasons inside Instrumentation and SimpleInstrumentation but has not been done in this class since no existing classes could have derived from it. If you want more performant behavior on methods you don't implement then this is the base class to use, since it will not delegate back to old methods and cause a new state to be allocated.

samuelAndalon: will update dataloader instrumentations as well then.

dariuszkuc (approved):

Issue #1747, opened 2023-04-12
[7.x.x] FederatedTypePromiseResolverExecutor exception handling does not work as intended


roookeee:
**Library Version**
7.0.0-alpha4

**Describe the bug**
`FederatedTypePromiseResolverExecutor` uses a `try{} catch{}` block when invoking the resolvers `resolve` method. The intention behind this was to allow exception to be turned into proper GraphQLErrors but the `resolve` invocation only starts the future while resolution errors most likely arrive when awaiting the future (e.g. a not found exception is not raised when starting a future).

Throwing an exception inside the returned `CompleteableFuture` makes the whole federated request return `data: null` which is unintended.

Thus there is currently no correct way to throw exceptions to create GraphQLErrors inside a `FederatedTypePromiseResolver`.

**To Reproduce**
Use `FederatedTypePromiseResolverExecutor` and throw an exception inside the `CompleteableFuture` resolution.

**Expected behavior**
Throwing exceptions inside a `FederatedTypePromiseResolver` should create the appropriate GraphQLError while still providing data for all non-exceptional resolutions.

samuelAndalon:
Thank you for reporting this one, i was able to repro it. The problem seems to be on the layer where we convert a `List<CompletableFuture<out T>>` into a `CompletableFuture<List<T>>` if one future from the list completes exceptionally the final completable future completes exceptionally regardless if other futures in the list completed correctly.

working on the fix

PR #1753, opened 2023-04-17
feat: FederatedTypePromiseResolver to settle all promises regardless of errors

samuelAndalon:
### :pencil: Description
the `FederatedTypePromiseResolver` was originally created with the `DataLoader` use case in mind, meaning that all returned `CompletableFuture`s are resolved synchronously, to let `DataLoader` dispatch mechanism to decide when to dispatch, so realistically, when using `DataLoader`s, there is not an scenario where a particular `CompletableFuture` might complete exceptionally.

However, nothing stops users to use the `FederatedTypePromiseResolver` without dataloaders, which means that now `CompletableFuture`s might complete exceptionally in a asynchronous way.

To solve that, create allSettled extension function 
```kotlin
fun List<CompletableFuture<out T>>.allSettled(): CompletableFuture<List<Result<T>>>
```

which will complete all promises even if one or more complete exceptionally. The type of each element in the returned list will be a [kotlin Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) that encapsulates a successful outcome with a value of type `T` or a failure with the exception.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1747

### DIFF
--- a/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
+++ b/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Expedia, Inc
+ * Copyright 2023 Expedia, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,6 @@
 package com.expediagroup.graphql.generator.federation.extensions
 
 import java.util.concurrent.CompletableFuture
-import java.util.concurrent.CompletionException
 
 /**
  * Returns a [CompletableFuture] that completes when all the input futures have completed,
--- a/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
+++ b/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
@@ -20,10 +20,32 @@ import java.util.concurrent.CompletableFuture
 /**
  * Returns a new CompletableFuture of a list with the resolved values of the given CompletableFutures.
  * the returned completableFuture will complete when all given CompletableFutures complete.
+ * If any of the given CompletableFutures complete exceptionally, then the returned CompletableFuture also does so,
+ * with a CompletionException holding this exception as its cause.
  */
-internal fun <T : Any?> List<CompletableFuture<out T>>.collectAll(): CompletableFuture<List<T>> =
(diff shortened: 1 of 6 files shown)


generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensionsKtTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import java.util.concurrent.CompletionException
+import java.util.concurrent.ExecutionException
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+import kotlin.test.assertTrue
+
+class CompletableFutureExtensionsKtTest {
+    @Test
+    fun `joinAll asynchronously collects a list of completable futures of elements into a completable future list of elements`() {
+        val firstPromise = "first promise".toMono().delayElement(Duration.ofMillis(500)).toFuture()
```
dariuszkuc, line 33: `toMono()`? no love for coroutines? :P

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
```diff
@@ -18,12 +18,42 @@ package com.expediagroup.graphql.generator.federation.extensions
...
+
+/**
+ * Returns a [CompletableFuture] that completes when all the input futures have completed,
+ * with a list of [Result] objects that indicate whether each future completed successfully or with an error.
+ * If a future completed with an error, the corresponding [Result] object will contain the exception that was thrown.
+ *
+ * @return a [CompletableFuture] that completes with a list of [Result] objects.
+ */
+@Suppress("TooGenericExceptionCaught")
+internal fun <T : Any?> List<CompletableFuture<out T>>.allSettled(): CompletableFuture<List<Result<T>>> {
```
dariuszkuc, line 43: instead of `allSettled()` maybe this should be `safelyCompleteAll()` or something along those lines

samuelAndalon: had the same doubt about how to name this extension fn, then saw this
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensionsKtTest.kt
samuelAndalon: one of each :P, stream, coroutine and promise

tapaderster (approved):

PR #1763, opened 2023-04-28
fix: Improve alias handling in client generation

### LINKED ISSUE #1760: [client] Aliases are not taken into account correctly when result classes are generated
**Library Version**
6.4.0

**Describe the bug**
Take a look at this schema and query:
```graphql
# schema.graphql
type Artifact {
    artifactName: String!
    someValue: String!
}

type Component {
    componentName: String!
    artifact(artifactName: String!): Artifact
}

type Bundle {
    bundleId: String!
    component(componentName: String!): Component
}

type Query {
    getBundle(bundleId: String!): Bundle
}
```
```graphql
# GetBundleQuery.graphql
query GetBundleQuery($bundleId: String!) {
    getBundle(bundleId: $bundleId) {
        bundleId
        componentFoo: component(componentName: "foo") {
            componentName
            artifactLorem: artifact(artifactName: "lorem") {
                artifactName
                someValue
            }
        }
        componentBar: component(componentName: "bar") {
            componentName
            artifactIpsum: artifact(artifactName: "ipsum") {
                artifactName
                someValue
            }
        }
    }
}
```
Using `generate-client` generates the following classes:
```kotlin
// Bundle.kt
@Generated
public data class Bundle(
    public val bundleId: String,
    public val componentFoo: Component? = null,
    public val componentBar: Component? = null,
)
```

```kotlin
// Component.kt
@Generated
public data class Component(
    public val componentName: String,
    public val artifactLorem: Artifact? = null,
)
```

```kotlin
// Artifact.kt
@Generated
public data class Artifact(
    public val artifactName: String,
    public val someValue: String,
)
```
kotlin-graphql incorrectly determines that the requested fields for `componentFoo` and `componentBar` are the same and reuses the `Component` class it generated from `componentFoo` for `componentBar`. With these generated classes it is not possible to access the `artifactIpsum` field on the `componentBar` component.

**To Reproduce**
I created a [separate repository with a minimal reproduction project](https://github.com/WIStudent/graphql-kotlin-alias-issue-reproduction).

**Expected behavior**
kotlin-graphql should detect that `componentFoo` and `componentBar` query different fields and generate something like this:
```kotlin
// Bundle.kt
@Generated
public data class Bundle(
    public val bundleId: String,
    public val componentFoo: Component1? = null,
    public val componentBar: Component2? = null,
)
```

```kotlin
// Component1.kt
@Generated
public data class Component1(
    public val componentName: String,
    public val artifactLorem: Artifact? = null,
)
```

```kotlin
// Component2.kt
@Generated
public data class Component2(
    public val componentName: String,
    public val artifactIpsum: Artifact? = null,
)
```

```kotlin
// Artifact.kt
@Generated
public data class Artifact(
    public val artifactName: String,
    public val someValue: String,
)
```

WIStudent:
### :pencil: Description
See #1760 for a description of the problem this PR tries to fix.

I first created the new test case and then worked through the code to figure out where `ComplexObject` was wrongly deduplicated. `verifySelectionSet` in `generateTypeName.kt` came to the conclusion, that 
```gql
  first: complexObjectQuery {
    nameA: name
  }
```
and
```gql
  second: complexObjectQuery {
    nameB: name
  }
```
contain the same fields although they do not. Using `selection.alias` (if it exists) instead of `selection.name` in `calculateSelectedFields`  fixed the issue.

### :link: Related Issues
fixes #1760

### DIFF
--- a/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generateTypeName.kt
+++ b/plugins/client/graphql-kotlin-client-generator/src/main/kotlin/com/expediagroup/graphql/plugin/client/generator/types/generateTypeName.kt
@@ -229,9 +229,10 @@ private fun calculateSelectedFields(
     selectionSet.selections.forEach { selection ->
         when (selection) {
             is Field -> {
-                result.add(path + selection.name)
+                val fieldName = selection.alias ?: selection.name
+                result.add(path + fieldName)
                 if (selection.selectionSet != null) {
-                    result.addAll(calculateSelectedFields(context, targetType, selection.selectionSet, "$path${selection.name}."))
+                    result.addAll(calculateSelectedFields(context, targetType, selection.selectionSet, "$path$fieldName."))
                 }
             }
             is InlineFragment -> {
--- a/plugins/client/graphql-kotlin-client-generator/src/test/data/generator/alias_nested/AliasNestedQuery.graphql
+++ b/plugins/client/graphql-kotlin-client-generator/src/test/data/generator/alias_nested/AliasNestedQuery.graphql
@@ -1,0 +1,8 @@
+query AliasNestedQuery {
+  first: complexObjectQuery {
+    nameA: name
+  }
+  second: complexObjectQuery {
+    nameB: name
+  }
+}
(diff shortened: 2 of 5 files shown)


dariuszkuc (approved):

dariuszkuc:
Thanks for the fix!

PR #1764, opened 2023-04-29
docs: documentation for new GraalVm plugin tasks/goal

dariuszkuc:
### :pencil: Description

Add documentation on how to generate GraalVM native images using Gradle and Maven plugins.

Also fixed references to maven central (as it is now sonatype central) and updated readmes to start discussions over new issues.


dariuszkuc:
Large PR size is mostly due to the bump in dependencies for docusuaurus.

samuelAndalon (approved):

PR #1769, opened 2023-05-03
feat: GraalVM native image support for Spring server

dariuszkuc:
### :pencil: Description

Changes to enable GraalVM native image support for `graphql-kotlin-spring-server`.


plugins/graphql-kotlin-maven-plugin/src/main/kotlin/com/expediagroup/graphql/plugin/maven/GenerateGraalVmMetadataMojo.kt
```diff
@@ -34,7 +34,7 @@ class GenerateGraalVmMetadataMojo : AbstractSourceMojo() {
     override lateinit var outputDirectory: File
 
     override fun generate() {
-        val metadataDirectory = File(outputDirectory, "META-INF/native-image/${project.groupId}/${project.name}")
+        val metadataDirectory = File(outputDirectory, "META-INF/native-image/${project.groupId}/${project.name}/graphql")
```
dariuszkuc, line 37: GraalVM plugin looks for reachability metadata under any `META-INF/native-image` subdirectories. Per GraalVM documentation they suggest to generate reachability metadata under `<groupId>/<artifactId>` subdirectory to avoid collisions. Adding explicit `graphql` subdirectory to avoid collision with SpringBoot plugin generating metadata under the same directory.

servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLConfigurationProperties.kt
```diff
@@ -29,13 +30,21 @@ data class GraphQLConfigurationProperties(
     val packages: List<String>,
     /** Boolean flag indicating whether to print the schema after generator creates it */
     val printSchema: Boolean = false,
+    @NestedConfigurationProperty
```
dariuszkuc, line 33: required extra metadata to generate GraalVM native image ([docs](https://docs.spring.io/spring-boot/docs/current/reference/html/native-image.html#native-image.advanced.nested-configuration-properties))

integration/graalvm/build.gradle.kts
```diff
@@ -1,5 +1,10 @@
 import java.util.Properties
 
+@Suppress("DSL_SCOPE_VIOLATION") // TODO: remove once KTIJ-19369 / Gradle#22797 is fixed
+plugins {
+    alias(libs.plugins.graalvm.native) apply false
```
dariuszkuc, line 5: GraalVM native plugin dynamically generates some "helper" methods so we need to ensure they are generated just once for whole composite project, otherwise build blows up with classpath issues.

samuelAndalon (approved):

PR #1780, opened 2023-05-24
chore: update dependencies to the latest

dariuszkuc:


dariuszkuc:
* cannot update to Kotlin `1.8.21` as we are using `kotlin-compile-testing` which currently only supports `v1.8.0`
* didn't update ktlint to newer version as I was hitting some weird caching issues

samuelAndalon (approved):

Issue #1752, opened 2023-04-17
Support per-request BatchLoaderContextProviders in data loaders and free up BatchLoaderEnvironment key contexts


alex-lange:
**Is your feature request related to a problem? Please describe.**

Per the [graphql-java data loader docs](https://www.graphql-java.com/documentation/batching#passing-context-to-your-data-loader) : 
> The data loader library supports two types of context being passed to the batch loader. The first is an overall context object per dataloader, and the second is a map of per loaded key context objects.

The current `KotlinDataLoader` interface makes it difficult to provide per-request objects (e.g. the `GraphQLContext`) for the former "overall context".  We would usually define this `BatchLoaderContextProvider` when creating the data loader, but the `KotlinDataLoader#getDataLoader()` function (which instantiates the data loaders for each request) has no means of passing in per-request context to make the `BatchLoaderContextProvider` helpful. 

And while it is possible to define the latter "key context" objects when the data loader `load` function is called, the graphql-kotlin documentation recommends overloading this behavior by passing in the `GraphQLContext` and using the  [DataFetchingEnvironment#getGraphQLContext()](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/executions/graphql-kotlin-dataloader-instrumentation/src/main/kotlin/com/expediagroup/graphql/dataloader/instrumentation/extensions/BatchLoaderEnvironmentExtensions.kt#L43) extension function to access it. 

By overloading per-item key contexts with the `GraphQLContext`, we are unable to use _both_ types of contexts in our data loaders.

**Describe the solution you'd like**

Ideally the API would have a way to specify the `BatchLoaderContextProvider` (or something similar) that is automatically included in the `KotlinDataLoader#getDataLoader` call. This definition would probably trickle down from the `GraphQLRequestHandler#executeRequest`, which calls `KotlinDataLoaderRegistryFactory#generate`, which then calls `KotlinDataLoader#getDataLoader`.

If this were possible, we could use it to provide the `GraphQLContext` in the "overall context", and free up the "key context" for what it is intended to be used for.

**Describe alternatives you've considered**

[...]

PR #1785, opened 2023-06-01
feat: pass graphQLContext to KotlinDataLoaderRegistryFactory

samuelAndalon:
### :pencil: Description
passing graphQLContext instance from GraphQLRequestHandler to KotlinDataLoaderRegistryFactory.generate to allow DataLoaders to easily pass it as context of the BatchLoaderContext, and free up the context by key.

### :link: Related Issues
resolves https://github.com/ExpediaGroup/graphql-kotlin/issues/1752


smyrick:
You could take the Apollo Server approach and now that the graphql-java context is just a map, you could let library users save their own data loaders into the context and fetch them

samuelAndalon:
>You could take the Apollo Server approach and now that the graphql-java context is just a map, you could let library users save their own data loaders into the context and fetch them. Then you could just delete the KotlinDataLoader class altogether

internally the DataLoaderRegistry is actually stored in the GraphQLContext, thats how data fetchers/resolvers can access to it through the DataFetchingEnvironment, KotlinDataLoader is a convenient way to allow users to register data loaders and let the server do the work of creating the registry and adding them to the context.

PR #1800, opened 2023-06-16
feat(batching): avoid calculating document ast height in advance

samuelAndalon:
### :pencil: Description
Batching by LEVEL requires to calculate the height of all graphql documents in a batch request to be calculated before deciding of a certain level was dispatched, the reason is because a batch request could contain multiple and potentially different operations, with different levels, so, to decide if a level was dispatched, we need to know which operations in the batch request share the same level.

This PR will remove that logic, which was take it from the graphql-java [MaxDepthInstrumentation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/analysis/MaxQueryDepthInstrumentation.java#L56).

Instead we will calculate the height on runtime, simplifying the dispatch by level instrumentation and removing some CPU usage by not doing an AST traversal to calculate height.

### DIFF
--- a/executions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/level/DataLoaderLevelDispatchedInstrumentationTest.kt
+++ b/executions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/level/DataLoaderLevelDispatchedInstrumentationTest.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Expedia, Inc
+ * Copyright 2023 Expedia, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


tapaderster (approved):

executions/graphql-kotlin-dataloader-instrumentation/src/test/kotlin/com/expediagroup/graphql/dataloader/instrumentation/level/DataLoaderLevelDispatchedInstrumentationTest.kt
```diff
@@ -158,4 +158,8 @@ class DataLoaderLevelDispatchedInstrumentationTest {
             kotlinDataLoaderRegistry.dispatchAll()
         }
     }
+
+    companion object {
+        private const val PLUS_1_LEVEL = 1
```
tapaderster, line 163: nit: `PLUS_1_LEVEL` -> `ONE_LEVEL`

Issue #1640, opened 2023-01-17
[client] Allow Automatic Persisted Queries


samuelAndalon:
**Is your feature request related to a problem? Please describe.**
[graphql-kotlin-server recently added Automatic Persisted Queries](https://opensource.expediagroup.com/graphql-kotlin/docs/server/automatic-persisted-queries/#automaticpersistedqueriesprovider) to allow clients to send a hash representation of a graphQL operation instead of the actual query string.

**Describe the solution you'd like**
Allow graphql-kotlin-client to implement APQ requests

gumimin:
Hi @samuelAndalon 
I am thinking of implementing this Issue.
Please let me know your implementation plan if you don't mind.

1. Is the scope of this Issue to add an interface to graphql-kotlin-client? (is it unnecessary to add implementations to graphql-kotlin-ktor-client and graphql-kotlin-spring-client?)

2. Is it correct to the modifications are as follows.

Add `AutomaticallyPersistedQueriesSettings` to `com.expediagroup.graphql.client.types`

```kotlin
interface AutomaticPersistedQueriesSettings<T : Any> {
    val enabled: Boolean
    val httpMethodForHashedQueries: T
    val httpMethodForDocumentQueries: T
```

Add the `automaticPersistedQueriesSettings` property to `com.expediagroup.graphql.client.GraphQLClient`

```kotlin
interface GraphQLClient<RequestCustomizer> {
    val automaticPersistedQueriesSettings: AutomaticPersistedQueriesSettings<*>
    ...
}
```

Add the `extensions` property to `com.expediagroup.graphql.client.types.GraphQLClientRequest`

```kotlin
interface GraphQLClientRequest<T : Any> {
    ...
    val extensions: Any?
        get() = null
    ...
}
```

samuelAndalon:
Hello, the plan is to enable this from a flag, i think thats the only thing we can let clients to configure, regarding the hash, it has to be sha256 and pretty much follow this pattern

https://opensource.expediagroup.com/graphql-kotlin/docs/server/automatic-persisted-queries/

gumimin:
@samuelAndalon 
Thank you for your reply.
Is the implementation correct as below?
https://github.com/gumimin/graphql-kotlin/commit/b7d97144802c2d682f4491621149e00359c571c3

gumimin:
Hi @samuelAndalon 
I made a pull request, could you review it?
[...]

PR #1802, opened 2023-06-23
add automatic persisted queries implementation to client

gumimin:
### :pencil: Description
- add apq interfaces to client
- add apq implementation to ktor client
- add apq implementation to spring client

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1640


clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesExtension.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.client.types
+
+interface AutomaticPersistedQueriesExtension {
```
dariuszkuc, line 19: why is this an interface and not just a data class?

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesSettings.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.client.types
+
+interface AutomaticPersistedQueriesSettings {
```
dariuszkuc, line 19: why not data class?

clients/graphql-kotlin-ktor-client/src/main/kotlin/com/expediagroup/graphql/client/ktor/GraphQLKtorClient.kt
```diff
@@ -39,16 +48,52 @@ import java.net.URL
...
-            setBody(TextContent(serializer.serialize(request), ContentType.Application.Json))
-        }.body()
-        return serializer.deserialize(rawResult, request.responseType())
+        val queryId = request.getQueryId()
+        val automaticPersistedQueriesExtension = object : AutomaticPersistedQueriesExtension {
+            override val version: Int
+                get() = automaticPersistedQueriesSettings.version
+            override val sha256Hash: String
+                get() = queryId
+        }
```
dariuszkuc, line 62: this should only be computed IF using APQ

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesExtension.kt
gumimin: There were times during my implementation when it was more convenient to use interface, but now it is no longer necessary to use interface, so I'll change it to data class.

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesSettings.kt
gumimin: I'll also change this to data class.

gumimin:
@dariuszkuc 
Thank you for your reviewing!
I've fixed points you commented.

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesSettings.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.expediagroup.graphql.client.types
+
+data class AutomaticPersistedQueriesSettings(
```
samuelAndalon, line 19: APQ requests can also be send through HTTP POST, this implementation forces GET, maybe in the APQSetting we can specify which HTTP Verb the client wish to use.

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/extensions/AutomaticPersistedQueriesExtensions.kt
```diff
@@ -0,0 +1,19 @@
...
+
+fun GraphQLClientRequest<*>.getQueryId(): String =
+    String.format(
+        "%064x",
+        BigInteger(1, MESSAGE_DIGEST.digest(this.query.toByteArray(StandardCharsets.UTF_8)))
+    ).also {
+        MESSAGE_DIGEST.reset()
+    }
+
+fun AutomaticPersistedQueriesExtension.toQueryParamString() = """{"persistedQuery":{"version":$version,"sha256Hash":"$sha256Hash"}}"""
```
samuelAndalon, line 19: whereas POST can also have query params , when using GraphQL with POST you have to pass query, operationName and extensions in the payload -- i would just call it `serialize()`.

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesSettings.kt
```diff
@@ -0,0 +1,27 @@
...
+    val enabled: Boolean,
+    val version: Int
+)
+
+val defaultAutomaticPersistedQueriesSettings = AutomaticPersistedQueriesSettings(
+    enabled = false,
+    version = 1
```
samuelAndalon, line 26: For now i would just expose the `enabled` setting, not the version, other APQ implementations does not allow specifying the version as of now there only exists 1.

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesSettings.kt
gumimin: @samuelAndalon 
I thought APQ had to be GET.
[The Apollo documentation](https://www.apollographql.com/docs/apollo-server/performance/apq/
) was also written in GET.

For a POST request in GraphQL, [the query field in the body should be required](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md).
If APQ allows POST requests, then inevitably a large query string will be included in the request, I think which is inconsistent with the benefits of APQ.

Am I correct that the query field is optional when APQ is used in the POST method?
If so, I will change `interface GraphQLClientRequest.query` to nullable.
Also can you please share any detailed documentation on APQ specs (especially HTTP request and response)?

samuelAndalon: for APQ requests, the `query` field of the JSON payload is not required. 
graphql kotlin server allows APQ requests for both GET and POST, i think we should be consistent with the client.

samuelAndalon: >If APQ allows POST requests, then inevitably a large query string will be included in the request, I think which is inconsistent with the benefits of APQ.

APQ worst case scenario (when persited query is not found) needs to send a second request, with persitedQuery extension and query

samuelAndalon: https://opensource.expediagroup.com/graphql-kotlin/docs/server/automatic-persisted-queries/

dariuszkuc: AFAIK there is no official APQ spec yet and there are number of implementations out there (including the linked one from Apollo). Per the graphql http spec [discussion](https://github.com/graphql/graphql-over-http/issues/216) and [issue](https://github.com/graphql/graphql-over-http/issues/38) it is still TBD.

As for changing the `query` to be nullable -> I have no idea what sort of impact that would be, guessing it should be ok as it is a handled by the client internals.

samuelAndalon: The server logic itself should handle the nullable `query`, at least graphql-kotlin server does that.

if you dont send an APQ request (request without entension), and no query, then the server will return 400.

samuelAndalon: when added APQ in graphql-kotlin had to make query value default to ""

https://github.com/ExpediaGroup/graphql-kotlin/commit/f773327b126832c6176c0fd64e26aa6fea023d94#diff-b4c1162353124139fdbd7ef3c7f304d95203c1c7b8392ed9a153ed7a4c67f96eR42

samuelAndalon:
@gumimin can you take a look at the build failure, i think its coverage, we might need to add a test for GET and POST

gumimin:
@samuelAndalon 
I was aware of this but did not have time to implement. I will add more tests over the weekend, so please wait.

clients/graphql-kotlin-client/src/main/kotlin/com/expediagroup/graphql/client/types/AutomaticPersistedQueriesSettings.kt
gumimin: @samuelAndalon @dariuszkuc 
Thank you for your comment!
I've change to allow POST requests and `GraphQLClientRequest.query` to be nullable.

gumimin:
@samuelAndalon 
I've fixed the test, please review again.

samuelAndalon (approved):

samuelAndalon:
@gumimin Thank you for your contribution!

PR #1726, opened 2023-03-15
build: update to gradle8

dariuszkuc:

PR #1803, opened 2023-06-25
Don't Merge: Attempt to Update Gradle, Kotlin, and Libraries

murphye:
I attempted to make a contribution to your project by upgrading Kotlin, Gradle, and such. I went through and carefully updated the dependencies and made some other changes due to deprecations and broken APIs.

However, when it came to ktlint, that is where things went downhill. It seems that ktlint was never fully working in this project originally as there are potentially hundreds of formatting errors that don't comply with the ktlint standards. I don't know how you would want to fix this as there are a lot of formatting and naming anomalies in the code. You may just want to disable ktlint altogether. I tried to disable a few rules but there are just a lot of code formatting problems in general.

Also, the `graphql-kotlin-client-generator` Integration Tests are failing and there doesn't seem to be much in the way of helpful error messages. 

I hope you may find the work I did here helpful, but I feel that the project maintainers should take it from here. Thank you for your efforts with `graphql-kotlin`!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,5 +26,6 @@ ktlint_standard_import-ordering = disabled
 ktlint_standard_filename = disabled
 ktlint_standard_no-unused-parameters = disabled
 ktlint_standard_class-naming = disabled
+ktlint_standard_enum-naming = disabled
 ij_kotlin_allow_trailing_comma_on_call_site=false
 ij_kotlin_allow_trailing_comma=false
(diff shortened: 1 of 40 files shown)


dariuszkuc:
Thanks for the PR and looking into the ktlint stuff! I am unsure how much value we get out of ktlint so maybe it does makes sense to just drop it altogether.

re: gradle8 update -> I got the [PR#1726](https://github.com/ExpediaGroup/graphql-kotlin/pull/1726) opened already
re: kotlin update -> we are currently blocked on the update to 1.8.22 as we need a compatible version of https://github.com/tschuchortdev/kotlin-compile-testing lib (this is what is causing the failures in the client generator -> we use it in our unit tests to compile generated code and kotlin compiler version has to match kotlin version used in the project)

dariuszkuc:
superseded by https://github.com/ExpediaGroup/graphql-kotlin/pull/1822

Issue #1756, opened 2023-04-18
GraphiQL uses currently unsupported `graphql-transport-ws` protocol for subscriptions


jaaqo:
**Library Version**
7.0.0-alpha.5

**Describe the bug**
GraphiQL subscriptions do not work because of protocol mismatch.

GraphiQL uses [graphql-ws client](https://github.com/enisdenjo/graphql-ws) (_graphql-transport-ws_ protocol) for subscriptions.

Currently `graphql-kotlin` provides default server implementation for older [apollo client](https://github.com/apollographql/subscriptions-transport-ws) (_subscriptions-transport-ws_ protocol).

**To Reproduce**
Open default latest version Spring Server GraphiQL UI

**Expected behavior**
If user is using deprecated apollo protocol, initiate the graphiql client with compatible subscription client.

**Partial solution**
1) In [graphql-graphiql.html](ttps://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-server/src/main/resources/graphql-graphiql.html) add dependency for deprecated protocol client:
```html
<script src="https://cdn.jsdelivr.net/npm/subscriptions-transport-ws@0.11.0/browser/client.js"
        integrity="sha384-I1OLTTX4GzmfsCxgHUuqdP/zG05aDwlVdlx8Q3rDQNInfapBSnnJDgLuWZkj2tEH"
        crossorigin="anonymous"></script>
```
2. Then initiate fetchers using `legacyWsClient` option:
```javascript
var fetcher = GraphiQL.createFetcher({
  url: serverUrl,
  legacyWsClient: new SubscriptionsTransportWs.SubscriptionClient(subscriptionUrl)
});
```

dariuszkuc:
Added support for `graphql-transport-ws` to Spring in https://github.com/ExpediaGroup/graphql-kotlin/pull/1810. Legacy `subscription-transport-ws` protocol will be dropped in next major (i.e. v8).

PR #1804, opened 2023-06-26
Ktor Server:  set "graphql-transport-ws" as default protocol to graphQLSubscriptionsRoute

thevietto:
### :pencil: Description

This is a small fix to set `graphql-transport-ws` subprotocol as (sensible) default for Ktor server `graphQLSubscriptionsRoute`. 

Without that line, users may encounter issues when sending requests using popular GQL clients like Apollo or even bundled GraphiQL UI, because of they are sending `Sec-Websocket-Protocol=graphql-transport-ws` header and that is being rejected by Ktor.

Users may still override this value, if needed

### :link: Related Issues

#1756


dariuszkuc (approved):

dariuszkuc:
Thanks for the fix!

PR #1805, opened 2023-06-26
fix federation composition integration

dariuszkuc:
### :pencil: Description

Pin composition to a specific version. Latest `rover` changes require either specific version or to wrap major in quotes.


dariuszkuc:
CI is broken due to the above.

dariuszkuc:
Fixed in https://github.com/apollographql/rover/pull/1651

Issue #1807, opened 2023-06-29
Sub directory for queries folder client maven plugin


robp94:
### Discussed in https://github.com/ExpediaGroup/graphql-kotlin/discussions/1806

<div type='discussions-op-text'>

<sup>Originally posted by **robp94** June 29, 2023</sup>
Is it possible to have subdirectories in the queries folder for the client generator?

Currently, we use this as config `<queryFileDirectory>${project.basedir}/src/main/resources/clients/graphql/queries</queryFileDirectory>`

If the queries are in exactly this folder it is working, if we add a new folder inside the queries folder they are ignored.</div>

dariuszkuc:
Support added in #1809

PR #1809, opened 2023-06-30
support subdirectories in maven/gradle client plugins

robp94:
### :pencil: Description

Support subdirectories for query files in client maven plugin

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1807


dariuszkuc:
@samuelAndalon if we support this for Maven I believe we also should automatically support this for Gradle? Technically you can also just script it out in your build file but I think we should have consistent behavior between plugins.

robp94:
I added it for gradle as well.

samuelAndalon (approved):

PR #1816, opened 2023-07-27
BREAKING(feat(federation)): full `@link` support

dariuszkuc:
### :pencil: Description

Introduce full `@link` support that includes namespacing and renaming of the imported elements.

The `@link` directive allows users to link definitions within the document to external schemas. It is the core feature of the Apollo federation v2. While imported elements still need to have their definitions in the local schema, `@link` allows users to namespace and/or rename those items to avoid any local type conflicts. By default, all external types that are not explicitly imported have to be namespaced using the spec name or a provided custom namespace.

Updated `@link` definition

```graphql
directive @link(url: String!, as: String, import: [link__Import]) repeatable on SCHEMA

scalar link__Import
```

* url - external specification url
* as - optional custom namespace
* import - list of elements to import, can either be simple Strings (e.g. `@key`) or custom imports that rename elements (e.g. `{ name: "@key", as: "@myKey" }`)

By default, `graphql-kotlin` will continue to apply `@link` directive using latest supported federation specification but will only auto-import federation specific directives up to version 2.3 and only if they are present (i.e. applied to an element) in the schema. All new fed v2.4+ won't be included in the auto-imports and instead will be namespaced with the spec name, e.g. `@federation__authenticated`.

Users can provide custom `@link` information by providing a schema object with `@LinkDirective` information, e.g.

```kotlin
@LinkDirective(url = "https://specs.apollo.dev/federation/v2.3", `as`: "fed", import = [LinkImport(name = "@key", `as` = "@myKey"), LinkImport(name = "@requires")])
class MyCustomLinkSchema
```

Will generate following schema

```graphql
schema @link(as: "fed", import : [{name : "@key", as : "@myKey"}, "@requires"], url : "https://specs.apollo.dev/federation/v2.3"){
  query: Query
}

// directive imported with custom name
"Space separated list of primary keys needed to access federated object"
directive @myKey(fields: fed__FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE

// directive imported with same name
"Specifies required input field set from the base type for a resolver"
directive @requires(fields: fed__FieldSet!) on FIELD_DEFINITION

// type imported with custom namespace
"Federation type representing set of fields"
scalar fed__FieldSet
```

When importing custom specifications, in order to be able to identify whether element is part of custom specification, we need to annotate it with new `@LinkedSpec` annotation. All federation directives were updated to rely on this mechanism, e.g.

```kotlin
@LinkedSpec(FEDERATION_SPEC)
@Repeatable
@GraphQLDirective(
    name = KEY_DIRECTIVE_NAME,
    description = KEY_DIRECTIVE_DESCRIPTION,
    locations = [DirectiveLocation.OBJECT, DirectiveLocation.INTERFACE]
)
annotation class KeyDirective(val fields: FieldSet, val resolvable: Boolean = true)
```


gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ federation = "3.0.1"
 graphql-java = "20.2"
 graalvm = "0.9.21"
 jackson = "2.15.0"
-kotlin = "1.7.21"
+kotlin = "1.8.21"
```
dariuszkuc, line 9: there is a reflection bug in 1.7 that causes runtime failures when trying to access annotation array elements :(

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateDirective.kt
```diff
@@ -66,43 +69,51 @@ internal fun generateEnumValueDirectives(generator: SchemaGenerator, field: Fiel
...
-                        val value = prop.call(directiveInfo.directive)
-                        argumentBuilder.valueProgrammatic(value)
-                    }
-                    ?.let { appliedDirectiveArgument ->
-                        builder.argument(appliedDirectiveArgument)
+                val argumentToBeModified = directive.getArgument(prop.name)
+                if (argumentToBeModified != null) {
+                    val value = prop.call(directiveInfo.directive)
+                    // annotations cannot have null values, we should skip known defaults
+                    if (value != DEFAULT_DIRECTIVE_STRING_VALUE && value != DEFAULT_DIRECTIVE_INT_VALUE) {
```
dariuszkuc, line 106: Currently we don't have mechanism for skipping the default (i.e., not provided) values. Since we cannot use `null` in annotations we should have some explicit (unlikely) defaults.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/SchemaGenerator.kt
```diff
@@ -70,24 +70,21 @@ open class SchemaGenerator(internal val config: SchemaGeneratorConfig) : Closeab
...
 
-        return GraphQLSchema.newSchema()
-            .query(generateQueries(this, queries))
-            .mutation(generateMutations(this, mutations))
-            .subscription(generateSubscriptions(this, subscriptions))
-            .additionalTypes(generateAdditionalTypes())
-            .additionalDirectives(directives.values.toSet())
-            .codeRegistry(codeRegistry.build())
+        return config.hooks.willBuildSchema(queries, mutations, subscriptions, additionalTypes, additionalInputTypes, schemaObject)
             .also { builder ->
```
dariuszkuc, line 74: instead of applying this logic at the end which required clearing up the directives and manually re-adding generated stuff -> we could just move it up front

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/hooks/SchemaGeneratorHooks.kt
```diff
@@ -50,11 +51,24 @@ import kotlin.reflect.KType
  * that allow users to customize the schema.
  */
 interface SchemaGeneratorHooks {
+
+    /**
+     * Called before any generation of GraphQL schema.
+     *
+     * Allows users to do some preprocessing and generate custom schema builder instance.
+     */
+    fun willBuildSchema(queries: List<TopLevelObject>,
```
dariuszkuc, line 60: `willBuildSchema` was executed AFTER we already build it, since I need to execute logic BEFORE we start building anything I standardized the hooks as `willBuildSchema -> BUILD -> didBuildSchema`

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/internal/types/generateDirective.kt
```diff
@@ -66,43 +69,51 @@ internal fun generateEnumValueDirectives(generator: SchemaGenerator, field: Fiel
...
     val directiveName = directiveInfo.effectiveName
-    val directiveFromHook = generator.config.hooks.willGenerateDirective(directiveInfo)
-        ?.let { generator.directives.putIfAbsent(directiveName, it) }
-
-    val directive = directiveFromHook ?: generator.directives.computeIfAbsent(directiveName) {
-        val builder = GraphQLDirective.newDirective()
-            .name(directiveInfo.effectiveName)
-            .description(directiveInfo.directiveAnnotation.description)
-            .repeatable(directiveInfo.repeatable)
+    val directive = generator.directives.computeIfAbsent(directiveName) {
```
dariuszkuc, line 72: previously we were always executing the `willGenerateDirective` hook instead of actually reading the previously generated value

dariuszkuc: fixed in https://github.com/ExpediaGroup/graphql-kotlin/pull/1818

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/hooks/SchemaGeneratorHooks.kt
dariuszkuc: fixed in https://github.com/ExpediaGroup/graphql-kotlin/pull/1817

samuelAndalon (approved):

PR #1822, opened 2023-08-23
update to Kotlin 1.8

dariuszkuc:
### :pencil: Description

Update Kotlin version to 1.8.22

### DIFF
--- a/plugins/client/graphql-kotlin-client-generator/build.gradle.kts
+++ b/plugins/client/graphql-kotlin-client-generator/build.gradle.kts
@@ -14,17 +14,25 @@ dependencies {
     implementation(libs.jackson)
     implementation(libs.ktor.client.apache)
     implementation(libs.ktor.serialization.jackson) {
-        exclude("com.fasterxml.jackson.core", "jackson-databind")
-        exclude("com.fasterxml.jackson.module", "jackson-module-kotlin")
+        exclude(group = "com.fasterxml.jackson.core", module = "jackson-databind")
+        exclude(group = "com.fasterxml.jackson.module", module = "jackson-module-kotlin")
     }
     implementation(libs.ktor.client.content)
     implementation(libs.slf4j)
     testImplementation(projects.graphqlKotlinClientJackson)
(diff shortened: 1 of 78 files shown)


plugins/client/graphql-kotlin-client-generator/build.gradle.kts
```diff
@@ -14,17 +14,25 @@ dependencies {
...
+        exclude(group = "com.fasterxml.jackson.module", module = "jackson-module-kotlin")
     }
     implementation(libs.ktor.client.content)
     implementation(libs.slf4j)
     testImplementation(projects.graphqlKotlinClientJackson)
     testImplementation(projects.graphqlKotlinClientSerialization)
     testImplementation(libs.wiremock.jre8)
-    testImplementation(libs.compile.testing)
+    testImplementation(libs.compile.testing) {
+        // there is no kotlin compile testing release supporting kotlin 1.8.22
```
dariuszkuc, line 26: This is a workaround that should be removed once compatible version is available. Currently there is no release of `kotlin-compile-testing` that is compatible with kotlin version 1.8.22 :( 

- kotlin 1.8 compatible version is broken for compiler plugins (which we use) due to missing argument
- we need version compatible with kotlin 1.8.20+ that includes a fix for a reflection bug
- kotlin 1.9 compatible version fixes above issues .... but `kotlin-serialization` JARs are no longer published with shaded dependencies and we can no longer resolve the `SerializationComponentRegistrar`

Since there were no breaking API changes in the compile-testing library between the 1.8.22 and 1.9.0 versions, the simplest workaround for now is to explicitly exclude transitive compiler dependencies and explicitly include compatible versions.

dariuszkuc:
Large number of changed files is due to `kotlin-poet` change to omit `public` modifier on overridden properties

samuelAndalon (approved):

dariuszkuc:
Note: we need to use Kotlin 1.8.20+ due to https://youtrack.jetbrains.com/issue/KT-53279

Issue #1831, opened 2023-08-31
graphql-response+json content type is the default


gumimin:
**Library Version**
[7.0.0-alpha.5](https://github.com/ExpediaGroup/graphql-kotlin/releases/tag/7.0.0-alpha.5)

**Describe the bug**
I implemented this PR: https://github.com/ExpediaGroup/graphql-kotlin/pull/1699, Issue: https://github.com/ExpediaGroup/graphql-kotlin/issues/1573

I fixed this according to following review comment.
https://github.com/ExpediaGroup/graphql-kotlin/pull/1699#discussion_r1127131395

However, the `*/*` was not taken into account and the change was not sufficient.
Most clients will request with `*/*` unless an any MediaType is specified, so this needs to be fixed.

**To Reproduce**
When a request is send without accept header, the response content type is returned as `application/graphql-response+json`

**Expected behavior**
I want to keep `application/json` as default instead of `application/graphql-response+json`
https://github.com/ExpediaGroup/graphql-kotlin/pull/1699#discussion_r1127131395

dariuszkuc:
Fixed in #1832

PR #1699, opened 2023-03-06
add support for application/graphql-response+json content type

gumimin:
### :pencil: Description
Based on the accept header, spring server respond either with application/json or application/graphql-response+json (default) media type.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1573

Issue #1573, opened 2022-10-10
[spring-server] support graphql-response+json content type


dariuszkuc:
**Is your feature request related to a problem? Please describe.**
Per [GraphQL over HTTP](https://github.com/graphql/graphql-over-http) spec, servers should default to  `application/graphql-response+json` media type for responses.

**Describe the solution you'd like**
Based on the accept header, spring server should respond either with `application/json` or `application/graphql-response+json` (default) media type.

**Describe alternatives you've considered**
N/A

**Additional context**
`application/graphql-response+json` should be optionally supported in the v7 and then default to it in the subsequent v8 release.

dariuszkuc:
Support added in https://github.com/ExpediaGroup/graphql-kotlin/pull/1699

PR #1832, opened 2023-08-31
Keep response content type to application/json as default instead of application/graphql-response+json

gumimin:
### :pencil: Description
Most clients will request with `*/*` unless an any MediaType is specified.
However, the `*/*` was not taken into account in current implementation.

I change to keep `application/json` as default instead of `application/graphql-response+json`.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1831
https://github.com/ExpediaGroup/graphql-kotlin/issues/1573
https://github.com/ExpediaGroup/graphql-kotlin/pull/1699


dariuszkuc (approved):

dariuszkuc:
Thanks for the fix!

PR #1842, opened 2023-09-12
fix: add explicit dependency between kapt and dokka

dariuszkuc:


dariuszkuc:
Last release failed due to missing dependency between tasks.

samuelAndalon (approved):

PR #1872, opened 2023-10-23
Rewrite CompletableFutureExtension.allSettled to not use CompletableFuture.supplyAsync

roookeee:
We noticed a big drop in performance after switching to the new `CompletableFuture` based `FederatedTypeResolver`. Our investigations led us to the implementation of `CompletableFutureExtension.allSettled` which has the following issues:

- Bug: It uses `CompletableFuture.supplyAsync` which looses all thread local data (as we switch to another thread pool we have no control over), which breaks most tracing frameworks and all other thread local usages
- Performance: 100 futures get wrapped into 100 separately scheduled futures, this is unnecessary as `thenApply` and `exceptionally` work without nesting of futures. This avoids non-neglegible overhead

I changed the accompanying test to expect another exception type as only `get()` throw a `ExecutionException`. This should not affect your public API as its an internal method and I could not find any other references to `ExecutionException` in `graphql-kotlin`s codebase.

### DIFF
--- a/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
+++ b/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
@@ -42,18 +42,9 @@ internal fun <T : Any?> List<CompletableFuture<out T>>.joinAll(): CompletableFut
 @Suppress("TooGenericExceptionCaught")
 internal fun <T : Any?> List<CompletableFuture<out T>>.allSettled(): CompletableFuture<List<Result<T>>> {
     val resultFutures = map { future ->
-        CompletableFuture.supplyAsync {
-            try {
-                Result.success(future.get())
-            } catch (e: Exception) {
-                Result.failure<T>(e)
-            }
-        }
+        future.thenApply { result -> Result.success(result) }
+            .exceptionally { exception -> Result.failure<T>(exception) }
     }
 
-    return CompletableFuture.allOf(
-        *resultFutures.toTypedArray()
-    ).thenApply {
-        resultFutures.map(CompletableFuture<Result<T>>::join)
-    }
+    return resultFutures.joinAll()
 }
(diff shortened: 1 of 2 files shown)


samuelAndalon (approved):

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt
```diff
@@ -42,18 +42,9 @@ internal fun <T : Any?> List<CompletableFuture<out T>>.joinAll(): CompletableFut
...
 internal fun <T : Any?> List<CompletableFuture<out T>>.allSettled(): CompletableFuture<List<Result<T>>> {
     val resultFutures = map { future ->
-        CompletableFuture.supplyAsync {
-            try {
-                Result.success(future.get())
-            } catch (e: Exception) {
-                Result.failure<T>(e)
-            }
-        }
+        future.thenApply { result -> Result.success(result)}
```
dariuszkuc, line 45: this fails ktlint
```
/home/runner/work/graphql-kotlin/graphql-kotlin/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/extensions/CompletableFutureExtensions.kt:45:60 Missing spacing before "}"
```

roookeee: Ah an oversight, will get to it tomorrow!

roookeee: Done

roookeee:
Could we get a bugfix release that includes this change in the near future? It's currently blocking our prod migration to graphql-kotlin 7.x because of performance regressions. Thank you!

dariuszkuc:
I'll try getting two federation fixes in over the weekend (unsure why builds time out on those) and we should be able to release afterwards.

Issue #1245, opened 2021-08-19
Reuse fragments in different Queries to generate shared data model


reneschroeder0000:
**Is your feature request related to a problem? Please describe.**
I would like to define a fragment that is used across multiple queries. E.g. querying entities with `findAll(): List<E>` and `findById(id: ID): E` and subsequently processing the entities. Currently this generates multiple data classes (E1, E2) as fragments cannot be re-used.

**Describe the solution you'd like**
I would like to use a fragment across multiple queries to only generate a single data class and simplify subsequent processing. Result: `findAll(): List<E>` and `findById(id: ID): E` respectively

**Describe alternatives you've considered**
Maybe it would be easier if more than 1 Query would be allowed to be defined within the .graphql file. My current workaround is a custom mapping from one generated type to the other  which makes data changes cumbersome.

joffrey-bion:
Is there any attention given to this? It is quite critical for me, at the moment the types generated for all queries are completely unrelated, even if they are the exact same type in the schema. I cannot make use of a common interface or anything like that, which is very very cumbersome.

matthiasdg:
Any roadmap or requirements for this? Would be unfortunate if the work of @robp94 gets lost.
Personally I'm just requesting multiple query fields in one go, also getting duplicate classes for some type fields (maybe because I have to deal with polymorphism in one case and not the other), so initially I was thinking that I maybe could create a fragment and reuse that. The query looks nicer now, but it sure would be nice to have a reusable fragment in code too!

altair8800:
It's working for me. Just define a fragment and consume in multiple queries in one file -> only one fragment model class generated.

robp94:
@altair8800 I do not understand how this should work. I still get this exception if I try to have more than one query per file.
`Execution default of goal com.expediagroup:graphql-kotlin-maven-plugin:8.2.1:generate-client failed: GraphQL client does not support query files with multiple operations`

altair8800:
@robp94 my bad - I was using Apollo Kotlin *d'oh*

PR #1890, opened 2023-11-22
Support for shared fragments

robp94:
### :pencil: Description
Support a shared fragments file to avoid duplicate file generation.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1245


dariuszkuc:
Since each GraphQL file can contain fragments (which may or may not overlap), I believe the solution should be accounting for this. Otherwise this would force clients to potentially rewrite all their queries.

Efficiently generating client code is a pretty complex topic and there are number of considerations and tradeoffs with the approaches. Martin Bonnin did a great write up on this topic in how they approach this problem in `apollo-kotlin` (see [link](https://www.apollographql.com/blog/mobile/kotlin/the-story-behind-apollo-kotlin-3-codegen/)).

PR #1918, opened 2024-01-25
feat: pass ExecutionInput to getOrElse method

samuelAndalon:
### :pencil: Description
trying to implement a distributed cache mechanism for parsed and validated operations, in the `getOrElse` method is where we would be setting the operation into an external cache, for that, we need access to the actual operation, which is included in the `ExecutionInput`.

No more changes are needed given that the action of storing the parsed and validated operation is potentially async, but we don't care about the result.


tapaderster:
can you please add a description with the reason for this?

tapaderster (approved):

Issue #1920, opened 2024-01-29
GraphQL Ktor plugin swallows context exceptions.


dakriy:
**Library Version**
7.0.2

**Describe the bug**
When building a custom graphql context that throws an exception, the exception gets completely swallowed by the ktor graphql server plugin making it hard to figure out what the issue is.

**To Reproduce**
Make a custom context builder that throws an exception and visit the graphql endpoint. It will return a 400 bad request with no information anywhere.

**Expected behavior**
The exception should be thrown so that it can be handled by the ktor exception handler.

curtiscook:
Hi, @smyrick or @dariuszkuc, any word on which release version this change will ship with?

smyrick:
@curtiscook I am not a maintainer of the package anymore. You can reach out to the contact info provided.

Also it was indicated in the PR that the release would happen after graphql-java: https://github.com/ExpediaGroup/graphql-kotlin/pull/1936#issuecomment-2047977674

curtiscook:
> @curtiscook I am not a maintainer of the package anymore. You can reach out to the contact info provided in the repo
> 
> Also it was indicated in the PR that the release would happen after graphql-java: [#1936 (comment)](https://github.com/ExpediaGroup/graphql-kotlin/pull/1936#issuecomment-2047977674)

Oops! Thanks, sorry I missed that

curtiscook:
> Not just Ktor - graphql-kotlin-server does this, which means that Spring and custom server implementations are also affected: https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt#L112

I ran out of time to fix this, but would be nice if someone else could pick it up

ujjawal-yadav:
I tried reproducing this on current `master` and got a slightly different picture, so posting in case it's useful.

The original repro doesn't seem to happen anymore. A `GraphQLContextFactory` that throws propagates out of `GraphQLServer.execute()` fine, there's no try/catch around `contextFactory.generateContext(request)`, so it reaches the server layer rather than being swallowed.

[...]

PR #1921, opened 2024-01-29
Don't swallow root server exceptions

dakriy:
Let root level bootstrap exceptions go to the ktor exception handler rather than being swallowed.

### :pencil: Description
Exceptions that happen when building the context are swallowed obscuring what actually went wrong and making it hard to fix.

### :link: Related Issues
#1920


curtiscook:
Hi @samuelAndalon can we get this merged?

dariuszkuc:
Server should not respond with exception stacktraces. Exceptions should be logged instead.

curtiscook:
> Server should not respond with exception stacktraces. Exceptions should be logged instead.

I don't understand your point here and it doesn't read constructively. 

Currently, any exception throws a 400 error with no message and I have no control over the error code or messaging. 

As you know... Some errors are 400, some are 401, 403, 500, or 503. Technically, since you're catching a base exception, it should be a 500 error. If you look through all the code it looks like maybe people should be catching errors and rethrowing a GraphQLError that can provide a message, but I'm still not seeing where I can set the status code. 

Honestly, the "correct" way to handle this should be to let KTor (or Spring) handle the exception and map any unhandled exceptions to a 500 status code, which could easily be part of the example. 

Or at the very least, I think there could be some documentation on suggested error handling.

dariuszkuc:
Superseded by https://github.com/ExpediaGroup/graphql-kotlin/pull/1936

PR #1937, opened 2024-03-11
feat: kotlinx serialization for GraphQLServerRequest

samuelAndalon:
### :pencil: Description
Switch from `jackson` to `kotlinx.serialization` for serialization/deserialization of `GraphQLServerRequest` types.

After running benchmarks we where able to identify that deserializing `GraphQLServerRequest` with `kotlinx.serialization` is quite faster than doing it with `jackson`, the reason ? possibly because jackson relies on reflections to identify deserialization process.

On the other hand, serialization/deserialization of `GraphQLServerReponse` type is still faster if done with `jackson`, possibly because of how `kotlinx.serialization` library was designed and the poor support for serializing `Any` type: 
https://github.com/Kotlin/kotlinx.serialization/issues/296, which causes a lot of memory comsumption.

As part of this PR also including the benchmarks. For that, i created a separate set of types that are marked with both `jackson` and `kotlinx.serialization` annotations.

Benchmarks results:
Executed on a MacBookPro 2.6 GHz 6-Core Intel Core i7.

#### GraphQLServerRequest Deserialization

`GraphQLBatchRequest`
4 batched operations, each operation is aprox: 30kb
image

`GraphQLRequest`
image

#### GraphQLServerResponse Serialization

`GraphQLBatchResponse`
image

`GraphQLResponse`
image


servers/graphql-kotlin-server/src/benchmarks/kotlin/testtypes/GraphQLServerError.kt
```diff
@@ -0,0 +1,38 @@
...
+package com.expediagroup.graphql.server.testtypes
+
+import com.expediagroup.graphql.server.types.serializers.AnyNullableKSerializer
+import com.expediagroup.graphql.server.types.serializers.GraphQLErrorPathKSerializer
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonInclude
+import kotlinx.serialization.Serializable
+
+/**
+ * GraphQL error representation that is spec complaint with serialization and deserialization.
```
tapaderster, line 26:
```suggestion
 * GraphQL error representation that is spec-compliant with serialization and deserialization.
```

samuelAndalon:
Important to mention that Ktor server will still use jackson as engine for deserialization, while spring server will use kotlinx.
In a separate PR we can make this configur

tapaderster (approved):

PR #1948, opened 2024-04-08
Fix links in documentation and add some additional info for DataFetchingEnvironment selectionSet

mitjade:
Fixed links and added info for `selectionSet`. It took me a while to find how to retrieve sub field information, so I figures it might help the next guy.

### DIFF
--- a/website/docs/schema-generator/execution/data-fetching-environment.md
+++ b/website/docs/schema-generator/execution/data-fetching-environment.md
@@ -44,7 +44,7 @@ Then the following query would return `"The parentField was foo and the childFie
 ```
 
 You can also use this to retrieve arguments and query information from higher up the query chain. You can see a working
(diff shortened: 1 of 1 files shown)


website/docs/schema-generator/execution/data-fetching-environment.md
```diff
@@ -44,4 +44,30 @@ Then the following query would return `"The parentField was foo and the childFie
...
+{
+  product {
+    id
+    title
+    price
+  }
+}
+```
+
+You can also use `selectionSet` to retrieve sub field information a client wants. It can be useful to know
```
samuelAndalon, line 71:
```suggestion
You can also use `selectionSet` to access the selected fields of the current field. It can be useful to know which selections have been requested so the data fetcher can optimize the data access queries. For example, in an SQL-backed system, the data fetcher can access the database and use the field selection criteria to specifically retrieve only the columns that have been requested by the client.
```

samuelAndalon (approved):

PR #1960, opened 2024-04-17
feat: ability to specify `@GraphQLName` on input types w/o suffix

dariuszkuc:
### :pencil: Description

Currently input types are always suffixed with `Input` which is problematic as there are use cases where users might want to provide different custom name. This change adds check whether specified type is input only and if thats the case it does not attempt to add `Input` suffix.

### :link: Related Issues
Supersedes https://github.com/ExpediaGroup/graphql-kotlin/pull/1949


dariuszkuc:
While technically it is possible that folks used valid location = input + custom name, this seems rather unlikely case (as they would need to explicitly specify `Input` suffix OR rely on auto suffix).

agarwalaman786:
I am having one [question](https://github.com/ExpediaGroup/graphql-kotlin/pull/1949#issuecomment-2060328906) on this.

samuelAndalon (approved):

PR #1966, opened 2024-04-22
1965: make `FunctionDataFetcher` handle blocking scopes resulting in a thrown `Error` as expected

### LINKED ISSUE #1965: FunctionDataFetcher: does not handle `Error` descendant `Throwables` in `runBlockingFunction` in conjunction with GraphQL Java `graphql/execution/ExecutionStrategy.java`
**Library Version**

Latest [7.1.1](https://github.com/ExpediaGroup/graphql-kotlin/releases/tag/7.1.1)

**Describe the bug**

The way this library interacts with `graphql-java`, any non-blocking resolvers that lead to a result of a thrown `Throwable` descendant from `Error` and not `Exception` will not be handled as expected in the `errors` field of the response, but rather result in a `HTTP - 500` error (depending on specific serve implementation).

So like this example here:

https://github.com/ExpediaGroup/graphql-kotlin/blob/862460e26e8dc6ddf3a4265b51a06b153c21b63b/website/docs/schema-generator/execution/fetching-data.md#L39-L45

If you instead had this (`fun` instead of `suspend fun`):

```kotlin
class Product(val id: ID, val name: String) {
  fun reviews(): List<Reviews> { ... }
}
```

You run into the problem where this will be handled as expected:

```kotlin
class Product(val id: ID, val name: String) {
  fun reviews(): List<Reviews> { throw Exception("it works as expected; this ends up in errors") }
}
```

but this will not:

```kotlin
class Product(val id: ID, val name: String) {
  fun reviews(): List<Reviews> { throw Error("it works as expected; this ends up in errors") }
}
```

This is because of the `graphql-java` parent scope that does catch any `Exception` but not `Error` descendant `Throwable`s:

https://github.com/graphql-java/graphql-java/blob/25667a13e528a2b2c4fa372b363ffbb00184316a/src/main/java/graphql/execution/ExecutionStrategy.java#L322-L339.

In my case a nested scope throwing the `kotlin.NotImplementedError` brought this to my attention.

**To Reproduce**

No particular schema/config needed, just have the following:

```kotlin
@Component
class FooQuery: Query {
    fun getFoo(): Foo {
        return Foo("hello world",)
    }
}
```

```kotlin
data class Foo(val bar: String) {
    fun nonSuspendFunctionException(): String {
        throw Exception("This is an exception - and works fine")
    }

    fun nonSuspendFunctionError(): String {
        throw Error("This is an error - and DOESN'T work fine")
    }

    suspend fun suspendFunctionException(): String {
        throw Exception("This is an exception - and works fine")
    }

    suspend fun suspendFunctionError(): String {
        throw Error("This is an error - and works fine")
    }
}
```

Attached is a minimal Spring Boot app you can run locally and easily observe the behavior using following query and uncommenting the various resolvers:

```graphql
query Reproduction {
  getFoo {
    # nonSuspendFunctionException
    # nonSuspendFunctionError
    # suspendFunctionException
    # suspendFunctionError
  }
}
```

[graphql-kotlin-error-example.zip](https://github.com/ExpediaGroup/graphql-kotlin/files/15067207/graphql-kotlin-error-example.zip)

**Expected behavior**

The expected `errors` handling like this:

image

Instead of the current behavior which is an unhandled server error like this:

image

wallind:
### :pencil: Description

Changes `FunctionDataFetcher` to always return a `CompletableFuture` (no matter if scope is `suspend` or not) in order to cleanly handle cases where `runBlockingFunction` results in a thrown `Error` (or descendant of `Error`). See #1965 for more detail.

### :link: Related Issues

Resolves #1965 

*_this is my first contribution to this repo; so if I need to do anything differently or in addition please let me know_

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
@@ -36,8 +36,8 @@ import kotlin.reflect.full.valueParameters
 /**
  * Simple DataFetcher that invokes target function on the given object.
  *
- * @param target The target object that performs the data fetching, if not specified then this data fetcher will attempt
- *   to use source object from the environment
+ * @param target The target object that performs the data fetching, if not specified, then this data fetcher will attempt
+ *   to use a source object from the environment
  * @param fn The Kotlin function being invoked
  */
 @Suppress("Detekt.SpreadOperator")
@@ -49,7 +49,7 @@ open class FunctionDataFetcher(
     /**
      * Invoke a suspend function or blocking function, passing in the [target] if not null or default to using the source from the environment.
      */
-    override fun get(environment: DataFetchingEnvironment): Any? {
+    override fun get(environment: DataFetchingEnvironment): CompletableFuture<Any?> {
         val instance: Any? = target ?: environment.getSource<Any?>()
         val instanceParameter = fn.instanceParameter
 
@@ -63,7 +63,7 @@ open class FunctionDataFetcher(
                 runBlockingFunction(parameterValues)
             }
         } else {
-            null
+            CompletableFuture.completedFuture(null)
         }
     }
 
@@ -82,10 +82,10 @@ open class FunctionDataFetcher(
      * Retrieves the provided parameter value in the operation input to pass to the function to execute.
      *
      * If the argument is missing in the input, and the type is not an [OptionalInput], do not return a mapping.
-     * This allows for default Kotlin values to be used when executing the function. If you need logic when a value
+     * This allows for default Kotlin values to be used when executing the function. Otherwise, if you need logic when a value
      * is missing, use the [OptionalInput] wrapper instead.
      *
-     * If the parameter is of a special type then we do not read the input and instead just pass on that value.
+     * If the parameter is of a special type, then we do not read the input and instead just pass on that value.
      * The special values include:
      *   - The entire environment is returned if the parameter is of type [DataFetchingEnvironment]
      */
@@ -120,12 +120,15 @@ open class FunctionDataFetcher(
         }
 
     /**
-     * Once all parameters values are properly converted, this function will be called to run a simple blocking function.
-     * If you need to override the exception handling you can override this method.
+     * Once all parameter values are properly converted, this function will be called to run a simple blocking function.
+     * If you need to override the exception handling, you can override this method.
      */
-    protected open fun runBlockingFunction(parameterValues: Map<KParameter, Any?>): Any? = try {
-        fn.callBy(parameterValues)
-    } catch (exception: InvocationTargetException) {
-        throw exception.cause ?: exception
-    }
+    protected open fun runBlockingFunction(parameterValues: Map<KParameter, Any?>): CompletableFuture<Any?>  =
+        CompletableFuture<Any?>().apply {
+            try {
+                complete(fn.callBy(parameterValues))
+            } catch (exception: InvocationTargetException) {
+                completeExceptionally(exception.cause ?: exception)
+            }
+        }
 }
--- a/generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcherTest.kt
+++ b/generator/graphql-kotlin-schema-generator/src/test/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcherTest.kt
@@ -24,7 +24,6 @@ import io.mockk.every
 import io.mockk.mockk
 import kotlinx.coroutines.coroutineScope
 import org.junit.jupiter.api.Test
-import java.util.concurrent.CompletableFuture
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertNotNull
@@ -91,7 +90,7 @@ class FunctionDataFetcherTest {
         val mockEnvironment: DataFetchingEnvironment = mockk {
             every { getSource<Any>() } returns null
         }
-        assertNull(dataFetcher.get(mockEnvironment))
+        assertNull(dataFetcher.get(mockEnvironment).get())
     }
 
     @Test
@@ -102,7 +101,7 @@ class FunctionDataFetcherTest {
             every { arguments } returns mapOf("string" to "hello")
             every { containsArgument("string") } returns true
         }
-        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment))
+        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment).get())
     }
 
     @Test
@@ -112,7 +111,7 @@ class FunctionDataFetcherTest {
             every { arguments } returns mapOf("string" to "hello")
             every { containsArgument("string") } returns true
         }
-        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment))
+        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment).get())
     }
 
     @Test
@@ -127,7 +126,7 @@ class FunctionDataFetcherTest {
             every { arguments } returns emptyMap()
             every { containsArgument(any()) } returns false
         }
-        assertEquals(expected = "foo", actual = dataFetcher.get(mockEnvironment))
+        assertEquals(expected = "foo", actual = dataFetcher.get(mockEnvironment).get())
     }
 
     @Test
@@ -138,7 +137,7 @@ class FunctionDataFetcherTest {
             every { arguments } returns mapOf("string" to "hello")
             every { containsArgument("string") } returns true
         }
-        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment))
+        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment).get())
     }
 
     @Test
@@ -149,7 +148,7 @@ class FunctionDataFetcherTest {
             every { arguments } returns emptyMap()
             every { containsArgument(any()) } returns false
         }
-        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment))
+        assertEquals(expected = "hello", actual = dataFetcher.get(mockEnvironment).get())
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
```diff
@@ -49,7 +49,7 @@ open class FunctionDataFetcher(
     /**
      * Invoke a suspend function or blocking function, passing in the [target] if not null or default to using the source from the environment.
      */
-    override fun get(environment: DataFetchingEnvironment): Any? {
+    override fun get(environment: DataFetchingEnvironment): CompletableFuture<Any?> {
```
samuelAndalon, line 52: i believe that this change is going to cause incompatiblity with the update to [graphql-java 22](https://github.com/graphql-java/graphql-java/releases/tag/v22.0)

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
```diff
@@ -120,12 +120,15 @@ open class FunctionDataFetcher(
...
-     * If you need to override the exception handling you can override this method.
+     * Once all parameter values are properly converted, this function will be called to run a simple blocking function.
+     * If you need to override the exception handling, you can override this method.
      */
-    protected open fun runBlockingFunction(parameterValues: Map<KParameter, Any?>): Any? = try {
-        fn.callBy(parameterValues)
-    } catch (exception: InvocationTargetException) {
-        throw exception.cause ?: exception
-    }
+    protected open fun runBlockingFunction(parameterValues: Map<KParameter, Any?>): CompletableFuture<Any?>  =
```
samuelAndalon, line 126: This will cause an allocation of an object of type `CompletableFuture` for fields that are already materialized (in-memory).

samuelAndalon:
>in order to cleanly handle cases where runBlockingFunction results in a thrown Error (or descendant of Error).

what if instead  you handle your exceptions in your data fetchers ?, that way, you will have a more resilient way to recover from an error state.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
```diff
@@ -120,12 +120,15 @@ open class FunctionDataFetcher(
...
+        CompletableFuture<Any?>().apply {
```
samuelAndalon, line 127: same as above, this change is going to cause incompatibility with the update to [graphql-java 22](https://github.com/graphql-java/graphql-java/releases/tag/v22.0)

samuelAndalon (commented):
.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
wallind: Can you help me understand more specifically what you mean so that maybe I can try to adjust my solution accordingly?

Otherwise, I'd say I'm happy to remove the change to this `get` function signature as it's not required for my solution to work. Would that be an okay compromise?

samuelAndalon: TLDR; graphql-java 22 introduces polymorphic resolution, before even if your data fetcher was returning a materialized object (not a CF), it was anyways creating a CF creating a terrible memory bottleneck, and GC pause times were paying the price.

having said that, i wonder, why would you want to catch `Error`s ?
screenshot taken from `Error` docs
image

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
wallind: I made the `CompletableFuture` wrap the success-case for consistency, but could also peel that back as it's really only the `throw ...` case that is the problem and inconsistent with the equivalent behavior for `suspend` resolvers.

Shall I do that?

wallind:
> > in order to cleanly handle cases where runBlockingFunction results in a thrown Error (or descendant of Error).
> 
> what if instead you handle your exceptions in your data fetchers ?, that way, you will have a more resilient way to recover from an error state.

If by _"... exceptions ..."_ you mean thrown `Error`s, then yeah I get where you're coming from. I could change our code to handle that scenario.

More broadly though, I'm trying to point out an inconsistency between the way `suspend` and non-`suspend` scopes are handled in the resolvers atm.

For `suspend` scopes (because of `CompletableFuture`) `Error` is _gracefully handled_ via the `errors` response sub-structure:

image

and for non-`suspend` `Error` it is not, and ends up handling by the server's implementation of error-handling:

image

So while yes my original Issue/PR are written as trying to make non-`suspend` behave as `suspend` does currently, I could see the option to actually make the inverse change as being something you'd want based on your response here https://github.com/ExpediaGroup/graphql-kotlin/pull/1966#discussion_r1575359703.

**TLDR** do you want me to adjust this work so that `Error` in the `suspend` case is not _gracefully handled_ in the `errors` response structure?

*_you'll have to pardon my naivety regarding `Exception` vs `Error` as this is a relatively new subject focus for me. After speaking with my colleagues I understand what you mean with things like `StackOverflowError` and `OutOfMemoryError` being descendant from `Error`_

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
wallind: > TLDR; graphql-java 22 introduces polymorphic resolution, in previous versions, even if your data fetcher was returning a materialized object (not a CF), it was anyways creating a CF causing a terrible memory bottleneck where GC pause times were paying the price.

😅  I need to read up on this, as tbh that is _pretty far out of my depth_ of current expertise lol. So I'm inclined to defer to your judgement and undo this part of my change regardless of where the larger change lands as accepted or not.

> having said that, i wonder, why would you want to catch Errors ?
screenshot taken from Error docs

In regards to this, please see my thoughts/discussion here https://github.com/ExpediaGroup/graphql-kotlin/pull/1966#issuecomment-2071060344 if you would.

samuelAndalon:
>TLDR do you want me to adjust this work so that Error in the suspend case is not gracefully handled in the errors response structure?
yes, that sounds good, we would appreciate that contribution

dariuszkuc:
Instead of changing the data fetcher I think we should probably just catch the error here -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt#L99-L137?

wallind:
> > TLDR do you want me to adjust this work so that Error in the suspend case is not gracefully handled in the errors response structure?
> 
> yes, that sounds good, we would appreciate that contribution

Sweet @samuelAndalon! happy to help. Will try to get to that sometime in the next few days...except now I'm a bit confused by @dariuszkuc's suggestion:

> Instead of changing the data fetcher I think we should probably just catch the error here -> https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/execution/GraphQLRequestHandler.kt#L99-L137?

Are you saying to make the change I originally proposed, to make non-`suspend` _not fail out_ but rather be gracefully handled? Or rather are you saying to change it so `suspend` doesn't gracefully handle `Error`..but to make that change in `GraphQLRequestHandler`? Not 100% sure I'm following.

samuelAndalon:
i understand both approaches and i see reasons to follow one or the other, however, IMHO, we should not attempt to catch `Errors`, we have to follow what java mentions about `Error`.

dariuszkuc:
Since `graphql-java` handles exceptionally completed futures (handles `Throwable`) I am unsure whether you can easily force our function data fetcher to blow up from coroutines in the same way as we do for sync execution. I believe this would require inspecting the result of the coroutine which would mean we would have to wait for it to complete....

wallind:
@samuelAndalon and @dariuszkuc I appreciate the thoughtful discussion from both of you, but must admit I am not clear on the desired direction.

For now I am going to defer to @dariuszkuc's suggestion, as he seems to be the primary repository maintainer, and do what I originally intended "_modify handling for non-`suspend` scopes to align with `suspend` as it exists today, meaning `Error` thrown from below is gracefully handled_"; that would be the following change to `GraphQLRequestHandler`:

```kotlin
private suspend fun execute(
  graphQLRequest: GraphQLRequest,
  batchGraphQLContext: GraphQLContext,
  dataLoaderRegistry: KotlinDataLoaderRegistry?,
): GraphQLResponse<*> =
  try {
    graphQL.executeAsync(
      graphQLRequest.toExecutionInput(batchGraphQLContext, dataLoaderRegistry),
    ).await().toGraphQLResponse()
  } catch (throwable: Throwable) {
    // Catch any Throwable, Error or Exception, and convert it to a GraphQLError
    val error = throwable.toGraphQLError()
    GraphQLResponse<Any?>(errors = listOf(error.toGraphQLKotlinType()))
  }
```

If that's not right, or I'm somehow missing the mark don't hesitate to let me know.

*_In the meantime I remain open to discussion, and will certainly not be rushing this through!_

dariuszkuc:
As Sam mentioned, when `Error` happens it generally would mean some "bad things" happened (like OOM) so I don't think we should be updating the default GraphQL server handler to catch those. Ideally coroutines should behave the same way but I don't think we can easily handle that. While this is an inconsistent behavior between coroutines and blocking functions, I think that this is probably a good tradeoff.

So the TLDR is -> we probably shouldn't change anything unless there is a good way to propagate errors out of the completable futures/errors.

wallind:
> As Sam mentioned, when `Error` happens it generally would mean some "bad things" happened (like OOM) so I don't think we should be updating the default GraphQL server handler to catch those. Ideally coroutines should behave the same way but I don't think we can easily handle that. While this is an inconsistent behavior between coroutines and blocking functions, I think that this is probably a good tradeoff.
> 
> So the TLDR is -> we probably shouldn't change anything unless there is a good way to propagate errors out of the completable futures/errors.

Understood! I will look into this _"unless there is a good way to propagate errors out of the completable futures/errors."_ bit sometime this week. From the digging I was doing earlier I don't think it is possible, as I'm sure you may already know.

However it sounds like a fun challenge, so I'll spend some time going after it. Thanks again for your time and energy talking this through with me!

*_I'll cleanup and close the PR/Issue with **non-resolveable** outcome/commentary later this week to keep the repo clean too_

PR #1967, opened 2024-04-25
graphql java 22 update

samuelAndalon:
### :pencil: Description
waiting for an official release of this PR https://github.com/graphql-java/graphql-java/pull/3571

UPDATE:
next graphql-java update will be on July,
https://github.com/graphql-java/graphql-java/milestone/69

they helped us releasing from this branch 
https://github.com/graphql-java/graphql-java/pull/3571

which will unblock us.

https://github.com/graphql-java/graphql-java/releases/tag/v22.0


generator/graphql-kotlin-federation/build.gradle.kts
```diff
@@ -6,7 +6,10 @@ plugins {
 
 dependencies {
     api(projects.graphqlKotlinSchemaGenerator)
-    api(libs.federation)
+    api(libs.federation) {
+        exclude(group = "com.graphql-java", module = "graphql-java")
+    }
+    api(libs.graphql.java)
```
dariuszkuc, line 12: nit: since this is a dependency of schema generator you don't really need to have this explicit dependency (also gradle by default would pick the latest between two transitive dependency but yeah good to stick to the version from schema generator)

generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/data/TestResolvers.kt
```diff
@@ -27,7 +27,7 @@ import java.util.concurrent.CompletableFuture
 internal class BookResolver : FederatedTypeSuspendResolver<Book> {
     override val typeName: String = "Book"
 
-    override suspend fun resolve(environment: DataFetchingEnvironment, representation: Map<String, Any>): Book? {
+    override suspend fun resolve(environment: DataFetchingEnvironment, representation: Map<String, Any>): Book {
```
dariuszkuc, line 30: nit: technically the implementations here always return objects but in the real life examples it should be nullable returns

generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/execution/ServiceQueryResolverTest.kt
```diff
@@ -71,7 +71,7 @@ type Query @extends
 
 type Review {
   body: String!
-  content: String @deprecated(reason : "no longer supported, replace with use Review.body instead")
+  content: String
```
dariuszkuc, line 74: why the change in schema?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FlowSubscriptionExecutionStrategy.kt
```diff
@@ -103,21 +104,33 @@ class FlowSubscriptionExecutionStrategy(dfe: DataFetcherExceptionHandler) : Exec
...
     private fun createSourceEventStream(
         executionContext: ExecutionContext,
         parameters: ExecutionStrategyParameters
     ): CompletableFuture<Flow<*>?> {
         val newParameters = firstFieldOfSubscriptionSelection(parameters)
 
         val fieldFetched = fetchField(executionContext, newParameters)
-        return fieldFetched.thenApply { fetchedValue ->
+        return if (fieldFetched is CompletableFuture<*>) {
+            val promiseFieldFetched = fieldFetched as CompletableFuture<FetchedValue>
```
dariuszkuc, line 116: no need to cast it here as kotlin will already know field fetched is CF

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FlowSubscriptionExecutionStrategy.kt
```diff
@@ -103,21 +104,33 @@ class FlowSubscriptionExecutionStrategy(dfe: DataFetcherExceptionHandler) : Exec
...
+        } else {
+            val fetchedValue = fieldFetched as FetchedValue
             val flow = when (val publisherOrFlow: Any? = fetchedValue.fetchedValue) {
                 is Publisher<*> -> publisherOrFlow.asFlow()
                 // below explicit cast is required due to the type erasure and Kotlin declaration-site variance vs Java use-site variance
                 is Flow<*> -> publisherOrFlow
                 else -> null
             }
-            flow
+            CompletableFuture.completedFuture(flow)
```
dariuszkuc, line 133: I dont' think this will work -> this will mark the execution as complete with the results from completing the flows so it is not really streaming response back anymore

website/docs/server/spring-server/spring-properties.md
```diff
@@ -31,4 +31,4 @@ details on the supported configuration properties.
 | graphql.subscriptions.keepAliveInterval     | **Deprecated**. Keep the websocket alive and send a message to the client every interval in ms. Defaults to not sending messages | null                          |
 | graphql.subscriptions.protocol              | WebSocket based subscription protocol. Supported protocols: APOLLO_SUBSCRIPTIONS_WS / GRAPHQL_WS                                 | GRAPHQL_WS                    |
 | graphql.batching.enabled                    | Boolean flag indicating whether to enable custom dataloader instrumentations for 1 or more GraphQL Operations                    | false                         |
-| graphql.batching.strategy                   | Configure which custom dataloader instrumentation will be used (LEVEL_DISPATCHED or SYNC_EXHAUSTION)                             | LEVEL_DISPATCHED              |
+| graphql.batching.strategy                   | Configure which custom dataloader instrumentation will be used                                                                   | SYNC_EXHAUSTION               |
```
dariuszkuc, line 34: since we only have single data loader batching strategy do we really need this config?

generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/execution/ServiceQueryResolverTest.kt
samuelAndalon: >why the change in schema?

https://github.com/apollographql/federation-jvm/blob/main/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/printer/ServiceSDLPrinter.java#L33

the `@deprecated` directive is hidden from the `ServiceSDLPrinter`

i don't know why with previous versions the test was passing though.

website/docs/server/spring-server/spring-properties.md
samuelAndalon: i would like to keep the door open for future dispatch mechanisms, `SYNC_EXHAUSTION` works but comes with a price, needs to hold state of each field, and memory usage could be a concern, in the future we could try to have a different dispatch mechanism that will use less memory maybe.

generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/data/TestResolvers.kt
samuelAndalon: just did this to keep the IDE happy.

website/docs/server/spring-server/spring-properties.md
dariuszkuc: The keyword is "might" -> Unless there is something already in the works/roadmap I am unsure if there is ever going to be one so I'd just drop it and reintroduce it later on if needed.

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FlowSubscriptionExecutionStrategy.kt
samuelAndalon: should we just imply that the result of `fetchField` will always be a `CompletableFuture` ?

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FlowSubscriptionExecutionStrategy.kt
```diff
@@ -103,21 +104,33 @@ class FlowSubscriptionExecutionStrategy(dfe: DataFetcherExceptionHandler) : Exec
...
+    @Suppress("UNCHECKED_CAST")
     private fun createSourceEventStream(
         executionContext: ExecutionContext,
         parameters: ExecutionStrategyParameters
     ): CompletableFuture<Flow<*>?> {
         val newParameters = firstFieldOfSubscriptionSelection(parameters)
 
         val fieldFetched = fetchField(executionContext, newParameters)
-        return fieldFetched.thenApply { fetchedValue ->
+        return if (fieldFetched is CompletableFuture<*>) {
```
dariuszkuc, line 115: I think we need to rethink this logic and potentially update it to just match [SubscriptionExecutionStrategy](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/SubscriptionExecutionStrategy.java#L93-L105), i.e. always wrap the fetched field in the future and keep the old logic as is?



- if it is a future -> we should handle it the same way as in the past (i.e. it should be publisher)
- if it is a flow ->

dariuszkuc: in our case `Flow` (same as `Publisher`) will always be non CF value (i.e. I am unsure in what situation you would get a CF for subscription value).

dariuszkuc: You might also want to check how `spring-graphql` is handling this (I'd assume they handle `Flow`)

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FlowSubscriptionExecutionStrategy.kt
samuelAndalon: just commited the change to always wrap.

website/docs/server/spring-server/spring-properties.md
dariuszkuc: Also I find `graphql.batching.enabled` property confusing -> we always support batching so this only applies to dataloader config (as per comments). Wondering whether this should be deprecated in favor of something like `graphql.loader.batching.strategy = SYNC_EXHAUSTION | NONE` (`NONE` would imply current `graphql.batching.enabled = false`)

samuelAndalon: >we always support batching
we added `graphql.batching.enabled` in order to opt out from built-in "by level" graphql-java dataloader instrumentation.

https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLSchemaConfiguration.kt#L90

dariuszkuc: yeah we added it originally for the custom data loaders -> just wondering whether we should revisit this property naming for the next major

dariuszkuc (approved):
LGTM. Data loader config properties can be fixed/dropped later if needed.

PR #1980, opened 2024-06-18
feat: Use official graphql java 22 version

eocantu:
### :pencil: Description

Follow up to https://github.com/ExpediaGroup/graphql-kotlin/pull/1967 to use an official graphql java release.

### DIFF
--- a/executions/graphql-kotlin-dataloader-instrumentation/build.gradle.kts
+++ b/executions/graphql-kotlin-dataloader-instrumentation/build.gradle.kts
@@ -9,6 +9,7 @@ dependencies {
     api(libs.graphql.java) {
         exclude(group = "com.graphql-java", module = "java-dataloader")
     }
+    testImplementation(projects.graphqlKotlinServer)
     testImplementation(libs.reactor.core)
     testImplementation(libs.reactor.extensions)
 }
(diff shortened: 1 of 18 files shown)


executions/graphql-kotlin-dataloader-instrumentation/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ dependencies {
     api(libs.graphql.java) {
         exclude(group = "com.graphql-java", module = "java-dataloader")
     }
+    testImplementation(projects.graphqlKotlinServer)
```
dariuszkuc, line 12: why do you need to bring this dependency?

eocantu: It's to have access to the [dataFetchingEnvironmentExtensions](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-server/src/main/kotlin/com/expediagroup/graphql/server/extensions/dataFetchingEnvironmentExtensions.kt) in the tests.

dariuszkuc: I think it is somewhat problematic as you end up with cyclic dependency -> i.e. `graphql-kotlin-server` depends on `graphql-kotlin-dataloader-instrumentation` as API dependency ([link](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/servers/graphql-kotlin-server/build.gradle.kts#L13)) and we depend here on the server as a test dependency.

Guessing Gradle is smart enough to handle the API vs test dependency.... but I'd avoid it and either copy over the extension locally to one of the tests or simply do what extension does....

dariuszkuc (approved):

PR #1989, opened 2024-06-27
chore(deps): use springboot 3.2.x for now

samuelAndalon:


tapaderster (approved):

eocantu (approved):

dariuszkuc:
Why the downgrade?

richeyh:
> Why the downgrade?

spring boot 3.3 introduced the brave 5 -> 6 major upgrade which may not have any problems in this library but can be problematic for teams using this that have custom integrations with brave.

eocantu:
> Why the downgrade?

Spring Boot 3.3 brings in a new major version of Brave 6 and we want to minimize the impact. Apps should still be able to pull in spring boot 3.3 if they desire.

Issue #2037, opened 2024-09-28
Ktor 3.xx support


ronjunevaldoz:
**Is your feature request related to a problem? Please describe.**
8.0.0 is currently based on ktor 2.x.x

**Describe the solution you'd like**
I'd like to have a full support for ktor 3.x.x

timemanx:
@ronjunevaldoz, @TheDome0 I created #2066 on top of @TheDome0's PR.
It passes all the build checks and works for me too.

If possible, could you check whether it works for you.

PR #2055, opened 2024-11-10
upgrade to ktor 3

TheDome0:
### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/2037

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,7 +12,7 @@ kotlinx-benchmark = "0.4.11"
 kotlinx-coroutines = "1.8.1"
 # TODO kotlin 1.9 upgrade: fix GraphQLTestUtils and GenerateKotlinxClientIT
 kotlinx-serialization = "1.6.3"
-ktor = "2.3.12"
+ktor = "3.0.1"
 fastjson2 = "2.0.53"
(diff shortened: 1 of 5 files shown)


TheDome0:
there are still three tests failing, two of them due to an NPE in `KtorGraphQLRequestParser` where I'm not really sure what's happening

ronjunevaldoz:
any update on this one  :)

TheDome0:
No, because in v9 I ran into a different problematic issue where you cannot have a class which implements an interface as a response for graphql queries.

`ConflictingTypesException: Conflicting class names in schema generation [class com.example.TestError, class com.example.TestError]`
```kotlin
sealed interface Err {
    val message: String
}

data object TestError: Err {
    override val message: String = "Test error"
}

class TestQueries : Query {
    suspend fun test(): TestError = TestError
}
```

Removing the message property from the interface fixes it, but there are also other situations like implementing multiple interfaces (typical error union setup) where this happens.

I tried adding a custom willgenerateGraphQLType hook but then I'm getting `You have redefined the type 'TestError' from being a 'GraphQLObjectType' to a 'GraphQLObjectType'`.

And this does not happen in v8 / prior to Kotlin 2.0

https://github.com/TheDome0/graphql-repros

PR #2079, opened 2025-04-10
feat: SingletonPropertyDataFetcher, avoid allocating a PropertyDataFetcher per property per graphql operation

samuelAndalon:
### :pencil: Description

Inspired by https://github.com/graphql-java/graphql-java/pull/3754.
Currently, graphql-kotlin, through the [KotlinDataFetcherFactoryProvider](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/KotlinDataFetcherFactoryProvider.kt#L62) creates a [PropertyDataFetcher](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/PropertyDataFetcher.kt) per source's property.

This instance is created [every single time ](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/GraphQLCodeRegistry.java#L100)the graphql-java [DataFetcherFactory](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/DataFetcherFactory.java) is invoked, which happens to be on runtime per property per graphql-operation.

This PR will introduce a new object class `SingletonPropertyDataFetcher` which will host its own  singleton factory that will always return `SingletonPropertyDataFetcher` which will store all `KProperty.Getter<*>`s in a ConcurrentHashMap.

Instead of just replacing the SimpleKotlinDataFetcherFactoryProvider, I am creating a new one, to avoid breaking changes, and to allow users to decide what they want, this switch might come with a cost, we are avoiding object allocations, in favor of a singleton that will possibly hold thousands of `KProperty.Getter<*>`s.

### DIFF
--- a/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLExecutionConfiguration.kt
+++ b/servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLExecutionConfiguration.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Expedia, Inc
+ * Copyright 2025 Expedia, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


servers/graphql-kotlin-spring-server/src/main/kotlin/com/expediagroup/graphql/server/spring/GraphQLExecutionConfiguration.kt
```diff
@@ -42,7 +42,7 @@ class GraphQLExecutionConfiguration {
     @Bean
     @ConditionalOnMissingBean
     fun dataFetcherFactoryProvider(applicationContext: ApplicationContext): KotlinDataFetcherFactoryProvider =
-        SpringKotlinDataFetcherFactoryProvider(applicationContext)
+        SpringSingletonKotlinDataFetcherFactoryProvider(applicationContext)
```
samuelAndalon, line 45: by default use `SpringSingletonKotlinDataFetcherFactoryProvider` if no `KotlinDataFetcherFactoryProvider` bean is in the context.

tapaderster: should this be an opt-in instead and become default in future releases?

tapaderster (approved):

richeyh (approved):

PR #2088, opened 2025-04-14
feat: add label as optional arg to @override directive

Samjin:
Support federation v2.7. Here's [directive details](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives)
Support [progressive @override directive](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#progressive-override). 

These directives will require specific federation versions to generate schema. 

1. `@authenticate` directive is generated from `2.5+` 
2. `@requiresScopes` directive is generated from `2.5+` 
3. `@policy` directive is generated from `2.6+` 
4. `@composeDirective` directive is generated from `2.1+` 
5. `@interfaceObject ` directive is generated from `2.6+` 
6. `@override` directive with 2nd argument of `label` from `2.7`

### DIFF
--- a/examples/client/server/src/main/kotlin/com/expediagroup/graphql/examples/client/server/FederatedQuery.kt
+++ b/examples/client/server/src/main/kotlin/com/expediagroup/graphql/examples/client/server/FederatedQuery.kt
@@ -96,7 +96,7 @@ data class ProductVariation(
 data class User(
     @ExternalDirective
     val email: String,
-    @OverrideDirective(from = "users", label = "Migrating name field")
+    @OverrideDirective(from = "users")
     val name: String,
     @ExternalDirective
     val totalProductsCreated: Int? = null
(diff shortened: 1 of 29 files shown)


dariuszkuc:
Currently this lib allows you specify you own custom `@link` federation spec import -> this change does not account for this and will update directive definitions even if folks specifically target older versions of federation

Samjin:
> Currently this lib allows you specify you own custom `@link` federation spec import -> this change does not account for this and will update directive definitions even if folks specifically target older versions of federation

Had a brief discussion, the plan is to check if progressive override label would work with older fed version from composition. if not, we'll have to check fed version per directives to make sure they are correctly generated.

Samjin:
Somehow fed `2.7.5` with Router `1.47.0` 
but `2.7.2` with `v1.45.1` worked. 
https://www.apollographql.com/docs/graphos/routing/federation-version-support

PR #2116, opened 2025-05-27
fix: respect custom coroutine scope when calling blocking functions in a data fetcher

maxkomarychev:
### :pencil: Description

When running blocking function data fetcher it does not respect coroutine scope provided in the GraphQL context.

In my project we're creating custom a coroutine scope in order to propagate MDC and OTEL contexts in a coroutine like this:

```
CoroutineScope(MDCContext() + Dispatchers.IO + Context.current().asContextElement())
```

(here `Context` is an instance of OTEL context).

This works well in suspended functions since they are using the context however normal functions are not using this context and therefore none of the OTEL context (trace id etc) is propagated.

### DIFF
--- a/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
+++ b/generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
@@ -24,6 +24,7 @@ import graphql.schema.DataFetcher
 import graphql.schema.DataFetchingEnvironment
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.future.future
+import kotlinx.coroutines.runBlocking
 import java.lang.reflect.InvocationTargetException
 import java.util.concurrent.CompletableFuture
 import kotlin.coroutines.EmptyCoroutineContext
@@ -60,7 +61,7 @@ open class FunctionDataFetcher(
             if (fn.isSuspend) {
                 runSuspendingFunction(environment, parameterValues)
             } else {
-                runBlockingFunction(parameterValues)
+                runBlockingFunction(environment, parameterValues)
             }
         } else {
             null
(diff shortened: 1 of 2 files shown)


zsiegel:
👍🏻

generator/graphql-kotlin-schema-generator/src/main/kotlin/com/expediagroup/graphql/generator/execution/FunctionDataFetcher.kt
```diff
@@ -123,8 +124,14 @@ open class FunctionDataFetcher(
...
      * If you need to override the exception handling you can override this method.
      */
-    protected open fun runBlockingFunction(parameterValues: Map<KParameter, Any?>): Any? = try {
-        fn.callBy(parameterValues)
+    protected open fun runBlockingFunction(
+        environment: DataFetchingEnvironment,
+        parameterValues: Map<KParameter, Any?>
+    ): Any? = try {
+        val coroutineScope = environment.graphQlContext.getOrDefault(CoroutineScope(EmptyCoroutineContext))
+        runBlocking(coroutineScope.coroutineContext) {
```
samuelAndalon, line 132: I do understand the reasoning behind this change, but think about it, running a blocking function in a coroutine could be catastrophic just because you need context propagation, there should be another way to deal with this issue.

maxkomarychev: > I do not know your use case, but this issue requires a different solution, and for now, what I would advice is to mark the resolver/function as suspend, what is blocking inside your resolver ?

Dataloaders don't play nice with suspended functions. Because of that all functions which resolve properties via dataloaders are not suspended and the trace breaks.

maxkomarychev: I understand the concern of running a blocking code in a coroutine but it's not dramatically worse than running it in an arbitrary thread, is it?

What if we supply another coroutine scope specifically for sync functions to avoid blocks on suspended functions?

samuelAndalon: is not blocking code in reality, is CPU bound work.

> but it's not dramatically worse than running it in an arbitrary thread

It could be, however, context switching it could be even worst and expensive, lets give it a try, but maybe make it optional through a configuration or something so we can opt in for the functionality and compare CPU usage (context switching concern)

maxkomarychev: sure! what would be the best place to introduce a configuration parameter for this kind of a thing?

PR #2165, opened 2026-04-15
fix: stop validating @requires field set contents

### LINKED ISSUE #1939: FederatedSchemaValidator rejects nested union type field set
**Library Version**
7.0.2

**Describe the bug**
When I have a union type and want to use it on `@requires` directive, it fails. When I use:
```kotlin
@RequiresDirective(fields = FieldSet("animal { ... on Dog { breed { name } } }"))
```

I get something like
```
com.expediagroup.graphql.generator.federation.exception.InvalidFederatedSchema: Invalid federated schema:
  - @requires(fields = "animal { ... on Dog { breed { name } } }") directive on <...redacted...> specifies invalid field set - field set specifies field that does not exist, field=...
  - @requires(fields = "animal { ... on Dog { breed { name } } }") directive on <...redacted...> specifies invalid field set - field set specifies field that does not exist, field=on
  - @requires(fields = "animal { ... on Dog { breed { name } } }") directive on <...redacted...> specifies invalid field set - field set specifies field that does not exist, field=Dog
 ```

However, if I do not add the annotation & just edit federated schema, or remove validations by overriding `didGenerateGraphQLType` on `FederatedSchemaGeneratorHooks`. Subgraph is federated and works as expected

**Expected behaviour**
support for this kind of directives, without validation errors.

Also, would it make sense to make `validator: FederatedSchemaValidator` injectable on `FederatedSchemaGeneratorHooks`  and open  for overrides so that there was an easy way to disable validation in this kind of edge-case scenarios?

Apollo mentions similar usage pattern on Provides directive documentation: https://www.apollographql.com/docs/federation/federated-types/federated-directives/#fields-1

### PR DESCRIPTION
**Pull request by glasser**
## Summary

The field set validator in `FederatedSchemaValidator` splits on spaces and parses field sets as flat token lists. It has no understanding of inline fragments (`... on TypeName { ... }`) or `__typename`, so it rejects valid `@requires` directives like:

```kotlin
@RequiresDirective(fields = FieldSet("animal { ... on Dog { breed } }"))
```

Rather than fixing the naive parser, this PR skips field set content validation for `@requires` as suggested by @dariuszkuc in #1939. Validation for `@key` and `@provides` is retained — `@key` in particular [disallows unions and interfaces per the federation spec](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1641-L1646).

(While we could improve the naive parser to handle more GraphQL syntax, its attempt to validate that fields exist is fundamentally doomed anyway — `@requires` fields don't have to be in *this subgraph*, so we'd need a full supergraph schema to properly validate `@requires` in the way we currently try to, which surely would have to be an optional feature.)

### Changes

- Make `fieldMap` parameter in `validateDirective` nullable; when null, only the non-empty check runs
- `FederatedSchemaValidator` passes `fieldMap` for `@key`/`@provides` but passes explicit `null` for `@requires`
- Remove `@requires` failure integration tests and their test data fixtures
- Add regression test for #1939: `@requires` with inline fragments on a union type
- Add unit test verifying that `fieldMap = null` skips deep validation

Fixes #1939.


glasser:
I reached out directly to @dariuszkuc and I'm going to make this only apply to `@requires`, not `@key`/`@provides`, so it'll be a much smaller diff.

glasser:
OK, that's better.

dariuszkuc:
additional context: since `@requires` references fields from other subgraphs, it has to be valid at a supergraph level (i.e. local graph may not have all the information about the final merged field, e.g. it may be unaware of some implementations and/or union members). Composition runs the `@requires` validation at the supergraph level.

samuelAndalon (approved):

PR #2173, opened 2026-05-14
feat: allow priority override when multiple SchemaGeneratorHooksProviders are registered

### LINKED ISSUE #2150: generateSDL to handle multiple SchemaGeneratorHooksProviders found on the classpath
- `graphql-kotlin-sdl-generator` uses `ServiceLoader` to find `SchemaGeneratorHooksProvider`, but [only allows a single match](https://github.com/ExpediaGroup/graphql-kotlin/blob/master/plugins/schema/graphql-kotlin-sdl-generator/src/main/kotlin/com/expediagroup/graphql/plugin/schema/GenerateSDL.kt#L52). 
- Multiple `SchemaGeneratorHooksProvider` configured on the classpath (in `META-INF/services/`) makes the plugin throw an exception, no questions asked. The plugin does not provide a way set precedence for multiple matches.
- If any library on the classpath (built on top of `graphql-kotlin`) defines `SchemaGeneratorHooksProvider` service, users have no choice but to exclude that library (or make [an ugly workaround with the shade plugin](https://stackoverflow.com/a/42541191)) if they want to redefine it.

To be able to set a clear precedence when multiple `SchemaGeneratorHooksProvider`s are defined. I would propose a solution similar to the `@Primary` and `@Order` annotations of the Spring Framework. The library can define it's own annotations or can consider to support the `javax.annotation.Priority` as well.

Example implementation:
```kotlin
ServiceLoader.load(SchemaGeneratorHooksProvider::class.java)
    .maxByOrNull { it::class.findAnnotation<Priority>()?.value ?: 0 }
```

As an alternative, we could play around with the `ClassLoader` mechanism, maybe to prefer definitions from the current project over the ones coming from the libraries. I haven't looked into the feasibility of this idea, however.

gaurav0107:
### :pencil: Description

Adds a backward-compatible `priority()` default method to the `SchemaGeneratorHooksProvider` SPI (returns `0` by default) and teaches both `generateSDL` and `generateGraalVmReflectMetadata` to pick the provider with the strictly highest priority when more than one provider is discovered through `ServiceLoader`.

**Behaviour preserved for existing users:**
- Zero providers still warn and fall back to `NoopSchemaGeneratorHooks`.
- Exactly one provider still wins unchanged.
- Multiple providers at the same priority still fail fast. The error message keeps the legacy `"Cannot generate SDL as multiple SchemaGeneratorHooksProviders were found on the classpath"` prefix but now also lists the tied class names and their priority so the conflict can be diagnosed without shade-plugin hacks.

**Opt-in path for library authors.** A library that ships a `SchemaGeneratorHooksProvider` which must override a transitively-imported one can now override `priority()` to return a positive value instead of telling consumers to exclude the upstream library.

**Scope.**
- SPI: adds a default method to `SchemaGeneratorHooksProvider` (source- and binary-compatible).
- Call sites: replaces the inline `when` block in `GenerateSDL.kt` and `GenerateGraalVmMetadata.kt` with a small package-local `selectHooksProvider(providers, logger)` helper in each module.
- Tests: adds unit tests for the new selection helper in both modules (single-provider passthrough, highest-priority wins, same-priority tie throws with descriptive message).

A shared extraction into `graphql-kotlin-hooks-provider` felt out of scope for this change — both helpers are tiny and duplicating them keeps the SPI module dependency-free on SLF4J. Happy to hoist them if maintainers prefer.

### :link: Related Issues

Closes #2150

### DIFF
--- a/plugins/schema/graphql-kotlin-hooks-provider/src/main/kotlin/com/expediagroup/graphql/plugin/schema/hooks/SchemaGeneratorHooksProvider.kt
+++ b/plugins/schema/graphql-kotlin-hooks-provider/src/main/kotlin/com/expediagroup/graphql/plugin/schema/hooks/SchemaGeneratorHooksProvider.kt
@@ -33,4 +33,17 @@ interface SchemaGeneratorHooksProvider {
      * Create a new instance of a SchemaGeneratorHooks that will be used to generate GraphQL schema in SDL format.
      */
     fun hooks(): SchemaGeneratorHooks
+
+    /**
+     * Priority used to break ties when multiple `SchemaGeneratorHooksProvider`s are discovered on the classpath.
+     *
+     * When more than one provider is found, the one with the strictly highest priority is selected. If multiple
+     * providers share the highest priority the generator plugins still fail with an error so the conflict is surfaced
+     * to the user. Providers that do not override this method fall back to the default priority of `0`, which
+     * preserves the pre-existing fail-fast behaviour for users that have not opted in.
+     *
+     * Library authors that ship a `SchemaGeneratorHooksProvider` which must override a transitively-imported
+     * provider should return a positive value. Larger values win.
+     */
+    fun priority(): Int = 0
 }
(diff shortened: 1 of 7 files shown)


gaurav0107:
Gentle check-in on this one — it's still open and mergeable. Happy to rebase or address any feedback whenever a maintainer has a moment to take a look. Thanks!

Issue #1891, opened 2023-11-22
How to annotate schema to use directives supported by graphql-java?


fediazgon:
`graphql-java` added support for the `@oneOf` directive [in the 21.2 release](https://github.com/graphql-java/graphql-java/releases/tag/v21.2). I understand that `graphql-kotlin` is still using the `21.1` release, but I was wondering if I will be able to use this directive out-of-the-box.

In the docs I can see that other directives like `@skip` are supported:

However, how do I annotate my schema and use `graphql-kotlin-schema-generator` to generate this annotations for me? What I want is to write Kotlin code like this:

```kotlin
@GraphqlOneOf
data class Some(
 val a: String?
 val b: Int?
)
```

and `graphql-kotlin-schema-generator` will generate code like:

```graphql
type Some @oneOf { ... }
```

Do I have to use custom directives for this?

dariuszkuc:
Hello 👋 
Currently there is no built-in mechanism to support it as `@oneOf` is still not part of the GraphQL spec. While some implementations already added the support, [RFC](https://github.com/graphql/graphql-spec/pull/825) is still open (but it is getting close to be finalized). Once RFC is merged to the spec we'll work on adding the support.

Haven't tried it but 'm guessing you might be able to support it by writing some custom hooks + data fetcher that could handle some sort of wrapper.

Thanks,
Derek

fediazgon:
My second part of the question was: it seems there are multiple directives that are supported according to the docs (e.g., `@skip`, `@include`). However these do not have their corresponding annotation in `com.expediagroup.graphql.generator.annotations` as, for example, `@deprecated` has `@GraphQLDeprecated`. How do I annotate my Kotlin schema to make use of these annotations?

dariuszkuc:
`@skip`/`@include` are executable directives so they are not part of your schema but instead are provided by clients writing their queries. Since there is nothing to represent in the schema, we don't provide annotations for them.

`@specifiedBy` directive is a schema directive that is applicable on custom scalars so technically we could provide corresponding annotation... but since you still need to provide manual configuration for your custom scalars (e.g. coercing logic) it's seemed like that would be the right place to do it.

ondrejivann:
Hello, I would like to ask when the @oneOf directive will be available? Thank you.

samuelAndalon:
released

PR #2183, opened 2026-06-17
feat: @oneOf input directive support

samuelAndalon:
### :pencil: Description
Add support for GraphQL `@oneOf` directive for input objects .

See: 
https://graphql.org/blog/2025-09-04-multioption-inputs-with-oneof/
https://spec.graphql.org/September2025/#sec-OneOf-Input-Objects

#### What graphql-java provides
graphql-java owns the directive definition and all input validation— runs before our convertArgumentValue:

- Directive def — @oneOf SDL definition baked into [graphql-java](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/Directives.java#L244).
- Input validation/coercion — ValuesResolverOneOfValidation (source) enforces oneOf rules at execution time, per spec:
- exactly one field supplied,
- [that field's value non-null](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/OneOfNullValueException.java),
- [rejects map with 0 or 2+ keys -> error before resolver runs](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/OneOfTooManyKeysException.java).

  So by the time graphql-kotlin runtime code sees the argument map, it is already validated: guaranteed single non-null entry. That's why mapToOneOfKotlinObject can safely
  do input.entries.single() — no defensive count check needed.

#### What graphql-java does NOT cover

graphql-java works in graphql land. It knows nothing about Kotlin sealed types. It does not:

- generate the @oneOf input object from Kotlin types
- know which sealed subtype a given field maps to
- construct the Kotlin instance at runtime during operation execution

#### What graphql-kotlin adds
1. Schema generation (generateOneOfInputObject / generateOneOfInputProperty)
    - Kotlin sealed interface -> @oneOf input object.
    - Each subtype (@GraphQLOneOfField) -> one input field
2. Argument mapping (mapToOneOfKotlinObject in convertArgumentValue)
    - Takes the already-validated single-entry map from graphql-java.
    - Matches fieldName -> sealed subtype via @GraphQLOneOfField.

### :link: Related Issues
https://github.com/ExpediaGroup/graphql-kotlin/issues/1891

### Examples
image
image

### DIFF
--- a/examples/server/spring-server/src/main/kotlin/com/expediagroup/graphql/examples/server/spring/query/OneOfQuery.kt
+++ b/examples/server/spring-server/src/main/kotlin/com/expediagroup/graphql/examples/server/spring/query/OneOfQuery.kt
@@ -16,6 +16,7 @@
 
 package com.expediagroup.graphql.examples.server.spring.query
 
+import com.expediagroup.graphql.generator.annotations.GraphQLDeprecated
 import com.expediagroup.graphql.generator.annotations.GraphQLDescription
 import com.expediagroup.graphql.generator.annotations.GraphQLOneOf
 import com.expediagroup.graphql.generator.annotations.GraphQLOneOfField
@@ -73,6 +74,7 @@ sealed interface ContentBlockInput {
 
     @GraphQLDescription("Paragraph content supplied as a wrapped @oneOf object field.")
     @GraphQLOneOfField("paragraph")
+    @GraphQLDeprecated("This paragraph field is deprecated.")
     data class Paragraph(
         @param:GraphQLDescription("The paragraph text.")
         val text: String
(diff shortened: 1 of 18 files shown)


Samjin:
When `UNWRAPPED` is used
```kotlin
@GraphQLOneOfField("id", UNWRAPPED)
  data class ById(
    @GraphQLType("ID")
    @GraphQLDescription("user id")
    @Deprecated("reason")
    val value: String
  ) : UserBy
```

It should return ID and not losing other metadatas
```
input UserBy @oneOf {
    id: String  // Should be ID
}
```

Samjin (approved):

PR #2201, opened 2026-07-09
feat(federation) Support Apollo Federation 2.15

BoD:
### :pencil: Description

Bump `FEDERATION_SPEC_LATEST_VERSION` `2.7` -> `2.15`, add add support for these directives:

- `@cacheTag` 
- `@cost`
- `@listSize`
- `@context`
- `@fromContext`


generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/types/ContextFieldValue.kt
```diff
@@ -0,0 +1,82 @@
...
+            else -> throw CoercingSerializeException(
+                "Cannot serialize $dataFetcherResult. Expected type 'ContextFieldValue' but was '${dataFetcherResult.javaClass.simpleName}'."
+            )
+        }
+
+    override fun parseValue(input: Any, graphQLContext: GraphQLContext, locale: Locale): ContextFieldValue =
+        when (input) {
+            is ContextFieldValue -> input
+            is StringValue -> ContextFieldValue::class.constructors.first().call(input.value)
+            else -> throw CoercingParseLiteralException(
```
Samjin: do you mean to return `CoercingParseValueException` since fn is parseValue?

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/directives/CostDirective.kt
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * The `@cost` directive defines a custom weight for a schema location. For GraphOS Router, it customizes the operation cost calculation of the
+ * [demand control feature](https://www.apollographql.com/docs/router/executing-operations/demand-control/).
+ *
+ * If `@cost` is not specified for a field, a default value is used:
+ * - Scalars and enums have default cost of 0
+ * - Composite input and output types have default cost of 1
+ *
+ * Regardless of whether `@cost` is specified on a field, the field cost for that field also accounts for its arguments and selections.
+ * ```
```
Samjin, line 41: nit: not needed?

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/directives/ListSizeDirective.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    if (listSizeDirective.sizedFields.isNotEmpty()) {
+        builder.argument(
+            GraphQLAppliedDirectiveArgument.newArgument()
+                .name(LIST_SIZE_DIRECTIVE_SIZED_FIELDS_PARAM)
+                .type(GraphQLList.list(GraphQLNonNull(Scalars.GraphQLString)))
+                .valueProgrammatic(listSizeDirective.sizedFields.toList())
+                .build()
+        )
```
Samjin, line 152: when `@ListSizeDirective(sizedFields = ["<field_name>"])` is used, can we validate the field name in case of typo which break operation at runtime?

Could you also take a look `slicingArguments` and `@fromContext` etc check if they should have validation in place.

generator/graphql-kotlin-federation/src/test/kotlin/com/expediagroup/graphql/generator/federation/directives/override/OverrideDirectiveTest.kt
```diff
@@ -35,7 +35,7 @@ class OverrideDirectiveTest {
     fun `verify override directive definition for fed27`() {
```
Samjin: nit: it's not longer fed27

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/types/ContextFieldValue.kt
BoD: Good catch! I mirrored how it's done with `FieldSetCoercing` which was also wrong, so I fixed both - and a few other places throwing the wrong exception too, in e3c1029fe34b705f5fd36ff9916a894838fe810c

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/directives/CostDirective.kt
BoD: Sorry, did you mean that last line? I just copied it from [the definition](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#cost). Happy to make the KDoc shorter if it's too verbose.

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/directives/ListSizeDirective.kt
BoD: Added validation for `@listSize` in 1d8d662ff9d16c88f44b197953cbc3da250034a2.

Note: a validation for `@context` vs `@fromContext` is also theoretically possible, but I think this would be outside the scope of the library.

Samjin (approved):

Samjin:
Approved, but please take a look of failed checks thanks.

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/validation/validateListSize.kt
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Validates a single field's `@listSize` directive usage (if present):
+ *
+ * Verifies:
+ * - `slicingArguments` references existing arguments on the field
+ * - `sizedFields` references existing fields on the field's return type
+ */
+internal fun validateListSize(typeName: String, field: GraphQLFieldDefinition): List<String> {
+    val directive = field.getAppliedDirective(LIST_SIZE_DIRECTIVE_NAME) ?: return emptyList()
```
Samjin: GK emits `directive @federation__listSize..` when the new annotation is used on a type, because linkSpecs is designed to not import the bare names after fed 2.3+. That means all directives after fed2.3+ will be namespaced without @link import. The `LIST_SIZE_DIRECTIVE_NAME` here maybe `listSize` which isn't namespaced and skip the validation.

Samjin (changes requested):

generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/validation/validateListSize.kt
Samjin: I think this kind of validation should happen at Apollo layers/supergraph instead of in this lib. Here's the details from [demand-control](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#configuring-dynamic-list-sizes). I only read the [Apollo Directive](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives) which doesn't mention the nested path below

```graphql
  search(input: SearchInput!): [Book]
    @listSize(slicingArguments: ["input.pagination.first"])

  # and nested sizedFields selection paths:

  deepContainer(first: Int): DeepContainer
    @listSize(
      slicingArguments: ["first"]
      sizedFields: ["results { page }"]
    )
```

I don't think we can validate such path in this lib. Apology for the back and forth.

BoD: Wow good point, I didn't know about nested paths, and so the validation I implemented is wrong - sorry about that. I also felt it was maybe a bit out of scope for the library to make these checks. I reverted the validation.

Samjin (approved):

